PR #4, opened 2020-01-14
UADIGITAL-2125: Added license and copyright files.

joeparsons:
I thought we had already added these files.


joeparsons:
The changes from PR #5 are also included here.

trackleft:
Before this license is official I think we will need to follow the steps outlined here: https://www.gnu.org/licenses/gpl-howto.en.html

PR #5, opened 2020-01-14
UADIGITAL-2210: Replaced BitBucket with GitHub in ProboCI config.

joeparsons:

PR #6, opened 2020-01-15
UADIGITAL-2211: Update Repository Workflow in Local Development

tadean:
This pull request updates the URLs that are used for building local development dependencies to use the github urls rather than the bitbucket urls.


tadean:
Merged in @joeparsons changes to the probo config in #5  so that automated tests could pass.

PR #10, opened 2020-01-24
Add general pull request template

jwmoore1:
This is a pull request template that contains some basic info that should be included when making a pull request.


.github/ISSUE_TEMPLATE/security_vulnerability.md
```diff
@@ -0,0 +1,13 @@
...
+title: ''
+labels: ''
+assignees: ''
+
+---
+
+**Do not disclose details of the security issue**
+
+Please disclose you found a security vulnerability and admins will contact you
+to get details.
```
joeparsons: Per Friday meeting discussion, let's not add a security vulnerability template and update the notice in the other templates to direct people to email us instead.

PR #12, opened 2020-01-24
Add contributing document

jwmoore1:
This is simply a starting point for general contributing information.
This will change over time

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -47,6 +47,12 @@ We follow Drupal
 
 First, make sure there is an issue associated with your pull request.
 
+Use proper branch naming conventions based on your issue type:
+ * `feature/<issue-number>`
+ * `bug/<issue-number>`
+
+So, bug fix for issue #123 would be on branch `bug/123`
+
 Create a draft pull request if you'd like to run automated tests and/or get 
 feedback before your pull request is completely ready for review.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,75 @@
+# How to contribute
+
+First off, thanks for taking the time to contribute to AZ Quickstart! 
+
+AZ Quickstart is created by [Arizona Digital](https://digital.arizona.edu/), a
+team of web-focused volunteers that meet weekly to create projects like [Arizona Bootstrap](https://digital.arizona.edu/ua-bootstrap) and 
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+```
+
+## Local development
+
+To create a local copy of az_quickstart and build a working Drupal 8 site from
+it, do the following:
+```
+git clone https://github.com/az-digital/az_quickstart.git foldername
+cd foldername
+lando start
```
joshuasosa: Any DDEV/alternative option? TBA?

jwmoore1: A DDEV option can certainly be added, this is just a starting point based off the work Troy did with Lando.

tadean: I believe the existing DDEV configuration still lives in the scaffolding repo, and probably could be brought over to this repository. We may also want to update it to the new model where packagist is sidestepped, eg. the composer project is set up to point to a local composer repository.

In the interim though we could probably at least give a list of ddev commands that accomplish those steps manually.

CONTRIBUTING.md
```diff
@@ -0,0 +1,84 @@
...
+
+### Security Issues
+
+If you found a security vulnerability and it's related to Drupal core or a 
+Drupal contrib module, please follow 
+[these](https://www.drupal.org/drupal-security-team/security-team-procedures/drupal-security-team-disclosure-policy-for-security) instructions.
+
+If it's a security issue related to `az_quickstart` code, please email us here:
+az-digital@web.arizona.edu
+
```
kevcooper: Got some feedback from ISO about this. Paul advised us to consult our ISRM surrounding the risk involved with accepting security reports via email. My opinion as the ISRM for Arizona Digital is that the risk of not providing an easy, universal, way to report an issue outweighs the risk of that report being discovered by someone gaining access to our private slack channel which the security email address feeds into. As such, we are accepting this risk in order to mitigate the greater risk of someone choosing to not report a vulnerability they have discovered in our codebase. Utilizing slack instead of a shared inbox also increases the timeliness of our response to the incoming security issue report which may be critical depending on the issue severity.

kevcooper: I think it may be best to use a dedicated security address for this though, something like `security@digital.arizona.edu` if possible.

kevcooper: We've decided to merge now and update the email address as a follow up issue

PR #16, opened 2020-01-30
Update security email to az-digital-security@list.ariz…

kevcooper:
This update changes the security email to az-digital-security@list.arizona.edu


trackleft:
https://list.arizona.edu/sympa/info/az-digital-security

PR #39, opened 2020-02-20
Fixes #38: Initial export of akslay's layout builder work

### LINKED ISSUE #38: Capture akslay's layout builder work
## Motivation
We need to capture @akslay's really cool layout builder work in a prototype module (or multiple modules).

## Proposed Resolution
Export config from @akslay's development site on Pantheon

joeparsons:
## Description

Fixes #38.  This is an attempt to capture the proof-of-concept work @akslay has been doing with layout builder & custom blocks to potentially replace paragraphs for Quickstart flexible pages in D8/D9.

The initial export of this configuration was a group exercise in our 2020-02-20 Quickstart 2.0 workshop.  We decided to organize the exported configuration as follows:
- az_blocks: contains exported custom block types and their associated field configuration including layout builder restrictions config
- az_layouts: Contains exported layout builder styles.  This is where we also intend to add custom layout plugins in the future.
- az_flexible_page: Contains the Flexible Page (currently named 'Page') node type with its layout builder config settings.

After the workshop, I added some additional contrib module dependencies (Display Suite, Field Group, Background Image Formatter) and added/renamed a few config entities that were missed to make the exported configuration actually work.  Some patches were also added to resolve various D9 deprecation warnings generated by the new contrib modules (a couple still remain).

There are some outstanding decisions that will need to be made about these changes:
- Do we want to add all of these contrib modules to Quickstart?
  - If we were to add our own custom layout plugins would all of these additional contrib modules be needed?
- If we do want to keep Display Suite, we'll want to find another patch to address the remaining deprecation warnings or consider switching to the 4.x branch of Display Suite which is intended for use with D9 (although doing that would require changing Quickstart to use D9 as well).

## Related Issue

Issue #38 

## How Has This Been Tested?

Minimal testing has been done in the ProboCI build by @joeparsons.

## Types of changes

## Checklist:


akslay:
Couple issues I noticed:
- Can only pick one **Layout Builder Style**; should be able to select multiple
- **Layout Builder Styles** is using the dropdowns instead of the checkboxes, which makes it easier to select multiple styles
- Looks like my **asset injectors** might be missing, which I used to correct some of the UI issues in layout builder and also to specify the appearance of the different components, where needed. Not really a problem for layout builder itself, but we will need those injectors for the components to look and behave properly. Probably want to put the CSS and JS in a module instead of an injector, though.
- **Background Wrapper component** is missing a field for the text inside the wrapper. Field appears to be under the field settings on the block type, but it's not actually showing up when you go to add a background wrapper
- **Text Area component** is also missing its field for the text. Also appears in the field settings, but not viewable when adding the component.
- **Background Image component** is missing the image upload field, similar to the text fields above.
- **Card component** is missing its text field. Same issue as the items above.

In general, it looks like any fields that exist on the block types aren't actually showing up for you to interact with. They exist under the "Manage Fields" section for each component, but don't appear when adding one of the components to a page. Only fields that show up are the "Title" field (default to all inline blocks) and the Layout Builder Styles list.

Issue #45, opened 2020-03-02
Decide whether or not to keep Display Suite


joeparsons:
## What is the problem that we want to solve?
The contrib [Display Suite module](https://www.drupal.org/project/ds) was added as a dependency as part of #38 and #39.  We need to decide whether we want to keep the module as a dependency long term or not.

Display Suite was used for #39 primarily because @akslay preferred the more minimal markup of the Display Suite provided layout plugin templates over the layout plugin templates provided by core out of the box.  It should be possible to create our own layout plugins and templates with whatever markup we want as a replacement if none of the other features provided by the module are needed/used.

As is, the module has added stub configuration to the `third_party_settings` of various other exported configuration entities including entity_view_display settings which we probably don't want to have to maintain if not really needed ([example third party settings](https://github.com/az-digital/az_quickstart/blob/d6de6c775b1e640e253c80024927d73da76d2a2a/modules/custom/az_blocks/config/install/core.entity_view_display.block_content.az_accordion.default.yml#L13-L38)).  We should determine if any of the exported Display Suite third party settings are actually needed/being used for anything.

## Conditions of satisfaction
- [ ] The considered options are documented
- [ ] A recommended next step is documented
- [ ] An estimate of time and effort is provided for the recommendation
- [ ] …

camikazegreen:
The decision was made to not include/remove Display Suite from Quickstart.

PR #180, opened 2020-04-24
Fixes #179 Update DS commit hash

### LINKED ISSUE #179: Composer Display Suite commit hash requirement being ignored
## Problem/Motivation
Our composer requirement of `3.x-dev#f73614254bb7baa311b99886b0393e2ab738d5f7` presently seems to be ignored. This manifested this morning as a failure for a patch to apply because the head of the branch has updated to a point that includes the patch.

### To Reproduce
Perform a composer build of the site.

## Proposed resolution
Update our composer requirements to use the more recent commit and remove the patch.

tadean:
## Description
This morning the composer build process of the profile broke due to the dev version of display suite having been updated to a version that includes a patch we were previously using and the patch consequently failing. Due to the commit referencing feature of composer being fairly bug prone, it's no longer requesting our specified commit hash. I've attempted to fix this by removing the patch in question and updating to a commit hash that included it.

## Related Issue
#179 

## How Has This Been Tested?
Locally. Review of display suite components is needed as this brings in several new display suite commits.

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,7 @@
         "drupal/cas": "1.6.0",
         "drupal/core-recommended": "8.8.5",
         "drush/drush": "9.7.1",
-        "drupal/ds": "3.x-dev#f73614254bb7baa311b99886b0393e2ab738d5f7",
+        "drupal/ds": "3.x-dev#170540060a644102ab3524cd1d31de408d492186",
         "drupal/externalauth": "1.2.0",
         "drupal/field_group": "3.0",
         "drupal/layout_builder_restrictions": "2.5",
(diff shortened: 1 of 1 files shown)


trackleft:
Alternatively we could pin to a tagged release and create a patch with all the improvements, lol.

tadean:
I do like that idea.

joeparsons:
Another possible "solution" to this problem would be removing Display Suite with our own custom layout plugins/templates (#45).

joeparsons:
I do think we should probably merge this sooner, rather than later, though, so we have working builds again and then consider a different long-term solution separately.

Issue #195, opened 2020-05-06
Demo Content Module


tadean:
## Motivation
In conjunction with #172 , it would be nice to be able to import links into the utility links menu to verify that the menu is present. This would also provide some groundwork for the upcoming demo content epic.

## Proposed Resolution
Add a simple `az_demo` module that will provide a storehouse for our migrations. It should handle containing the all of the migrations, looping over them, and running them programmatically.

### Describe the solution you'd like
`az_demo` module that will contain migrations and run them when enabled.

Issue #172, opened 2020-04-22
Add utility links menu for UA Zen parity


tadean:
## Motivation
In D7, we had a Utility Links menu available. We could add that back into our D8 implementation.

### Describe the solution you'd like
Create a menu for the utility links and provide configuration to place the menu block during installation.

![image](https://user-images.githubusercontent.com/51835841/80039690-39abf200-84ad-11ea-8f77-0af65234feca.png)

Issue #33, opened 2020-02-14
Investigate moving dev dependencies and configuration to profile repo


joeparsons:
## What is the problem that we want to solve?
Our dev and test dependencies are currently defined and configured primarily in our [az-quickstart-scaffolding](https://github.com/az-digital/az-quickstart-scaffolding) repo.

As we've continued to work more on Quickstart, we've found that:
- We frequently need to make pull-requests on both the scaffold and profile repos at the same time to make changes to any of our dev/testing tools.
- Keeping various dependencies "in sync" between our scaffold and profile repos is becoming increasingly burdensome.  This is particularly true for dependencies that are tied to specific versions of Drupal core such as [drupal/core-dev-pinned](https://github.com/drupal/core-dev-pinned) and [drupal/core-composer-scaffold](https://github.com/drupal/core-composer-scaffold).

## Proposed resolution
Investigate and experiment with making our scaffolding repo as minimal as possible by moving dependencies from the scaffold repo to the profile repo.

A potential blocker that was brought up by @tadean was that the `zaporylie/composer-drupal-optimizations` composer plugin we are currently using has to be defined/configured in the "root" composer project to function correctly.  This may also be true for `drupal/core-composer-scaffold`.

@tadean and @joeparsons discussed some potential changes/workarounds that would be worth experimenting with:
- Keep composer.json configuration for `drupal/core-composer-scaffold` in scaffold repo but experiment with moving actual dependency to profile repo
- (Or) use less specific version constraint for `drupal/core-composer-scaffold` in the scaffolding repo's composer.json and let it select the version defined by the profile automatically
- `zaporylie/composer-drupal-optimizations` helps a lot, but it wouldn't be the end of the world to not have it. (And maybe we could implement something with a similar in the profile to what it's doing - it's basically just composer trickery that truncates symfony packages.)

## Conditions of satisfaction
- [ ] The considered options are documented
- [ ] A recommended next step is documented
- [ ] An estimate of time and effort is provided for the recommendation

trackleft:
General strategy that might meet our needs:
Replace drupal specific dependencies version constraints in scaffolding repo with less specific versions so that it can rely on quickstart repo for version constraints.

PR #198, opened 2020-05-06
#172 Add Utility Links Menu and Block.

tadean:
This pull request adds a utility links menu and block placement to the configuration for quickstart.

## Description
`az_utility_links` is added as a menu block, and the block is placed above the search box in the `Header 1` region. A template is added to handle the addition of necessary bootstrap 4 classes.

This PR also adds a basic `az_demo` module that imports demo content. This could be fleshed out further when we start in earnest on our demo content epic.

## Related Issue
#172 #195 

## How Has This Been Tested?
Locally.

## Types of changes

## Checklist:

### DIFF
--- a/az_quickstart.info.yml
+++ b/az_quickstart.info.yml
@@ -11,6 +11,7 @@ install:
   - az_cas
   - az_core
   - az_flexible_page
+  - az_demo
   - node
   - history
   - block
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,9 @@
         "drupal/externalauth": "1.2.0",
         "drupal/field_group": "3.0",
         "drupal/layout_builder_restrictions": "2.5",
-        "drupal/layout_builder_styles": "1.0-beta1"
+        "drupal/layout_builder_styles": "1.0-beta1",
+        "drupal/migrate_plus": "5.1",
+        "drupal/migrate_tools": "5.0"
     },
     "extra": {
         "patches": {
--- a/config/install/block.block.az_utility_links.yml
+++ b/config/install/block.block.az_utility_links.yml
@@ -1,0 +1,24 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - system.menu.az-utility-links
+  module:
+    - system
+  theme:
+    - az_barrio
+id: az_utility_links
+theme: az_barrio
+region: header
+weight: -6
+provider: null
+plugin: 'system_menu_block:az-utility-links'
+settings:
+  id: 'system_menu_block:az-utility-links'
+  label: 'Utility Links'
+  provider: system
+  label_display: '0'
+  level: 1
+  depth: 1
+  expand_all_items: false
+visibility: {  }
--- a/config/install/system.menu.az-utility-links.yml
+++ b/config/install/system.menu.az-utility-links.yml
@@ -1,0 +1,7 @@
+langcode: en
+status: true
+dependencies: {  }
+id: az-utility-links
+label: 'Utility Links'
+description: 'The utility links menu contains secondary navigation links that are not part of the main menu.'
+locked: false
--- a/modules/custom/az_demo/az_demo.info.yml
+++ b/modules/custom/az_demo/az_demo.info.yml
@@ -1,0 +1,11 @@
+name: 'Quickstart Demo Content'
+type: module
+description: 'Provides sample content to demonstrate various features of the AZ Quickstart distribution.'
+core_version_requirement: ^8.8 || ^9
+package: 'The University of Arizona'
+dependencies:
+  - az_core
+  - az_flexible_page
+  - migrate
+  - migrate_plus
+  - migrate_tools
--- a/modules/custom/az_demo/az_demo.install
+++ b/modules/custom/az_demo/az_demo.install
@@ -12,7 +12,7 @@ use Drupal\migrate\MigrateMessage;
  * Implements hook_install().
  */
 function az_demo_install() {
-  $tag = 'az_demo_content';
+  $tag = 'Quickstart Demo Content';
 
   // Run the migrations that are tagged as demo content.
   $migrations = \Drupal::service('plugin.manager.migration')->createInstancesByTag($tag);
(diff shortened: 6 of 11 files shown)


tadean:
Additional work is needed on this before merging, to programmatically add some demo menu links:

- [ ] Add placeholder `az_demo` module
- [ ] Add simple migration for menu links

tadean:
To finish off work on this, it was desirable that some utility links actually be populated; however, in Drupal 8 menu links represent content rather than configuration entities (in D7 this sort of thing was imported as part of our demo content). We haven't actually started off on our epic regarding demo content yet, but given the need of this particular ticket I hoped to lay a little future groundwork for demo content imports.

I've added the following changes:

- Addition of an **az_demo** module that contains a simple framework for the programmatic importing of migrations.
- When **az_demo** is enabled, it will attempt to run all migrations tagged with the `az_demo_content` tag. This is in hopes of laying groundwork for future demo content migrations.
- Added an `az_demo_menu_links` migration that creates menu links from JSON
- `az_demo_menu_links` handles the following use cases:
    - internal paths and external URLS
    - setting expanded menu links
    - nesting menu links under parents, via `menu_link_parent`
    - weights to order links
    - targeting different menus by machine name
    - lookup of menu link destination by **URL alias**, for internal paths.
- Added one test to verify the successful import of the utility links.

I'm hopeful the above will offer enough flexibility to handle some future more complicated menu entries like the main navigation demo content

tadean:
Result of successful import of demo utility links:

image

joeparsons:
> I'm hopeful the above will offer enough flexibility to handle some future more complicated menu entries like the main navigation demo content

This all looks great to me.  Thanks @tadean!

Looks like `migrate_plus` provides the Url migrate source plugin as well as the data_fetcher and data_parser plugins, correct?

Is `migrate_tools` just needed so that these migrations could be imported with Drush (if desired)? 

Do you think it would make sense to also add the `Content` tag to our migrations per the recommendation in this Drupal Core change record?
https://www.drupal.org/node/2944527
> Contributed modules that provide migrations should categorize all migrations to either 'Content' or 'Configuration' using migrate_tags. The following example has two migration_tags, 'Drupal 6' and the new 'Configuration' tag.

I'm thinking probably not since or demo content migrations are kind of more special purpose but I guess I'm not sure.

tadean:
Correct, `migrate_plus` essentially includes the necessary plugins for allowing the import of the file-based JSON content. My reasoning for including `migrate_tools` was for providing the drush commands necessary for developing demo content migrations since it's otherwise a little unwieldly to work with them.

It's worth noting that the two included versions were for Drupal 9 compatibility (the stable 4.x versions have `drupal-check` errors)

I think it would be reasonable to include a `content` tag. There might be some potential that it could conflict with migrations users are doing on their own sites that are their actual content - but maybe that won't matter since presumably the demo content module would be disabled in that case.

joeparsons:
I'm good with whatever you think is best regarding the `Content` tag.  FWIW, it looks like the examples on the Drupal core change record are all capitalized though.

tadean:
Thanks! Do you think we should adopt a similar convention in our own tag name, eg. something like "AZ Demo Content" as the tag name, rather than the machine-name like `az_demo_content` that it currently has?

joeparsons:
I guess that might be best 🤷‍♂️.  I probably would have done like you and used a "machine" name for tags, too, if I hadn't seen the examples on that change record.

tadean:
Added the `Content` tag to the `az_demo_menu_links` migration, and swapped the existing `az_demo_content` machine-name-style tag to the human-readable `Quickstart Demo Content` tag to follow the Drupal.org convention.

composer.json
```diff
@@ -32,7 +32,9 @@
         "drupal/externalauth": "1.2.0",
         "drupal/field_group": "3.0",
         "drupal/layout_builder_restrictions": "2.5",
-        "drupal/layout_builder_styles": "1.0-beta1"
+        "drupal/layout_builder_styles": "1.0-beta1",
+        "drupal/migrate_plus": "5.1",
```
trackleft, line 36: can these be require-dev?

az_quickstart.info.yml
```diff
@@ -11,6 +11,7 @@ install:
   - az_cas
   - az_core
   - az_flexible_page
+  - az_demo
```
trackleft, line 14: Can `az_demo` be optional?

trackleft: Here is a reference page that didn't help me: https://www.drupal.org/docs/8/distributions/creating-distributions/how-to-write-a-drupal-8-installation-profile#s-infoyml-file

composer.json
tadean: `migrate_plus` is a required dependency as it contains the plugins to import from a serialized URL/file and can't be optional.

`migrate_tools` could be a dev dependency, but composer only installs dev dependencies for the root package (which in our case is the scaffolding, not this installation profile). This is one reason we don't currently have dev dependencies in this composer.json. That's something we're looking into as part of #33 .

az_quickstart.info.yml
tadean: I think this list in `az_quickstart.info.yml` isn't a required list of modules like in module dependencies, and only ones that are first installed by the installation profile. It seemed to allow me to disable the module (which rolls back the content when `hook_uninstall()` fires).

Is it preferred that the demo content be not enabled by default?

trackleft: Ok, really just wanted to be able to disable it,sorry.

composer.json
trackleft: Oh,makes sense

camikazegreen:
I think that these would be worth addressing in a followup, but there are some design differences between the D7 and D8 versions. 
D7:
- font-weight: 400
- :hover {text-decoration: underline}
D8:
- font-weight: 700 (I actually recommend font-weight:500, since the text-muted color is slightly lighter than the gray used in D7)
- :hover {text-decoration: none} (We will need _some_ form of link indicator)

PR #200, opened 2020-05-07
Add admin user credentials to contributing doc

### LINKED ISSUE #199: Update admin Drupal user passwords to match usernames
## Motivation
Probo, Lando, and DDEV all create Drupal users with different passwords for no real reason, so updating them to match the user names will make development easier.

## Proposed Resolution
Update the passwords to be the same as the usernames. Also, put the credentials in the contributing documentation so it is easier to find than looking through the Lando/DDEVProbo config files.

jwmoore1:
Resolves #199

## Types of changes

## Checklist:

- [ ] I've added this Pull Request to the AZ Quickstart project in the right column.


joshuasosa:
Curious, is there a purpose why the pw is different across the different installs (lando, ddev, probo)? Any reason not to just make it admin / admin?

trackleft:
> Curious, is there a purpose why the pw is different across the different installs (lando, ddev, probo)? Any reason not to just make it admin / admin?

Changing them to be uniform makes total sense to me

jwmoore1:
@joshuasosa I'm assuming it's just because that was how it was setup. I agree, making them all the same would be great!

tadean:
I believe it was patterned after the credentials used in D7 probo/localdev. I'd also be in favor of making them all identical.

jwmoore1:
Do we want to keep azadmin, or just make it admin/admin? Also, do we care about the test user accounts created by Probo? Is it OK to make the password changes as part of this PR and just change the issue?

joeparsons:
I'd also be OK with making the credentials the same across Probo/lando/ddev but using `azadmin` (and originally `uaqs_admin`) instead of a generic `admin` username was definitely an intentional choice after some discussion in a Friday meeting a long time ago.

joeparsons:
I'd also be OK with using the username as the password for all of the ProboCI accounts (azadmin:azadmin, azcontentadmin:azcontentadmin, azcontenteditor:azcontenteditor) but would probably prefer to keep `azadmin` as the username over `admin`.

Issue #174, opened 2020-04-22
AZ Barrio PHPCS Cleanup


tadean:
## Conditions of satisfaction
`az_barrio` presently has a few warnings from `phpcs` that are being thrown, though none is severe enough to actually cause an error. It would be good to resolve these existing warnings (the warnings involved are largely use of `==` over `===` and line length warnings)

joshuasosa:
This issue wasn't completely resolved with #19 

These warnings still exist:

> 5FILE: ...les/custom/az_quickstart/themes/custom/az_barrio/az_barrio.theme
6----------------------------------------------------------------------
7FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 1 LINE
8----------------------------------------------------------------------
9 8 | WARNING | Possible RFI detected with dirname on include_once
10 8 | WARNING | Filesystem function dirname() detected with dynamic
11   |         | parameter
12 8 | WARNING | Possible RFI detected with __FILE__ on include_once
13----------------------------------------------------------------------
14
15
16FILE: .../custom/az_quickstart/themes/custom/az_barrio/theme-settings.php
17----------------------------------------------------------------------
18FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 1 LINE
19----------------------------------------------------------------------
20 10 | WARNING | Possible RFI detected with dirname on include_once
21 10 | WARNING | Filesystem function dirname() detected with dynamic
22    |         | parameter
23 10 | WARNING | Possible RFI detected with __FILE__ on include_once
24----------------------------------------------------------------------
25
26
27FILE: ...ckstart/modules/custom/az_core/src/QuickstartConfigInstaller.php
28----------------------------------------------------------------------
29FOUND 0 ERRORS AND 3 WARNINGS AFFECTING 3 LINES
30----------------------------------------------------------------------
31  33 | WARNING | Filesystem function is_dir() detected with dynamic
32     |         | parameter
33 102 | WARNING | Filesystem function file_exists() detected with
34     |         | dynamic parameter
35 104 | WARNING | Filesystem function file_get_contents() detected
36     |         | with dynamic parameter
37----------------------------------------------------------------------

davearch:
Hello can I take a stab at this issue? Also what is the specific command I should use to test everything?

I ran `lando phpcs --standard=Drupal --extensions=php,module,inc,install,test,profile,theme,css,info,txt,md,yml web/profiles/custom/az_quickstart/themes/custom/az_barrio`

[...]

PR #222, opened 2020-06-07
removed some RFI and dynamic parameter warnings from PHPCS

davearch:
## Description

The changes I made removed most but not all PHPCS warnings. The only warnings that were left over were 'possible RFI detected' by the use of `drupal_get_path()` in the az_barrio.theme file and the theme-settings file and 'filesystem function with dynamic parameter' on the use of `file_get_contents()` in the QuickstartConfigInstaller.php file. However I would argue that in keeping with the spirit of the security measures that PHPCS is helping us mitigate, these warnings should be acceptable. For one, the code sniffer tool is by its own admission a generator of 'false positives'. Please refer to their [README](https://github.com/FloeDesignTechnologies/phpcs-security-audit#annoyances) as well as this interesting issue [conversation](https://github.com/FloeDesignTechnologies/phpcs-security-audit/issues/69#issuecomment-599313528). Secondly, seeing as how a Remote File Inclusion attack relies on user input I would think that the simple `drupal_get_path() . '/includes/common.php';` would not be at a high risk. And with `file_get_contents()`, I simply could not figure out a way around it but to be fair it also relies on basically static input. The Drupal codebase itself is littered with this function also.

## Related Issue

#174

## How Has This Been Tested?

I ran lando phpcs and lando phpunit.

## Types of changes

- [x ] Bug fix (non-breaking change which fixes an issue)

## Checklist:

- [ ] I've added this Pull Request to the AZ Quickstart project in the right column. (not sure what this means)
- [x ] My code follows the code style of this project.
- [x ] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes. (I could write tests but I am not sure where to put them)

### DIFF
--- a/modules/custom/az_core/src/QuickstartConfigInstaller.php
+++ b/modules/custom/az_core/src/QuickstartConfigInstaller.php
@@ -30,7 +30,7 @@ class QuickstartConfigInstaller extends ConfigInstaller {
     }
     $config_install_path = $this
       ->getDefaultConfigDirectory($type, $name);
-    if (!is_dir($config_install_path)) {
+    if (stream_resolve_include_path($config_install_path) === FALSE) {
       return;
     }
     $storage = new FileStorage($config_install_path, StorageInterface::DEFAULT_COLLECTION);
@@ -99,7 +99,7 @@ class QuickstartConfigInstaller extends ConfigInstaller {
     $app_root = \Drupal::root();
     $filename = $app_root . '/' . drupal_get_path($type, $name) . '/' . $name . '.az_config_overrides.yml';
 
-    if (file_exists($filename)) {
+    if (stream_resolve_include_path($filename)) {
       // If an override file exists, parse it for overrides.
       $yaml = Yaml::decode(file_get_contents($filename));
       if (!empty($yaml['overrides'])) {
--- a/themes/custom/az_barrio/az_barrio.theme
+++ b/themes/custom/az_barrio/az_barrio.theme
@@ -5,7 +5,7 @@
  * Functions to support theming in the Arizona Barrio theme.
  */
 
-include_once dirname(__FILE__) . '/includes/common.inc';
+include_once drupal_get_path('theme', 'az_barrio') . '/includes/common.inc';
 
 use Drupal\Core\Link;
 use Drupal\Core\Render\Markup;
--- a/themes/custom/az_barrio/theme-settings.php
+++ b/themes/custom/az_barrio/theme-settings.php
@@ -7,7 +7,7 @@
  * Provides theme settings for Arizona Barrio.
  */
 
-include_once dirname(__FILE__) . '/includes/common.inc';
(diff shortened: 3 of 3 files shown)


davearch:
I've clearly done something wrong but I am not able to see the details of the CI checks.

davearch:
I've clearly done something wrong but I am not able to see the details of the CI checks.

joeparsons:
@davearch Thanks for working on this.  We have the main CI tool we use (ProboCI) configured to work with dev branches on the main az-digital/az_quickstart repo but not forks (there are some hard-coded references to the main repo in the `.probo.yaml` file).

FWIW, it actually looks like ProboCI got farther along with building your PR than I would have expected so maybe we could look into to tweaking our .probo.yaml config file to better support PRs from forks.

tadean:
> I've clearly done something wrong but I am not able to see the details of the CI checks.

@davearch I don't see any immediate issues - I think the issue with CI checks might be because our CI script may make some incorrect assumptions about where the branch lives (I think it currently expects that the branch lives in the az_quickstart repo). We may need to make a few tweaks to allow for it to operate correctly with a fork. I'll check this out today.

tadean:
To get around the update delay on packagist, we'd put in a fix a while back where the repository itself is set up as a direct vcs repository rather than using packagist. I think we might potentially be able to leverage this to support forks, if probo has access to the forked location.

joeparsons:
@tadean It'd be cool if we could get that working but I'm afraid the info/variablesprovided by ProboCI might not give us all the context needed (at least not enough to make it very easy):
https://docs.probo.ci/build/environment-variables/

tadean:
That's true :frowning_face:  It looks like it might be missing some of the needed information. I'm wondering if maybe there isn't really any reason why the VCS repo needs to be one that's web-hosted...potentially we could set it up so that it treats the local copy of the repo inside the probo build as the package repository so we can disregard what account it's actually hosted on.

joeparsons:
Honestly I'm surprised ProboCI worked at all with this PR based on the experience we had early on when we first started using ProboCI.  We explored the idea of supporting PRs from forks and ran into some pretty big issues mainly having to do with our ProboCI subscription being tied to the main repo and ProboCI refusing to even do anything on repos without an associated subscription.  It's possible that either ProboCI has changed the way things work somewhat since then or that the ProboCI integration with GitHub works different than the integration with BitBucket so I'd definitely be open to exploring this fork PR workflow again.

joeparsons:
@jwmoore1 Suggested that it may be possible to use special code comments to basically whitelist certain lines in our code for certain PHPCS security checks.  We discussed this in a meeting last week and agreed that we'd prefer to pursue that route instead of changing many of the affected lines of code.

davearch:
Hey @joeparsons thanks for the update! @jwmoore1 is the man! Let me know if there is anything else I can do to support.

joeparsons:
We decided to merge #239 instead.

PR #248, opened 2020-07-14
Fixes #247: Improve compatibility with latest ddev release + backwards-compatible D9 ddev config changes.

### LINKED ISSUE #247: Quickstart ddev config not working with latest ddev version
## Problem/Motivation
Our Quickstart-packaged ddev configuration does not work when using the latest version of ddev (1.15.0).

### Describe the bug
Something seems to have changed in the way ddev works such that our ddev configuration no longer works as expected.  Specifically the custom `ddev install` command fails with the following errors:
```
fatal: destination path '/var/www/html' already exists and is not an empty directory.
                                                                 
  [RuntimeException]                                               
  File "./composer.json" cannot be found in the current directory  
                                                                   
config [-g|--global] [-e|--editor] [-a|--auth] [--unset] [-l|--list] [-f|--file FILE] [--absolute] [--] [<setting-key>] [<setting-value>]...

Composer could not find a composer.json file in /var/www/html
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
Command site-install needs a higher bootstrap level to run - you will need to invoke drush from a more functional Drupal environment to run this command.                                                    [error]
The drush command 'site:install' could not be executed.                                                                                                                                                      [error]
```
### To Reproduce

1. Install ddev version 1.15.0 locally
2. Clone az_quickstart
3. Run the following commands within the cloned working copy of the repo:
```
cp .ddev/config.quickstart.yaml .ddev/config.yaml
ddev start
ddev install
```
4. See errors

## Proposed resolution
Look into moving the `composer install`, and `yarn install` steps out of our custom ddev `install` command script and adding them as `exec` commands in the `post-start` hook in the ddev configuration file (see https://ddev.readthedocs.io/en/stable/users/extending-commands/#supported-command-hooks)

joeparsons:
## Description
- ~Moves compser and yarn script steps out of custom ddev `install` command into ddev `post-start` hook~
- ~Replaces step that git clones scaffolding repo with git init, git fetch, and git checkout commands due to weirdness with `/var/www/html` directory already existing~
- Replaces existing command to remove files in the `/var/www/html` directory with `rm -rf /var/www/html/*` to fix problems with latest ddev release
- Bumps PHP + MariaDB versions in ddev config to improve D9 compatibility (while preserving backwards compatibility)

## Related Issue
#247 

## How Has This Been Tested?
Locally with ddev version 1.15.0

## Types of changes

## Checklist:

- [ ] I've added this Pull Request to the AZ Quickstart project in the right column.

### DIFF
--- a/.ddev/commands/web/install
+++ b/.ddev/commands/web/install
@@ -1,7 +1,13 @@
 #!/bin/sh
 
-## Description: Install Drupal using the Quickstart install profile.
+## Description: Create a new quickstart project and use the installation profile as a non-packagist repository using the matching scaffolding branch if available.
 ## Usage: install
 ## Example: "ddev install"
 
+rm -rf /var/www/html/*
+BRANCH_NAME=$(git -C /usr/local/quickstart-install-profile rev-parse --abbrev-ref HEAD); if [ $(git ls-remote --heads https://github.com/az-digital/az-quickstart-scaffolding.git $BRANCH_NAME | wc -l) = 1 ]; then SCAFFOLD_BRANCH="$BRANCH_NAME"; else SCAFFOLD_BRANCH="master"; fi; git clone --branch $SCAFFOLD_BRANCH https://github.com/az-digital/az-quickstart-scaffolding.git /var/www/html
+composer config repositories.localdev path /usr/local/quickstart-install-profile && composer require --no-update drupal/core-recommended:* zaporylie/composer-drupal-optimizations:* az-digital/az_quickstart:\*@dev
+composer install
 drush --root=/var/www/html/web site:install --account-mail=noreply@email.arizona.edu --account-name=azadmin --account-pass=azadmin --db-url=mysql://db:db@db:3306/db -y --verbose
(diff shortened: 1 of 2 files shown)


joeparsons:
This also contains some backwards-compatible D9 changes.

joeparsons:
Thanks for the explanation @joshuasosa.  I'd be OK with keeping the composer install steps in a custom script/command but I think I might still prefer having it separate from the Drupal site install steps.  Would the best way to accomplish that be to add another script/command for the composer install steps?

joshuasosa:
Yes, a `ddev build` command could be added and used before `ddev install`.

.ddev/commands/web/install
```diff
@@ -4,7 +4,7 @@
 ## Usage: install
 ## Example: "ddev install"
 
-rm -rf /var/www/html/docroot && rm /var/www/html/index.html
+rm -rf /var/www/html/*
```
kevcooper, line 7: I believe this syntax will miss files that start with `.`

PR #257, opened 2020-07-31
Fixes #194 Add Breadcrumb Settings

### LINKED ISSUE #194: Add UA Zen breadcrumb features to AZ Quickstart
## Motivation
The breadcumb block in AZ Quickstart works but is missing some configuration options like showing/hiding the home page and showing/hiding the content title.

Ability to show/hide home page part and show/hide current page part in breadcrumbs.

Instead of including custom breadcumb functionality in AZ Barrio, consider using a contrib module like [Easy Breadcrumb](https://www.drupal.org/project/easy_breadcrumb). Using a contrib module would negate the need to replicate ua_zen_preprocess_breadcrumb and ua_zen_breadcrumb.

- Administrator needs to be able to configure site-wide breadcrumb settings.

## Additional context
In the below examples, a hierarchy of Home > Test > Test 2 exists.

Default breadcrumbs show home page with no option to disable home page.
default_breadcumb

Easy breadcrumbs allows easy disabling of home page.
easy_breadcrumb

tadean:
This pull request adds breadcrumb settings to `az_quickstart` that mimic the settings that were in place in `ua_zen` @joshuasosa suggested in the original ticket that this could be done using the contrib `easy_breadcrumb` module which is the approach taken in this PR.

## Description
This PR adds the `easy_breadcrumb` module, incorporates it into the install profile, provides default settings for it, and also provides a template override for the barrio breadcrumbs to apply the necessary aria labels, some of which weren't present in the original barrio theme.

By default, the **Home** breadcrumb entry isn't shown, and the "leaf" current breadcrumb showing the page isn't shown, as it's right next to the page title. If the leaf breadcrumb is enabled later, it has the proper aria property.

## Related Issue
#194 

## How Has This Been Tested?
Create nested menu pages and see if

## Types of changes

## Checklist:

### DIFF
--- a/az_quickstart.info.yml
+++ b/az_quickstart.info.yml
@@ -31,6 +31,7 @@ install:
   - options
   - path
   - pathauto
+  - easy_breadcrumb
   - page_cache
   - dynamic_page_cache
   - big_pipe
--- a/composer.json
+++ b/composer.json
@@ -29,6 +29,7 @@
         "drupal/core-recommended": "8.9.2",
         "drush/drush": "9.7.1",
         "drupal/ds": "3.x-dev#170540060a644102ab3524cd1d31de408d492186",
+        "drupal/easy_breadcrumb": "1.13",
         "drupal/externalauth": "1.3",
         "drupal/field_group": "3.1",
         "drupal/layout_builder_restrictions": "2.7",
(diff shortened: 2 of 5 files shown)


tadean:
**Current page disabled:**
image

**Current page enabled (showing aria attribute):**
image

tadean:
The following default configuration was removed from the configuration provided by the `easy_breadcrumb` module when it was copied to the profile:

```
dependencies:
  module:
    - easy_breadcrumb
  enforced:
    module:
      - easy_breadcrumb
```

This seemed to prevent the install profile from actually using the configuration. This should be looked into prior to merging. It seemed to work fine for the module itself, but was causing the profile to not use the configuration. I'm not sure if it's normal for a module to declare dependency on itself like this and maybe this is confusing the profile's install process.

- [ ] Investigate configuration dependency issue

Issue #263, opened 2020-08-13
Create testable prototype of Layout Builder flexible page.


camikazegreen:
## Conditions of satisfaction
- [ ] Should be ready for Shoshana to use in User Testing
- [ ] Should be in a stable environment (Not Probo)
- [ ] The page should have similar page elements as the front page of kitten: https://kitten.ltrr.arizona.edu/
- [ ] Should have enough page elements to be able to User Test the reordering functionality.

joeparsons:
Changes that were needed to sandbox Quickstart site on Pantheon to fix various layout builder styling issues (we need to make these improvements in Quickstart):

- Additional CSS styles copied from @akslay's D8 Registrar's Office sandbox site (added via Asset Injector module for now)
- Changed the display settings for the background image block type so it’s targeting “background_image” and not “background_wrapper”
- Fixed the layout builder style for the “card” class (removed “az” from the class)

PR #266, opened 2020-08-14
#263: Various changes that were required to build Layout Builder test site.

joeparsons:
## Description
@akslay Did most of the work for this.  I just exported the updated configs, etc.  These are changes that we needed to make to the prototype Layout Builder Sandbox site that was setup for the user test to resolve various styling issues and other problems.  @akslay also added an image field to the Card block type.

## Related Issue
#263 

## How Has This Been Tested?
Locally with lando.

## Types of changes

## Checklist:


modules/custom/az_blocks/css/az_blocks_az_card_body.css
```diff
@@ -7,3 +7,23 @@
   */
   padding: 1.5rem;
 }
+
```
trackleft, line 10: Looks like we might need to add a template to make this unnecessary.

Issue #305, opened 2020-09-09
Update Text paragraph


danahertzberg:
**Description:** Update and breakout styles, etc.

**Requirements:** 
- Undo/Redo
- Styles (Normal text, Page title, Heading 2-5)
- Font size (Small, Normal, Medium, Large, Extra-large)
- Capitalize/Lowercase/Uppercase
- Bold/Italic/Strikethrough
- Text color (brand colors only)
- Highlight color (brand colors only - transparent?)
- Link/unlink
- Button
- Align left/center/right
- Insert unordered/ordered list
- Increase/decrease indentation
- Block quote
- Insert table
- Embed media/blocks/content
- Super/sub script
- Insert special characters
- Remove formatting
- Paste plain text
- View source

**Notes:**
- Ideally, this is only for text content and media content can be added via the Media paragraph. Not sure how migration would work though.
-- Keep media embed functionality
- Do we need a view source feature?
-- Yes
- Create paragraph type for grid using Layout Paragraphs experimental module
- Allow table and grid markup

_**Dependent on AZ Digital's review/edits and approval these requirements**_

PR #324, opened 2020-09-15
Fixes #312 Add az_standard text format.

### LINKED ISSUE #312: Create Basic Text Format
Requirements:
Clone an existing Text Format that can be referenced by Paragraph types
Name: az_standard
Display name: Standard

tadean:
This Pull request adds an `az_standard` content type to serve as a placeholder that can be referenced by other paragraphs.

## Description
This PR changes the existing `az_advanced_text` text format into the `az_standard` text format and allows content editors to use the format.

This PR also adds the `entity_embed` module and provides some placeholder settings for allowing the embedding of pages and blocks in an anticipation of the requirements for #305 

Revisions were made to the configuration originally from `az_advanced_text` further allowing the embedding of media entities using the media library. (and the associated necessary allowed HTML elements).

`View media` permission was added to the content editors as this was causing the media library to not work for them.

## Related Issue
#312 

## How Has This Been Tested?
Verify that `Standard` text format exists.
Create a page with placeholder text.
Create a media entity (eg. image, etc)
Create a second page.
Verify that content editor role can embed media entities, pages, and blocks in the paragraph.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:


tadean:
Added permission for anonymous users to see published media entities. Otherwise, embedded images were not being rendered for anonymous users when viewing pages (due to not having access to see the media entities).

tadean:
Probo builds don't currently seem to result in `files` folder permissions that allow generated images to be created.

image

tadean:
image

danahertzberg:
I tested the following functionality and it all works. There are a lot of usability changes that could be added, however. I assume these should be separate tickets (media overlay location, media type select, text format settings, page editing experience).

- Verify that Standard text format exists.
- Create a page with placeholder text.
- Create a media entity (eg. image, etc)
- Verify that content editor role can embed media entities, pages, and blocks in the paragraph.

modules/custom/az_paragraphs/az_paragraphs.info.yml
```diff
@@ -4,4 +4,7 @@ description: 'Provides custom paragraph types.'
 core_version_requirement: ^8.8 || ^9
 package: 'The University of Arizona'
 dependencies:
+  - az_flexible_page
```
joeparsons, line 7: Just curious: is there a specific reason `az_flexible_page` needs to be a dependency here?  Looks like `az_paragraphs` already a dependency for `az_flexible_page` so I'm not sure we want/need the circular dependency (?)

tadean: I believe I put this due attempting to resolve a configuration dependency issue in the provided configuration of the entity embed to pages (this module provides the entity_embed configuration to pages and so pages have to exist).

However now that I think about it, this dependency here failed to resolve the entity embed configuration dependency issue (it was still complaining at install time that the configuration didn't exist) and it was only resolved by making the embed configuration optional rather than install.

Good catch on this being a circular dependency...since this dependency didn't seem to actually resolve the config dependency issue for the page embed maybe this can just be removed, since the actual fix seemed to be flagging the entity_embed configuration as optional so it can be installed after the fact.

tadean: It looks like it works fine without this. I think it'd be good to remove it.

tadean:
Updated to remove unnecessary circular dependency.

tadean:
Retested with dependency removed:

image

mmunro-ltrr (approved):

joeparsons (approved):

Issue #331, opened 2020-09-25
Paragraph Styles Architecture Plan


mschmeling14:
Determine if child elements on paragraphs are implemented with child paragraphs or compound fields.
Update our Contributing docs with our recommendations on Paragraph behavior plugins, view modes, etc.

PR #334, opened 2020-09-30
Default Paragraph Behavior

tadean:
This PR adds a default paragraph behavior plugin to serve as the base for future modification.

## Description
This pull request adds a default paragraph behavior plugin (**AZDefaultParagraphBehavior**) to serve as a base class for paragraph styles. This is intended to contain the global paragraph styles that are potentially available to all paragraphs.
Currently it handles adding a default bottom spacing element to paragraphs.

Previous functionality that handled loading paragraph-specific css and js libraries has been refactored and placed into
this plugin as well.

The new behavior is enabled on the existing **az_text** paragraph.

The existing paragraph test is altered to check that the plugin is working.

Contributing documentation is updated.

## Related Issue
#331 

## How Has This Been Tested?
Create a text paragraph. Verify that the Additional Options detail pane exists and contains the Bottom Spacing dropdown.Verify that the selected spacing element is added to the paragraph in rendering.

## Types of changes

## Checklist:


tadean:
image

image

joeparsons (approved):

mmunro-ltrr (approved):
After messing up the live demo, it's only polite to approve this!

jwmoore1 (approved):

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZDefaultParagraphBehavior.php
```diff
@@ -0,0 +1,154 @@
...
+   * Drupal\Core\Asset\LibraryDiscoveryInterface definition.
+   *
+   * @var \Drupal\Core\Asset\LibraryDiscoveryInterface
+   */
+  protected $libraryDiscovery;
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
```
joeparsons, line 48: I think the preferred way of injecting additional dependencies (beyond what the parent class does is to actually override the `__construct()` method with additional typeHinted dependencies in the method signature, and then inject all of the dependencies (both those defined by the parent and the additional ones defined in the extended class) in the `create()` method.  We have an example of this in the az_quickstart_core module here:
https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_core/src/Form/QuickstartCoreSettingsForm.php#L31-L57

joeparsons (changes requested):

camikazegreen:
In the help text for the Bottom Spacing, can we link directly to: https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/spacing

joeparsons:
This looks AMAZING overall!

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZDefaultParagraphBehavior.php
```diff
@@ -0,0 +1,154 @@
...
+ * Provides a base behavior for Quickstart Paragraph Defaults.
+ *
+ * @ParagraphsBehavior(
+ *   id = "az_default_paragraph_behavior",
+ *   label = @Translation("Default Quickstart Paragraph Behavior"),
+ *   description = @Translation("Provides default Quickstart paragraph settings."),
+ *   weight = 0
+ * )
+ */
+class AZDefaultParagraphBehavior extends ParagraphsBehaviorBase {
```
joeparsons, line 22: Nitpick:  Would it be better to name our behavior `AZDefaultParagraphsBehavior` (with an `s`).  It seems that within the Paragraphs module, behaviors are always referred to as Paragraphs Behaviors.

trackleft (approved):
Magic!

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZDefaultParagraphBehavior.php
joeparsons: @tadean This documentation on drupal.org provides some similar examples: https://www.drupal.org/docs/8/api/services-and-dependency-injection/dependency-injection-for-a-form

tadean: Thanks @joeparsons ! Updated to the style of overriding that overrides the constructor.

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZDefaultParagraphBehavior.php
tadean: I agree, that seems better. Updated to change the name to `AZDefaultParagraphsBehavior`.

tadean:
Updated with the following:

- Remove `az_` scoping on FAPI elements inside the `az_display_settings` element.
- Update style of container dependency injection.
- Remove a bare call to `t()`
- Added link to spacing documentation
- Changed name of class to `AZDefaultParagraphsBehavior`

joeparsons (approved):

trackleft (approved):

PR #356, opened 2020-10-26
Update composer.json with minimum version numbers that are compatible with composer 2.0

### LINKED ISSUE #349: Composer 2.0
We need to investigate how ready we are to move to Composer 2.0, which will be released soon.

Packagist has started triggering the following warnings in the output of composer commands run using version 1.x:
```
Warning from https://repo.packagist.org: You are using an outdated version of Composer. Composer 2.0 is about to be released and the older 1.x releases will self-update directly to it once it is released. To avoid surprises update now to the latest 1.x version which will prompt you before self-updating to 2.x.
```

Upgrade guide:
https://getcomposer.org/upgrade/UPGRADE-2.0.md

trackleft:
Closes #349

### DIFF
--- a/.probo.yaml
+++ b/.probo.yaml
@@ -6,7 +6,6 @@ steps:
     plugin: Script
     script:
       - composer self-update
-      - composer global require hirak/prestissimo
       - if [ $(git ls-remote --heads https://github.com/az-digital/az-quickstart-scaffolding.git $BRANCH_NAME | wc -l) = 1 ]; then SCAFFOLD_BRANCH="$BRANCH_NAME"; else SCAFFOLD_BRANCH=main; fi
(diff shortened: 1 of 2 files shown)


joeparsons (approved):

tadean (dismissed):

joeparsons:
Should we also remove this line from the probo config for this repo, too?
https://github.com/az-digital/az_quickstart/blob/main/.probo.yaml#L9

tadean:
> Should we also remove this line from the probo config for this repo, too?
> https://github.com/az-digital/az_quickstart/blob/main/.probo.yaml#L9

Yes, I think so.

joeparsons (approved):

tadean (approved):
Well done!

Issue #292, opened 2020-09-04
(Needs review) Create View paragraph


danahertzberg:
**Name:** View
**Description:** Add a pre-defined view of other content types, such as people, events, news.

**Requirements:** 
- View selection should show only views available to the end-user (no admin views)
- View selection should show only human-readable name
- Include arguments

**Notes:**
MVP 2 could include upgrading the argument field to be a selection of taxonomy terms rather than needing an ID

_**Dependent on AZ Digital's review/edits and approval these requirements**_

smayden:
Looks good

joeparsons:
I think we should consider this module instead of the `viewfield` module we use in D7 for adding views to Paragraphs:
https://www.drupal.org/project/viewsreference

camikazegreen:
Looks good to me.

Issue #72, opened 2020-04-03
Figure out ChromeDriver dependencies for tooling stack


joeparsons:
## What is the problem that we want to solve?
We need to be able to run D8 Javascript Functional Tests both locally and on our CI tools.

We should look into what Drupal.org is doing for this and also at what tools are already installed/available on ProboCI's docker images.

UC Davis' SiteFarm project has Functional Javascript tests and they use ProboCI.

## Conditions of satisfaction
- [ ] Test dependencies are documented
- [ ] Lando/ddev support is considered

camikazegreen:
Troy: That issue has not been resolved yet. The views PR linked it because a unit test couldn't be written for the Views paragraph because it would necessitate being a javascript test and we didn't have chromedriver yet in our tool stack.

joeparsons:
Drupal.org Documentation Re: Running PHPUnit Javascript tests:
https://www.drupal.org/docs/automated-testing/phpunit-in-drupal/running-phpunit-javascript-tests

Docker Image used by Drupal.org test runners:
https://hub.docker.com/r/drupalci/webdriver-chromedriver/

PR #359, opened 2020-10-28
Fixes 292: Add Views Paragraph

tadean:
This pull request adds the `viewsreference` module, and an associated paragraph bundle for creating views references.

## Description
This pull request adds the `viewsreference` module, and adds a related `az_paragraphs_view` module that is responsible for creating a View paragraph. A template is also added that causes the view title to be rendered as an h2, eg. as a block title. By default it was rendered as a div. The module is set to be enabled by default in the install profile.

## Related Issue
#292 

## How Has This Been Tested?

1. Verify that the Views paragraph exists (`az_view_reference`)
2. Create content.
3. Add a paragraph using the Add View button.
4. Select one of the default system views for testing purposes.
5. Save the content.
6. Verify that the view is rendered in the resultant page.

## Types of changes

## Checklist:

### DIFF
--- a/az_quickstart.info.yml
+++ b/az_quickstart.info.yml
@@ -11,6 +11,7 @@ install:
   - az_cas
   - az_paragraphs
   - az_paragraphs_text
+  - az_paragraphs_view
   - az_core
   - az_flexible_page
   - az_demo
--- a/composer.json
+++ b/composer.json
@@ -39,7 +39,8 @@
         "drupal/token": "1.7",
         "drupal/paragraphs": "1.12",
         "drupal/entity_embed": "1.1",
-        "drupal/embed": "1.4"
+        "drupal/embed": "1.4",
+        "drupal/viewsreference": "2.0-beta2"
     },
     "require-dev": {
         "az-digital/az-quickstart-dev": "~1"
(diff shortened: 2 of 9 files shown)


trackleft:
This is awesome https://take.ms/lspPY

tadean:
image

image

trackleft (approved):

tadean:
Updated per feedback to include only the Argument extra option setting for the referenced view:

image

trackleft (approved):

camikazegreen:
This looks really good so far. 
The one thing that might be worth discussing is that the "Collapse" version is the entire view (which can be _very_ long).

tadean:
> This looks really good so far.
> The one thing that might be worth discussing is that the "Collapse" version is the entire view (which can be _very_ long).

I agree. Maybe it makes more sense to display something else other than actually executing the view in collapse mode.

tadean:
The widget in use by `viewsreference` seems to use an AJAX refresh for some of its functionality. That means that a proper unit test for it would likely need to make use of `JavascriptTestBase` as its base class. It's possible that a test for this paragraph depends on the resolution for issue #72 . Perhaps adding that test should be a followup after #72 .

PR #406, opened 2020-12-04
remove hirak/prestissimo

### LINKED ISSUE #405: Remove composer 1 performance plugins from local dev config files.
## Problem/Motivation
Can't build az_quickstart locally.

### Describe the bug
Because hirak/prestissimo is not supported in composer 2 we get errors when trying to build locally after composer self-update.

### To Reproduce
Clone repo
lando start

### Error message
                                                                                                                                             
```
  [InvalidArgumentException]                                                                                                                 
  Package hirak/prestissimo at version ^0.3.9 has a PHP requirement incompatible with your PHP version, PHP extensions and Composer version  
                                                                                                                                             

```

trackleft:
## Description
Remove hirak/prestissimo from .lando file since it is now outdated with composer2

According to the official documentation on this page
https://github.com/hirak/prestissimo
We should no longer be using it.

Closes #405

### DIFF
--- a/.lando.yml
+++ b/.lando.yml
@@ -8,8 +8,6 @@ config:
   xdebug: false
 services:
   appserver:
-    composer:
-      hirak/prestissimo: ^0.3.9
     overrides:
       environment:
         SIMPLETEST_DB: 'mysql://drupal8:drupal8@database/drupal8'


mmunro-ltrr (approved):
You got there first! I've been having to do this manually on all recent local lando builds. I was commenting out the line, but can already say that I've tested this. Was in the process of making the pull request.

joeparsons (approved):

joeparsons:
I've also been meaning to make a similar PR to this, thanks!

PR #415, opened 2020-12-09
Closes #414 - Adds Coffee module

### LINKED ISSUE #414: Add Coffee module to install profile
## Motivation
Quickstart 1.0 includes the [Coffee module](https://www.drupal.org/project/coffee), which allows site administrators to quickly navigate the admin interface of the site, without having to manually click through the interface. This functionality should also be available in Quickstart 2.0.

## Proposed Resolution
Add Coffee module to AZQS `composer.json` and `az_quickstart.info.yml` files.

On the current UAQS, only site administrators have access to utilize the Coffee module. Suggested to keep the same permissions, then revisit adding other roles once roles/permissions have been hashed out for the distribution as a whole.
- Administrator

akslay:
## Description
Closes #414 by adding the Coffee module to `composer.json` and `az_quickstart.info.yml` files.

**Probo build:** https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-415.probo.build/


trackleft (dismissed):

composer.json
```diff
@@ -71,6 +72,9 @@
             },
             "drupal/draggableviews": {
                 "Draggableviews select view that stores order issue": "https://www.drupal.org/files/issues/2019-12-17/draggableviews-sort_handler_specify_order_view-2767437-66.patch"
+            },
+            "drupal/coffee": {
+                "Coffee test, support for Drupal 9 core_version_requirement": "https://www.drupal.org/files/issues/2020-10-27/3174680-5-make-test-module-d9-compatible.patch"
             }
         }
```
joeparsons, line 79: It would be good to enable automatic package sorting in the composer.json file with this addition (in the extra section after patches):

```suggestion
        },
        "config": {
            "sort-packages": true
        }
```

composer.json
```diff
@@ -20,32 +20,33 @@
...
     ],
     "require": {
-        "php": ">=7.0",
         "az-digital/arizona-bootstrap": "2.0.5",
```
joeparsons, line 23: With automatic package sorting enabled (see comment near bottom of this file), the `php` version constraint still remains at the top of the list so we should probably leave the php line where it is.

```suggestion
        "php": ">=7.0",
        "az-digital/arizona-bootstrap": "2.0.5",
```

composer.json
```diff
@@ -20,32 +20,33 @@
...
-        "drupal/entity_embed": "1.1",
-        "drupal/embed": "1.4",
-        "drupal/viewsreference": "2.0-beta2",
-        "drupal/media_library_form_element": "2.0",
-        "drupal/ckeditor_bs_grid": "2.0.3",
-        "drupal/auto_entitylabel": "3.0-beta3",
         "drupal/views_bootstrap": "4.3",
-        "drupal/draggableviews": "2.0.0"
+        "drupal/viewsreference": "2.0-beta2",
+        "php": ">=7.0"
```
joeparsons, line 49: Per comment above, I think we actually want to leave this php line at the top of the list.
```suggestion
```

joeparsons (changes requested):

composer.json
joeparsons, line 49: Per comment above, I think we actually want to leave this php line at the top of the list.

```suggestion
        "drupal/viewsreference": "2.0-beta2"
```

danahertzberg:
Seems like this PR could benefit from separation of the Chosen module addition and organization of modules. I recommend creating a new ticket and separate PR for the organization task so we can merge Chosen more quickly and easily.

config/install/user.role.az_content_editor.yml
```diff
@@ -7,6 +7,7 @@ weight: 2
 is_admin: false
 permissions:
   - 'access administration pages'
+  - 'access coffee'
```
danahertzberg, line 10: The ticket says this should be allowed only for administrators. If new discussions warrant the need for this module to be accessible by content editors/admins, it should only be added to the content editor role (not the content administrator role)

kevcooper (approved):

joeparsons (approved):

jwmoore1 (approved):

danahertzberg (approved):

Issue #421, opened 2020-12-16
Pin GitHub Actions to Specific PHP Version


kevcooper:
Our processes are not yet ready to utilize PHP 8.x.x, we need to be able to pin actions to a specific version of PHP 7.x

## Conditions of satisfaction
- [ ] Action runs using PHP 7.x

PR #432, opened 2021-01-07
Set actions PHP version to 7.3

kevcooper:
This change uses https://github.com/shivammathur/setup-php to set the PHP version of this action to 7.3

This will close issue #421

### DIFF
--- a/.github/workflows/create-upload-release.yml
+++ b/.github/workflows/create-upload-release.yml
@@ -14,6 +14,10 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v2
       - name: Build Variables
+      - name: Setup PHP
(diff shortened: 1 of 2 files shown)


joeparsons (approved):

jcheek (approved):
looks good.  I notice that PHP7.3 has already reached EOL.

joeparsons:
> looks good. I notice that PHP7.3 has already reached EOL.

@jcheek Yeah, we need to look into upgrading everything to PHP 7.4 soon but currently all of our other CI tools (Probo, Lando, etc) are pinned at 7.3 as well.

jwmoore1 (approved):

trackleft (approved):

mmunro-ltrr (approved):

tadean (approved):

Issue #116, opened 2020-04-10
Create Person content type


barrettbaffert:
The MVP of the Person content type will match the D7 version of the content type. The addition of user functionality may or may not be added post-MVP (pending decision).

- [x]  Create content type and add necessary fields
- [x]  Style main view mode
- [x]  Add and configure "card" view mode
- [x]  Add and configure "row" view mode
- [x]  Add "Category" taxonomy
- [x]  Add "Category - Secondary" taxonomy
- [ ]  Add CSS file + styles for any Person-specific styles not included in bootstrap
- [ ] Export all configuration/templates/code & create pull request

akslay:
Drupal issue related to exposing the title field in view modes:
https://www.drupal.org/project/contrib_tracker/issues/2607758

trackleft:
It was determined that using the User entity type as content would be less ideal than adding a permission that allows authenticated users to edit their own content.

Related to: #3553

PR #433, opened 2021-01-07
Closes #413 - fix Person image on grid display

### LINKED ISSUE #413: Fix Person content type images on grid & row displays
## Problem/Motivation
The profile images on the Person content type should have the same height for each image in the Card view mode. Image should also be square on the Row view mode. 

### Describe the bug
If you upload a portrait orientation image rather than a landscape orientation image, it's taller than the other images. 

## Proposed resolution
Make use of the available image styles and add needed Media view modes to make use of image styles. Update node view modes to utilize those image styles.

### Expected behavior
Uniform image heights in the Person grid view. Square image in the Person row view.

### Considerations

- "Card Image" image style (which is also used in other places than Person content type)
- Cropping module (which is still pending)

akslay:
Closes #413

## Description

- Removes the pre-process function from `az_quickstart.theme` previously added in the initial Person content type PR
- Adds the "Card Image" view mode as an option for Media
- Enables the "Card Image" view mode for Media of type image
- "Card Image" view mode utilizes the pre-existing "Card Image" image style
- Enables "Card" vide mode for Person. Updates Person's "Card" view mode to use "Card Image" Media view mode
- Adds the "Square" view mode as an option for Media
- Enables the "Square" view mode for Media of type image
- Adds "Square" image style
- "Square" view mode utilizes "Square" image style
- Enables "Row" view mode for Person. Updates Person's "Row" view mode to utilize the "Square" Media view mode


danahertzberg:
Just a couple questions/comments.

1. To match QS1, the /people-list view should not use cropped images (just the same width)
2. I should not expect to be able to edit the crop on a file as a part of this PR, correct?
3. Is it possible to edit the photo from the person page (like in QS1)?

akslay:
> Just a couple questions/comments.
> 
> 1. To match QS1, the /people-list view should not use cropped images (just the same width)
> 2. I should not expect to be able to edit the crop on a file as a part of this PR, correct?
> 3. Is it possible to edit the photo from the person page (like in QS1)?

1. Ahh, I was looking at Kitten's view (https://kitten.ltrr.arizona.edu/people-list) and they were all at 220x220. But it looks like further down there's other sized images. I'll update the PR to remove everything pertaining to the Square and Row changes.
2. No, the cropping functionality is a separate issue--I'm not sure we've even decided which module we're going to use yet.
3. Yes, you can edit the image from the node view by using the contextual link that appears when you hover over the image. On the grid directory view, though, the contextual link for media was sacrificed in favor of the ability to edit the node via contextual link instead. Drupal core has active issues for overlapping contextual links, which make them unusable in general. By hiding one of the contextual links (media in this case) we can at least access the node edit form, and you can edit the image or node content from there. I have the Drupal issue link added in comments about that on the Person MVP ticket: https://github.com/az-digital/az_quickstart/issues/116 I'll make sure to add that to the comments on Person MVP Epic so it's in one place.

danahertzberg:
@akslay Thanks for making the changes!

For #3, I was referencing the ability to edit the image from the node _edit_ view. When editing the node now, you can only remove the image and add a new one. But maybe this is also related to the cropping module issue.

akslay:
@danahertzberg Ohh I see what you mean--thanks for the clarification. But I'm not sure on that one. I would think media would allow you to edit images after you add them. I will have to check and see if Media has any pending issues for that.

geijutsuka (approved):

mmunro-ltrr:
Looked ok on Probo, but did notice the possible reordering problem.

mmunro-ltrr (approved):

danahertzberg:
Looks like the /people-list view could benefit from a consistent width of the photos when the content stacks. See https://uasites.arizona.edu/people-list as an example.

modules/custom/az_media/config/install/core.entity_view_mode.media.az_square.yml
```diff
@@ -0,0 +1,9 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - media
+id: media.az_square
+label: Square
+targetEntityType: media
+cache: true
```
trackleft, line 9:
```suggestion
cache: true

```

modules/custom/az_media/config/install/core.entity_view_mode.media.az_card_image.yml
```diff
@@ -0,0 +1,9 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - media
+id: media.az_card_image
+label: 'Card Image'
+targetEntityType: media
+cache: true
```
trackleft, line 9:
```suggestion
cache: true

```

akslay:
> Looks like the /people-list view could benefit from a consistent width of the photos when the content stacks. See https://uasites.arizona.edu/people-list as an example.

I agree. It's not currently in UAQS but I think it's an easy enough improvement to include with this PR. I'll update it shortly to include this and Chris' new lines as well.

mmunro-ltrr (approved):

joeparsons (approved):

geijutsuka (approved):

danahertzberg (approved):

trackleft (approved):

jwmoore1 (approved):

PR #458, opened 2021-01-21
Fix name for "Create Release" action

kevcooper:
The name of this action was not being set correctly, this change should fix that.


joeparsons:
Not sure what's going on with ProboCI not reporting status to GitHub...

joeparsons (approved):

trackleft (approved):

PR #483, opened 2021-02-04
Bug/471 field az addresses

trackleft:
I was having problems making the PR agains the bug/471 branch, hopefully it works now.

Change field type to "Text (formatted, long)" with Plain text only,
Change machine name to "field_az_address",

### DIFF
--- a/modules/custom/az_core/config/install/field.storage.node.field_az_address.yml
+++ b/modules/custom/az_core/config/install/field.storage.node.field_az_address.yml
@@ -4,10 +4,10 @@ dependencies:
   module:
     - node
     - text
-id: node.field_az_addresses
-field_name: field_az_addresses
+id: node.field_az_address
+field_name: field_az_address
 entity_type: node
-type: text_with_summary
+type: text_long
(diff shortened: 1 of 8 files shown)


camikazegreen:
This PR is also updating the Twig template for the Person content type to use:
- field_az_body instead of field_az_bio
- field_az_attachments instead of field_cv_documents

camikazegreen (approved):
Everything that this PR touches looks correct on Probo.

smayden:
Two questions on this one:

- Is the address supposed to be plain text now? (I still see CKeditor)
- Is it just me or is there now slightly more spacing between address and phone number now?

image

trackleft:
> * Is the address supposed to be plain text now? (I still see CKeditor)
Oops, I did the configuration, but forgot to export it lol
> * Is it just me or is there now slightly more spacing between address and phone number now?
Address elements add style to things, we could  keep the `<address>` element out for now.

smayden (approved):
Address field now set to plain text for person content type and everything displaying as expected.

PR #487, opened 2021-02-05
add `/news` to page title block visibility rules config so the block will not appear on news content.

trackleft:


smayden (approved):
Verified that double news titles no longer appear on stories and I see that the correct configuration appears at: /admin/structure/block/manage/az_barrio_page_title?destination=/admin/structure/block

joeparsons (approved):

PR #510, opened 2021-02-15
Bug/365 Fix how Jquery is loaded with HTML Paragraph

### LINKED ISSUE #365: Ensure HTML Paragraph supports jQuery
## Motivation
Using jQuery in HTML areas to render Tableau charts is a common use case that is supported in Quickstart 1.0.
Examples include:
- https://uair.arizona.edu/content/admissions-pipeline
- https://covid19.arizona.edu/dashboard

If the same embed code that works in QS 1.0 is added to an HTML paragraph in Quickstart 2.0, it will fail to render because jQuery isn't being loaded until later in the page.

We could add jQuery as a dependency of the HTML paragraph to force it to load at the appropriate time when HTML is used on a page.

caosborne89:
Resolves #365

Updated the az_paragraphs configurations so that Jquery is loaded earlier in the DOM.

## Description

Thanks for a fellow developer at the Library (shout out to Ruiqi Yu!), who spoon fed me the solution, I committed the solution for this after wrapping my brain around it. Added a configuration to the az_paragraphs.libraries.yml file for az_html. That configuration adds core/jquery as a dependency for az_paragraphs_html. Then  enabled az_default_paragraph_behavior in the install config for az_html so that the library would get attached. This resolves #365 

## Related Issue

Bug #365 

## How Has This Been Tested?

I tested this by first enabling az_paragraphs_html. I then created a new page and added an HTML paragraph type to the page and inserted some JQuery into the HTML field. Before the fix the Jquery would not work, but after the code changes it seems to be working.

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_paragraphs/az_paragraphs.libraries.yml
+++ b/modules/custom/az_paragraphs/az_paragraphs.libraries.yml
@@ -38,3 +38,8 @@ az_paragraphs.preview:
   dependencies:
     - az_barrio/global-styling
     - az_barrio/arizona-bootstrap
+az_paragraphs.az_html:
+  version: VERSION
+  header: true
+  dependencies:
+    - core/jquery
--- a/modules/custom/az_paragraphs/az_paragraphs_html/config/install/paragraphs.paragraphs_type.az_html.yml
+++ b/modules/custom/az_paragraphs/az_paragraphs_html/config/install/paragraphs.paragraphs_type.az_html.yml
@@ -10,4 +10,4 @@ icon_default: null
 description: 'Add markup that can not be added with a text area or embedded media.'
 behavior_plugins:
   az_default_paragraph_behavior:
-    enabled: false
+    enabled: true


modules/custom/az_paragraphs/az_paragraphs.libraries.yml
```diff
@@ -38,3 +38,8 @@ az_paragraphs.preview:
   dependencies:
     - az_barrio/global-styling
     - az_barrio/arizona-bootstrap
+az_paragraphs.az_html:
+  version: VERSION
+  header: true
```
trackleft, line 43: I know that the ticket says move JQuery up, but I don't think it is required to be in the header, I think just the dependency is enough?

trackleft:
For a future ticket, I think we should consider a checkbox or some automated process for determining if jQuery is needed, or should be added, similar to how we add dependencies for material design icons in `az_barrio_library_info_alter()`
See https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/az_barrio.theme#L318-L324

modules/custom/az_paragraphs/az_paragraphs.libraries.yml
caosborne89: @trackleft I tested removing the header option, but it looks like JQuery still gets added near the closing tag for the body tag. It only seams to work for me when I add it to the header.

trackleft: That's so weird, we must have a bug in our JS somewhere that breaks jquery after it.

joeparsons (approved):
Since the issue we're trying to solve is caused by inline Javascript that is added by content editors inside of paragraphs, it makes sense to me that jQuery would need to be loaded in the HTML head.

This fix seems reasonable to me.  Thanks @caosborne89!

If we really want to try to optimize how/when jQuery is loaded, I think we'd need to look into making custom components (block plugins or paragraph types) for each of the types of things that content editors are embedding in HTML fields that require inline Javascript so we can convert the inline Javascript into libraries so we can control how/when the javascript for those embedded items is loaded.

modules/custom/az_paragraphs/az_paragraphs.libraries.yml
trackleft: Oops I read that comment incorrectly disregard my comment.about the bug in JS somewhere, that doesn't make any sense.

tadean (approved):

trackleft:
@camikazegreen can you please provide the Jquery javascript that you were using when you found the issue #365 in the first place.

We had some discussion about this ticket in our Wednesday PR meeting on Feb 17 2021 and did some testing on quickstart builds prior to this PR, and jquery seemed to work without it.

There was agreement that adding jquery as a dependency to this paragraph type would be a good idea, but found that adding jquery into the header, rather than the footer, was not strictly necessary for jquery to work.

caosborne89:
@trackleft I didn't report the original ticket, but I'll provide the JQuery I was testing.

```
<script> 
jQuery(document).ready(function(){
  jQuery("#flip").click(function(){
    jQuery("#panel").slideDown("slow");
  });
});
</script>
<style> 
#panel, #flip {
  padding: 5px;
  text-align: center;
  background-color: #e5eecc;
  border: solid 1px #c3c3c3;
}

#panel {
  padding: 50px;
  display: none;
}
</style>

<div id="flip">Click to slide down panel</div>
<div id="panel">Hello world!</div>
```
I just tested this on the commit before mine (on the same `bug/365` branch. I enabled the `az_paragraphs_html` module, created a new page and selected 'Add HTML' and placed that code in the text area. I get an 'Uncaught ReferenceError: jQuery is not defined' error message. This is all in a lando docker container fresh install. I do see `/sites/default/files/js/js_{random strings}.js` loaded on the page twice near the footer, and I think that includes JQuery. I was assuming that the browser was processing the 'az_paragraphs_html' block before JQuery is being defined.

The same thing happens if I add `core/jquery` as a dependency to `az_paragraphs.az_html`, but `header: true` is not set.

If I include `header: true` then `/sites/default/files/js/js_{random strings}.js` is included three times, once in the header and twice in the footer. In this case the inline JQuery works correctly.

camikazegreen:
Andy's comment above answers my one lingering question: "Was this PR still needed, or did we fix this problem during other work?"

He has verified that this is still broken until he submitted his fix, so I will add my approval as well.

camikazegreen (approved):

Issue #514, opened 2021-02-16
Improve reordering of paragraphs


smayden:
**Do not disclose security vulnerabilities here, please email us instead: 
az-digital-security@list.arizona.edu**

## Motivation
In our survey with content editors, reordering paragraphs came up as the #1 pain point. We also experimented with a solution when testing the paragraphs editor with content editors and they responded very favorably. I think this could be an simple, yet very impactful, improvement to Quickstart 2 and would love to see it in the beta release.

### Is your feature request related to a problem? Please describe.
Currently in QS 1, editor often rely on row weights to reorder paragraphs on a flexible page. This can be very tedious when needing to add content in the middle of the page. Our current implementations in QS 2 is somewhat improved as paragraphs appear collapsed by default, but editors still had difficulty using the cross hairs to drag and drop.

## Proposed Resolution
Enable drag and drop functionality for the Paragraphs widget. Content editors would then see:

![image](https://user-images.githubusercontent.com/10873398/108114010-7ad14c00-7055-11eb-8fb0-9931355fd8f3.png)

![image](https://user-images.githubusercontent.com/10873398/108114107-9f2d2880-7055-11eb-8bc4-2d6a7c815348.png)

### Describe the solution you'd like
When testing paragraphs with content editors (for example this page: https://paragraphs-try-gutenberg.pantheonsite.io/paragraphs-demo-page) we turned on functionality to Collapse/Edit all. This gave an option to easily Drag and Drop paragraphs (which they loved during testing). 

We may need to explore possible bugs with the Claro theme (possible patch: https://www.drupal.org/project/drupal/issues/3099026) and also how this would work with paragraphs using compound fields (but even if it didn't work there, would still be big improvement)

Settings are on the form display for the content type: https://paragraphs-try-gutenberg.pantheonsite.io/admin/structure/types/manage/flexible_page/form-display

![image](https://user-images.githubusercontent.com/10873398/108113924-5b3a2380-7055-11eb-82cb-22cf454d9930.png)

### Describe alternatives you've considered
When explored other solutions for laying out content (Gutenberg, Layout Builder) previously. Enabling the drag and drop functionality for paragraphs tested best for reordering things with content editors

### Roles and Permissions considerations
A clear and concise description of how each of the following roles would be impacted by this change:
- Anonymous user - not impacted
- Authenticated user - not impacted
- Content editor - easier editing experience, will need permissions
- Content administrator - easier editing experience, will need permissions
- Administrator - easier editing experience, will need permissions

PR #519, opened 2021-02-18
Resurface edit all/drag and drop functionality for paragraphs

smayden:
I originally thought that the "Edit All" and Drag and Drop functionality needed to be enabled on the Page content type, but it looks like it already was enabled but not displaying.

Chris Green kindly gave me the needed code to add to the az_core module that allows the table header to display properly.

## Description
This change allows the display of:
- Edit All button at the top of the content area that opens all paragraphs for editing
- Drag and Drop option under the the three dots on the right
- Collapse all option under the three dots on the right

The Drag and Drop functionality gives content editors an ultra-condensed view of the paragraphs on the page to easily drag and drop. This functionality was previously tested with content editors and was highly desired.

Note: It does not work with content inside accordions and cards. We may want to consider other solutions for that.

image

image

## Related Issue
https://github.com/az-digital/az_quickstart/issues/514

## How Has This Been Tested?
Tested locally

## Types of changes

## Checklist:


camikazegreen:
Dang Shoshana! This is one of those improvements that current site owners are going to be clamoring for! I am excited just to show it to them. 

The interface has one confusing feature (that should not block merging), I would just like to capture it here: After reordering you have to hit the "Complete Drag & Drop" button which is at the top of the workflow, when the natural next step would be to hit the "Save" button at the bottom of the workflow. Currently "Save" will disregard all of your changes.

smayden:
I agree that the save button is confusing. I wonder if there is a way to remove it and/or move the completion button to the bottom. That definitely came up in testing. Or least good documentation if we go forward with this.

Side bonus: I realized that paragraphs in D9 already have an added "duplicate" feature (that's working in the pre-release), something several content editors were asking for.

danahertzberg:
I like the functionality! Are there options for making them all exposed buttons? I would hate for this to be a "hidden" feature. 

Side note: Showing all buttons wouldn't be an issue when the Claro patch is working to not constrain the editing area (not sure why it's not working)

Also looks like there are some coding standard issues that need to be resolved.

modules/custom/az_core/az_core.module
```diff
@@ -73,3 +73,12 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   }
   return AccessResult::neutral();
 }
+/**
```
danahertzberg, line 76:
```suggestion

/**
```

trackleft:
Claro function this is overriding https://api.drupal.org/api/drupal/core%21themes%21claro%21claro.theme/function/claro_preprocess_field_multiple_value_form/8.8.x

modules/custom/az_core/az_core.module
```diff
@@ -73,3 +73,12 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+ * Implements hook_theme_registry_alter().
+ */
+function az_core_theme_registry_alter(&$theme_registry) {
+  $key = array_search('paragraphs_preprocess_field_multiple_value_form', $theme_registry['field_multiple_value_form']['preprocess functions']);
+  unset($theme_registry['field_multiple_value_form']['preprocess functions'][$key]);
+  $theme_registry['field_multiple_value_form']['preprocess functions'][] = 'paragraphs_preprocess_field_multiple_value_form';
+}
+
```
joeparsons, line 84:
```suggestion
```

modules/custom/az_core/az_core.module
```diff
@@ -73,3 +73,12 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   }
   return AccessResult::neutral();
 }
+/**
+ * Implements hook_theme_registry_alter().
+ */
+function az_core_theme_registry_alter(&$theme_registry) {
+  $key = array_search('paragraphs_preprocess_field_multiple_value_form', $theme_registry['field_multiple_value_form']['preprocess functions']);
```
tadean, line 81: I would suggest the following changes:

- `array_search` should use the optional argument to indicate strict comparison.
- the code should check if the array key is set before array searching.

modules/custom/az_paragraphs/az_paragraphs.module
```diff
@@ -22,3 +22,16 @@ function az_paragraphs_help($route_name, RouteMatchInterface $route_match) {
     default:
   }
 }
+
+/**
+ * Implements hook_theme_registry_alter().
+ */
+function az_paragraphs_theme_registry_alter(&$theme_registry) {
+  if ($theme_registry['field_multiple_value_form']) {
+    if ($theme_registry['field_multiple_value_form']['preprocess functions']) {
```
joeparsons, line 31: Nitpick:  I think it might be preferable to use `isset()` here:
```suggestion
    if (isset($theme_registry['field_multiple_value_form']['preprocess functions'])) {
```

trackleft:
https://7097f176-5f52-49c7-adf2-fcad5c28b32e.probo.build/

trackleft (approved):

tadean (approved):

camikazegreen (approved):

PR #546, opened 2021-02-26
Adding the webform module. Closes #222

camikazegreen:
## Description
This adds the Drupal 9 Webform module. 

This is left disabled by default. 

To get similar functionality to what we had in Quickstart 1.0, you will want to enable:
- Webform
- Webform UI
- Webform Node

All permissions are currently reserved for Admins.

## Related Issue
https://github.com/az-digital/az_quickstart/tree/issue/222

## How Has This Been Tested?
Locally with Lando. I have added a simple webform and added it to a page. 

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,6 +61,7 @@
         "drupal/token": "1.9",
         "drupal/views_bootstrap": "4.3",
         "drupal/viewsreference": "2.0-beta2",
+        "drupal/webform": "6.0",
         "drupal/xmlsitemap": "1.0"
     },
     "require-dev": {


camikazegreen:
The Drupal 9 version of Webform comes with  _a lot_ of sub-modules. Is there any way to exclude most of them from our build?

tadean (dismissed):

trackleft (dismissed):

composer.json
```diff
@@ -61,6 +61,7 @@
         "drupal/token": "1.9",
         "drupal/views_bootstrap": "4.3",
         "drupal/viewsreference": "2.0-beta2",
+        "drupal/webform": "6.0",
```
joeparsons, line 64: I think we should probably use the full latest version number:
```suggestion
        "drupal/webform": "6.0.1",
```

joeparsons (changes requested):

smayden (dismissed):
Able to enable webform and create a very simple form

jwmoore1 (changes requested):
There is a security update for the Webform module, so this should be updated to 6.0.2 before we merge.

joeparsons:
> There is a security update for the Webform module, so this should be updated to 6.0.2 before we merge.

Wow, good (timely) catch!  I updated my suggestion above.

config/install/user.role.az_content_editor.yml
```diff
@@ -6,18 +6,25 @@ label: 'Content editor'
 weight: 2
 is_admin: false
 permissions:
-  - 'access administration pages'
+   - 'access administration pages'
```
camikazegreen, line 9:
```suggestion
  - 'access administration pages'
```

trackleft (approved):

jwmoore1 (approved):

smayden (approved):
I tested use of webform with the admin, content admin, and content editor roles and things are working well/as expected.

I did note and discuss with Cameron that content editors are able delete webform submissions that aren't their own (e.g. anon users) from the Operations dropdown on the page listing submissions, which is unexpected. We tried some different permutations with the permissions to try and address this unsuccessfully. It make be a bug that I think could be addressed separately from this ticket.

joeparsons:
Some open questions I still have about Webforms for QS 2.0 include:
- Are we sure we want/need a separate `Webform` content type?  (Instead of just enabling webforms to be added to existing content types?)
- Is there anything we should be doing to ensure the webform module configuration is reasonably secure? (forms not allowed to upload files into public file systems, etc)

camikazegreen:
@joeparsons 

- I don't know that we need to create a separate content type, but I think that issue can mostly be handled by documentation.
- Would adding any kind of configuration require making an azqs_webform module to store the configuration? I think that could be a great improvement in the future and would probably also allow us to more simply enable a standard set of sub-modules. I would recommend pushing this improvement to General Availability.

Issue #439, opened 2021-01-08
Third Party Libraries


tadean:
## What is the problem that we want to solve?
Related to #409, we need a robust way of automating the download of third party libraries. `chosen` provides a drush command, but this isn't ideal because it then implies we need a way to automate the running of drush commands on different environments, e.g. pantheon, local, etc. Some potential options:

- Using composer asset packages (possibly as forks if none already exist)
- Libraries API module?
- Altering requests for a libraries to a CDN location rather than a file one
- Forcing a drush command to be run in environments

PR #561, opened 2021-03-05
Fixes #409, #439 Third Party Libraries and Add Chosen

### LINKED ISSUE #409: Add Chosen to AZQS install profile
## Motivation
It's nice to be able to select multiple elements, so a checkbox is preferred over a select list, but some sites have extremely long lists of taxonomy terms that make the checkbox interface too cumbersome. 

In Quickstart 1.0, we solved this using the Chosen module:
https://www.drupal.org/project/chosen

This module provides a select list interface, but makes it easy to use autocomplete and intuitively multi-select.

*I don't really think that this fits within the Person MVP, but I do think that it could easily be added as part of MVP and is part of parity with QS1. I do think though that this would classify more as "polish" and is therefore lower priority in my mind.

tadean:
This Pull request adds Chosen and a method of adding third party client-side libraries to Quickstart. **There is a matching pull request in the scaffolding repository** that adds [Asset Packagist](https://asset-packagist.org/) as a package repository.

## Description
This pull request uses [Asset Packagist](https://asset-packagist.org/) to add chosen and its client side library to Quickstart. This is presently being done by using Asset Packagist, which is a standalone packagist repository that allows importing NPM and Bower assets as if they were composer packages. A topic for discussion is if this provides enough flexibility for client side libraries we may want to import in the future.

It is worth noting that Chosen's library (`harvesthq/chosen`) is **already available** as a composer package. However, it is not configured correctly and lists its package type as `library`, (the default). This means that by default it is installed inside the `vendor` folder rather than where it is needed. This is a long term issue with the chosen library and hasn't been fixed yet as none of the current maintainers have many ties to PHP. There is a pull request regarding this waiting in the chosen repository.

Since the type is not correct, the only way to target it via `composer/installers` would be to mention chosen by name in the composer installer settings in the scaffolding. This didn't seem the best practice for long-term maintainability, so instead I have used an equivalent asset-packagist package, `bower-asset/chosen`.

An equivalent possible solution would be to mention a custom install location for `harvesthq/chosen` in the **scaffolding** repository. I've elected to use the current asset packagist solution in this PR as it addresses the **Third Party Libraries** #439 ticket as well.

This implies that chosen really is being added twice (once by asset-packagist and once from the incorrectly-maintained version that's added as a dependency by chosen), but once is to the Vendor folder where it isn't terribly useful:
```
  - Installing harvesthq/chosen (v1.8.7): Extracting archive
  - Installing drupal/chosen (3.0.1): Extracting archive
  - Installing drupal/chosen_lib (3.0.1)
  - Installing bower-asset/chosen (v1.8.7): Extracting archive
```

## Related Issue
#409 
#439 

## How Has This Been Tested?

- Visit `/admin/config/user-interface/chosen` and verify that it doesn't complain that the chosen library is missing
- Verify that chosen is set to be used only on admin pages
- Verify that chosen is set to always apply on both single select and multiselect.
- Verify that chosen is in use on admin pages.

## Types of changes

## Checklist:


joeparsons:
I know composer repositories should really only be added to root-level composer files (like our scaffolding repo) but do you think we could consider adding https://asset-packagist.org as a composer repository in this project's composer.json as well (like we did for https://packages.drupal.org/8).  Having it defined in both places would be helpful for things like dependabot...

tadean:
image

tadean:
> I know composer repositories should really only be added to root-level composer files (like our scaffolding repo) but do you think we could consider adding https://asset-packagist.org as a composer repository in this project's composer.json as well (like we did for https://packages.drupal.org/8). Having it defined in both places would be helpful for things like dependabot...

Good point. I don't think there'd be any harm in doing so, even if it's ignored by composer.

tadean:
Updated to add asset packagist repository settings to this copy of the `composer.json`, for dependabot purposes.

joeparsons (approved):

trackleft (approved):

PR #598, opened 2021-03-15
Closes #401: Fix blockquote styling

### LINKED ISSUE #401: Fix Blockquote styling in Standard text format
AZ Bootstrap now requires a class rather than just the tag to have styles actually applied to the blockquote tag. This ticket should investigate how to implement adding a class automatically to the blockquote tag when it's added with the WYSIWYG button.

**New requirements:**
`<blockquote class="blockquote">`

barrettbaffert:
## Description

AZ Bootstrap now requires a class rather than just the tag to have styles actually applied to the blockquote tag. This ticket should investigate how to implement adding a class automatically to the blockquote tag when it's added with the WYSIWYG button.

New requirements:
`<blockquote class="blockquote">`

## Related Issue

https://github.com/az-digital/az_quickstart/issues/401

### I tested locally with Lando, as follows:

- Create page content type` /node/add/az_flexible_page` using a text and card paragraph content type 
- Save to confirm the class has been added
- [Alternatively] Confirm `Bootstrap Utilities - Blockquote Classes` is checked on both text format editors `/admin/config/content/formats/manage/az_standard` and `/admin/config/content/formats/manage/basic_html`

### Questions & Concerns:

- Did I add the configuration correct?
- I didn't anticipate the need to add filter for az_standard 
- I'm also feeling uncomfortable about module and permissions attribute
- While it seems excessive to require a contrib module for this little problem (a) I'm not finding a CKEditor plugin that would do the trick and (b) this [bootstrap_utilities](https://www.drupal.org/project/bootstrap_utilities) might include other functionality we can consider using in the future? 

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

## Checklist:

- [X] I've added this Pull Request to the AZ Quickstart Github project and set it to "Review in progress".


barrettbaffert:
Blockquotes in action here

https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-598.probo.build/blockquotes

Keep in mind, when you view source in ckeditor, you don't see the class="blockquote" but then when you see published page, it's added

composer.json
```diff
@@ -32,6 +32,7 @@
         "drupal/auto_entitylabel": "3.0-beta4",
         "drupal/background_image_formatter": "1.7",
         "drupal/bootstrap_barrio": "4.30",
+        "drupal/bootstrap_utilities": "^1.0",
```
joeparsons, line 35: We'll want to pin to a specific version of the module.
```suggestion
        "drupal/bootstrap_utilities": "1.0",
```

config/install/filter.format.az_standard.yml
```diff
@@ -0,0 +1,62 @@
+langcode: en
```
joeparsons, line 1: The AZ Standard text format is already included in az_paragraphs/config/optional.  You'll want to update the existing exported config rather than add a new one here.  Ultimately we probably need to find a more intuitive home for our text formats (maybe az_core) but that can be a separate improvement from this PR.

joeparsons (commented):
I like this approach overall.  I looked into what would be needed to fix this problem via CKEdior plugins and the CKEditor and never found anything very viable.  The text filter approach used by the contrib module you've added here makes a lot of sense to me.  Added a couple of specific comments for some things that will need to be changed.

config/install/filter.format.az_standard.yml
barrettbaffert: Thanks! Missed this

az_quickstart.info.yml
```diff
@@ -62,6 +62,7 @@ install:
   - chosen:chosen
   - ckeditor_bs_grid:ckeditor_bs_grid
   - coffee:coffee
+  - bootstrap_utilities:bootstrap_utilities
```
joeparsons, line 65: I think this module dependency probably needs to get moved into the `az_media` module to prevent the unmet dependencies exception getting thrown during the Drupal install step on Probo (which is also causing the PHPunit tests to fail)
https://app.probo.ci/#/dashboard/build/46fe5b32-4723-4591-bfc2-6e4715ef3c40/task/5e0f50c0f6000003
```
In UnmetDependenciesException.php line 98:
                                                                               
  [Drupal\Core\Config\UnmetDependenciesException]                              
  Configuration objects provided by az_media have unmet dependencies: filter.format.az_standard 
  (bootstrap_utilities)
```

joeparsons: Actually, since the basic format is provided by the install profile, the module dependency probably needs to be in both `az_quickstart.info.yml` _and_ `az_media.info.yml`...

barrettbaffert: Oh, darn, sorry I missed this before making rapid fire commits. And thanks @joeparsons !!

I removed from `az_quickstart.info.yml` but added to `az_media.info.yml` and it's passed tests. If it works, should I leave it out of `az_quickstart.info.yml`?

joeparsons: > I removed from `az_quickstart.info.yml` but added to `az_media.info.yml` and it's passed tests. If it works, should I leave it out of `az_quickstart.info.yml`?

As-is, things work because other `az_quickstart` dependencies require `az_media` which in turn requires `bootstrap_utilities` but since we're adding `bootstrap_utilities`-specific config to the basic text format inside `az_quickstart/config/install`, I think it would be better to explicitly declare `bootstrap_utilities` as a hard dependency for `az_quickstart` as well.

barrettbaffert: Gotcha! Ok, I added it back to this file. Really appreciate you walking me through it

barrettbaffert:
Unfortunately, adding `blockquote.blockquote|Blockquote` as an option in the styles dropdown didn't prevent the `blockquote` class from getting stripped [here](https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-598.probo.build/blockquotes). ([Screenshot](https://arizona.box.com/s/xfm8nx39jft9b4v3fd9igoabe2y91jix))

I did confirm it's available in the dropdown ([screenshot](https://arizona.box.com/s/ejjws0aqz9uu5l88v4u2nj2a74dygxow)). But on save, the class attribute is stripped of the `blockquote` value.

barrettbaffert:
With @joeparsons's patch merged into this pull request, things are working as I was hoping. Whether I copy-paste the markup with classes from Arizona Bootstrap, or just type into the CKEditor and click the blockquote button (and even apply text align style options), it's working well!

Example https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-598.probo.build/blockquotes

(Compare link above to [screenshot before Joe's patch](https://arizona.app.box.com/s/vxig2tsm6elzmp5tnfynj40t91z0siid))

joeparsons (approved):

trackleft (approved):

tadean (approved):

Issue #572, opened 2021-03-10
Create Reorder Carousel view


danahertzberg:
## Motivation
Provide ability to reorder carousel content type items

## Conditions of satisfaction
- [ ] Should be restricted to authenticated users
- [ ] Should use row weights or draggable weights
- [ ] Should update the order of carousel items on save

PR #605, opened 2021-03-22
Add Carousel Item content type

### LINKED ISSUE #512: Build new front page carousel
Implement as a content type - same as QS1

akslay:
## Description
Closes #512 by adding the custom `az_carousel` module to the distribution. The module adds the Carousel Item content type as an option to install onto your AZQS site (off by default).

**Review Link:** https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-605.probo.build/

## Related Issue
https://github.com/az-digital/az_quickstart/issues/512

## How Has This Been Tested?
- Tested locally by cloning the project and creating a fresh build––no errors present.
- Ran `lando phpunit`––no errors related to this task.

## Types of changes

## Checklist:

### Requirements
The following contrib modules were added as requirements of this module:
- Rabbit Hole (routes anonymous users away from the node display and back to the home page)
   - Users with editor/admin roles bypass this setting 
- Link Class (ability to add class to link field display)
   - I added this to the install profile since other modules could also benefit from it, not just the Carousel
- Slick (provides the functionality for the carousel)
- Slick Views (provides the views functionality for the carousel)
- Blazy (dependency of Slick)
   - Added Slick, Slick Views, and Blazy to install profile since the paragraphs implementation may or may not use Slick as well.

The following 3rd party libraries were added (dependencies of the corresponding modules):
- Slick Carousel
- Blazy

### Code Additions & Changes
- Added Carousel Item content type
   - Default display mode for the Carousel Item content type
   - Form display for Carousel Item content type
- Added `az_carousel_item` media display mode
- Added `az_carousel_item` image style (1400x500 scale and crop). This is normally about 1100px, but had to increase a bit since this carousel is wider than the D7 version.
- Added `AZ Carousel` view
   - Block view display for the home page
- User role permissions (matches other content types)
   - Content editors & content admins can create/edit
   - Content admins can delete
- Slick configuration for "Center Mode" (what carousel is using)
- Code block similar to AZ Person that provides a quick link to edit carousel items (since contextual links aren't working). This block will also be the future home of the button to reorder carousel items (separate ticket). Block is restricted to site editors & admins visibility.
- Modified Smart Title config to be enabled for Carousel Item content type.

### Notes & Follow-ups
- I was unable to get the "navigation" for the carousel to work the way we had it in D7. I was able to get it to show up, but the navigation would slide just like the carousel itself, and I don't think we want that sort of experience. If we want to spend further time investigating adding a textual navigation to the Carousel, it might be better to do it as a follow-up ticket rather than have it be a blocker for this ticket.
- In lieu of the textual navigation, I enabled the "dots" navigation below the carousel. This will provide users with the same level of visual indication as to how many slides are present in the carousel, since there is infinite scrolling.
- Was unable to find any markup hidden in the UI for contextual links. Might want to consider submitting a Drupal issue for Slick Views. It seems like it might not be getting rendered at all. Didn't find any issues related to this problem.
- Reorder carousel view (https://github.com/az-digital/az_quickstart/issues/572)


smayden (commented):
This is slick! I like the overall feel of it better than the current carousel implementation.

I was curious about whether you could add the carousel view via the view paragraph to other pages so I tried it out (is that a use case we'd want to support?). It displayed well, but caused what I think is a bug on the edit screen when I subsequently went to edit the page.

image

The main content bar extends across the page covering some of the panels on the right hand side (but only when I add the carousel).  Perhaps we just remove carousel as an option from adding with Views Paragraph?

akslay:
> Perhaps we just remove carousel as an option from adding with Views Paragraph?

Thanks for the review and finding the bug Shoshana. (: I think I'd be of the opinion that we don't want people to embed as a block, and to instead use the regular block section to place it on their site. Especially since this particular one is _supposed_ to be for the "home page". Technically we can't enforce that though, I suppose.

camikazegreen:
I agree that this should not be one of the limited Views available to the Views paragraph item.

camikazegreen:
Is this module being enabled by default? I can't tell.

akslay:
> Is this module being enabled by default? I can't tell.

No it's off by default. Turned on on the Probo build for testing purposes.

camikazegreen (approved):

smayden:
One question on the permissions: Should content editors be allowed to delete their own content?  We allow them to for the other content types (but delete any content is reserved for content admins)

image

akslay:
> One question on the permissions: Should content editors be allowed to delete their own content? We allow them to for the other content types (but delete any content is reserved for content admins)

Ahh yes, those permissions were meant to be the same as the other content types. I'll fix that––thanks for catching that.

danahertzberg:
Can the rabbit hole settings be removed from the node level? Or is that the only way to configure it? It could be confusing to content editors.

akslay:
> Can the rabbit hole settings be removed from the node level? Or is that the only way to configure it? It could be confusing to content editors.

The rabbit hole settings are on a node-type basis. That's what allows anonymous users to be routed away from the node display page of any Carousel Item and back to the carousel as a whole on the home page instead. This does not apply to site editors though, so that they can still get to the individual carousel items to edit them, and won't be routed back to the home page every time they try to access the carousel item they're trying to view/edit.

For clarification, which part do you think would be confusing for the editors?

smayden:
@danahertzberg I was worried about this too, but I think only admins see the Rabbit Hole settings. Content editors and content admins don't see them on the edit screen:

danahertzberg:
Yes, that clarifies it @smayden. Thanks! As long as content admins/editors can't see it I think we are in the clear.

@akslay Is there a way to have those settings only on the Carousel Item content type and not Page or other types?

akslay:
> Is there a way to have those settings only on the Carousel Item content type and not Page or other types?

Yes (: the Rabbit Hole settings are applied on a per-node type basis. So they are currently only enabled for the Carousel Item content type and none of the other content types.

composer.json
```diff
@@ -54,22 +54,28 @@
...
         "drupal/media_library_form_element": "2.0.3",
         "drupal/media_library_theme_reset": "1.0",
         "drupal/metatag": "1.15",
         "drupal/migrate_plus": "5.1",
         "drupal/migrate_tools": "5.0",
         "drupal/paragraphs": "1.12",
         "drupal/pathauto": "1.8",
+        "drupal/rabbit_hole": "1.0-beta10",
         "drupal/redirect": "1.6",
+        "drupal/slick": "2.2",
```
joeparsons, line 67: `drupal/slick` requires `drupal/blazy` in its own composer.json file:
```
  "require": {
    "drupal/blazy": "~2.0"
  }
```

It might be good to add `drupal/blazy` as an explicit dependency to az_quickstart so we can pin to a specific version for more deterministic builds...

modules/custom/az_carousel/az_carousel.module
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * @file
+ * Contains az_carousel.module.
+ */
+
+/**
+ * Implements hook_preprocess_node().
+ */
+function az_carousel_preprocess_node(&$variables) {
```
tadean, line 11: It looks like this may be adding the carousel library as a dependency to every node.

joeparsons: Agreed.  It'd be good to scope this to just the one content type.

danahertzberg:
Notes from 3/26 Arizona Digital meeting:
- Remove Rabbit hole module
- Update permissions (see Shoshana's comment)
- Exclude this content type from site search
   - Could also make node path use `/admin` to hide from public
- Investigate weirdness on mobile
Screen Shot 2021-03-26 at 10 12 26 AM

composer.json
akslay: Update: decided not to do this––allow the modules to track their own dependencies.

danahertzberg (approved):
Follow up tickets need to be created:

- [ ] Fix white space issues on 1750px+ (@danahertzberg reported)
- [ ] Look into inconsistent offset issues (@danahertzberg reported)
- [ ] Look into arrow below content on desktop issue (@smayden reported)

tadean (approved):

PR #606, opened 2021-03-23
Configure Content Moderation

### LINKED ISSUE #577: Add Workbench moderation
Match Quickstart 1 features and permissions.

https://www.drupal.org/project/workbench

---------------------------------

**Update as of 3/24/2021:**
Instead of matching UAQS 1.0 we will modify the transitions to the following:

- Draft
- Set to published
- Set to unpublished

Modify permissions as follows:
- Allow editors to publish/approve content

akslay:
## Description
Workbench Moderation has been replaced by Content Moderation in Drupal core. Content Moderation will remain uninstalled by default. Configuration has been updated to include workflows that match UAQS D7. UAQS D7 did not have an editor role to compare to, so I made an education assumption on what permissions should be (content admins can publish and reject revisions).

**Review link:** https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-606.probo.build/

## Related Issue
Closes #577

## How Has This Been Tested?
Tested locally.

## Types of changes
Changes are modifications to existing files only––no new files were added––configuration file for workflow already existed in Quickstart so I modified it to match D7's workflows. 

## Checklist:


config/optional/workflows.workflow.editorial.yml
```diff
@@ -4,48 +4,42 @@ dependencies:
   module:
     - content_moderation
 id: editorial
```
joeparsons, line 6: Should we change the id of this workflow to be `az_editorial` since we're modifying the one copied from from the Standard install profile?  If so, it probably be best to prefix all of the state names as well and to rename this file (`workflows.workflow.az_editorial.yml`).

akslay: > Should we change the id of this workflow to be az_editorial since we're modifying the one copied from from the Standard install profile?

I actually ran into an issue when I tried to make a fresh workflow called `az_moderation`. The view wasn't working properly (filters had to be modified) and if we have to modify the view it would require a way for us to be able to override the default config in the view. Which requires one of those "quickstart" folders to be added. And I wasn't sure if we wanted to do that or not since the config currently isn't inside a separate `az_content_moderation` module or anything.

joeparsons: Ok, I see.  Looks like that "Moderated content" view is [provided by the core content_moderation module](https://git.drupalcode.org/project/drupal/-/blob/9.1.x/core/modules/content_moderation/config/optional/views.view.moderated_content.yml) (accessible at `admin/content/moderated`).    I think you're right that we'd have add a modified version of that view to a `config/quickstart` directory somewhere to override it since that is essentially a default administration view.  Are there other changes we'd want to make to that view, anyway, or not really?

I am kind of wondering if it would be better to package this up functionality in a separate `az_content_moderation` module...  I kind of like that idea although there wouldn't necessarily be much to include in that module yet (unless we do want to also override that view).  Currently I think that custom module would just need to enable the core content_moderation and workflows modules and probably house this workflows.workflow.editorial.yml config.

akslay: I'm not opposed to the idea of having an `az_content_moderation` module if we plan on having it more customized. Perhaps we could rope in the "anonymous reviewer workflow" into that module as well?

danahertzberg:
Based on conversations in Slack:

Defaults should be:
- Start in draft
- Set to publish
- Set to unpublished (back to draft)
   - Newest draft should always be the editable version (whether there is a published version or not)

Permissions for all states should be available to:
- Content editor

config/optional/workflows.workflow.editorial.yml
akslay: @joeparsons Do you think we should remove "editorial" workflow all together and just have one for `az_moderation`? Or just revert Editorial back to whatever it had before and add `az_moderation` in addition to that?

joeparsons: > @joeparsons Do you think we should remove "editorial" workflow all together and just have one for `az_moderation`? Or just revert Editorial back to whatever it had before and add `az_moderation` in addition to that?

Good questions.  I think we should have a group discussion about this.  I do think it could be good to have our own workflow that doesn't conflict with the standard editorial one but it sounds like that would mean we'd have to maintain our own moderated content view as well...

akslay: Decision: we will keep the PR as is for now. And in the future, if we end up needing to modify the default view at all, we can create a new workflow at that point and provided a database update to remove the old workflow.

akslay:
This one should be good to go––just needed to merge in upstream changes.

camikazegreen (approved):

smayden (approved):
This worked as expected for me

Issue #617, opened 2021-03-26
Change kitten workflow watch timeout to 5 minutes


trackleft:
Building a review site on pantheon shouldn't take longer than 5 minutes, so a timeout for longer than that is unnecessary.
Once timeout occurs, workflow can continue to the next step.

PR #618, opened 2021-03-26
Change timeout minutes from 10 to 5 for workflow watch

trackleft:
## Related Issue
#617


camikazegreen:
We are going to try to fix the "watching the logs" failing, rather than bumping down the timeout.

Issue #377, opened 2020-11-13
Quickstart Card module follow-up tasks


joeparsons:
On 2020-11-13 we merged an initial version of the Quickstart cards module containing card field plugins and a template but we identified several follow-up improvements that are still needed.

## Feedback from 2020-11-13 Meeting:

- [x] Use Standard format by default for body (fix now)
- [ ] How do we delete cards? (investigate now)
- [x] Move background color selection to top
- [x] Move media subfield to second top (fix now)
- [x] Change link_uri label to Card Link URL (fix now)
- [ ] Can we remove blank card that shows up at bottom? (investigate now)
- [x] 2-step edit process for cards? (follow-up)

smayden:
Some additional comments from reviewing PR #634:

-When I click edit on cards (actually all paragraphs) it seems to jump me down to the bottom of the edit area. I think this will confuse content editors
-We might consider making the edit/collapse buttons for each card in the upper right-hand corner to be consistent with where these buttons appear elsewhere on the edit screen.
-Is there a way to make the whole card clickable? I couldn't figure out how to do that (maybe we decided not to do that?)
-Should we consider being able to change the color of the link button? I made a card with red background and button blends with background (could button switch to reverse similar to how text does based on color?)
-Is there a reason we default to basic html rather than the standard text format? (I probably missed or forgot the conversation...) I was trying to make a card with my own button in the card body and was unable to do so until I switched to standard (and agreed to the scary warning).

tadean:
Addressed the following in #634 

- Use standard format for body by default
- Move background color selection to top
- Move media subfield to second top
- Change link_uri label to Card Link URL

PR #634, opened 2021-04-01
Fixes #390, Fixes #427, Fixes #538, Fixes #613 Card Widget Updates

### LINKED ISSUE #390: Investigate Two-step Edit Forms for Card Widget

During comparison between the approach of nested paragraphs and custom field plugins for cards, one benefit of the nested paragraphs approach is that the preview of each card was rendered and cards could be edited individually. This was convenient since it meant fewer active edit forms and kept page length down, and also made it easier to reorder cards within a deck.

By comparison, with the card field plugin, opening a card deck means rendering the edit form of every card, which makes it relatively more difficult to navigate and find the card you need to edit.

Now that we have custom plugins, we could investigate making the edit form of each card start off with a card preview and only render the edit form for that card if an edit button is pressed. (implementing this as a two-step form with an ajax refresh)

This would give some of the **UI advantages** of the nested paragraph approach without the corresponding database bloat issues of paragraphs.

## Conditions of success

Experiment if the card forms can be rendered as two-step pop open forms with an edit button.

### LINKED ISSUE #427: Card Widget Updates
For MVP1 the custom card field type needs a little additional polish, consisting of the following:

- implementation of the settings subfield that gets saved as serialized php settings.
- ability to delete an individual card (general d9 field plugin issue: see [multiple value field issue](https://www.drupal.org/project/drupal/issues/1038316) )

### LINKED ISSUE #538: Correct primitive type Cards issue
When I try to add card media I get an error saying “the value should be of the correct primitive type”. And it doesn’t let me delete or edit the media, or save the page either once that error shows up. Only option is to exit the page and lose all of the content you were trying to add to the page––this will be a big problem for content editors.

Original documentation: https://docs.google.com/document/d/1NlE2MqbG19mbDmYf3BggrpKKcVBZsrYHe2XnYStZuPE/edit#heading=h.2whr3gihc8cw

### LINKED ISSUE #613: Add background color selection to Cards
At the card level (per card), I should be able to select the background color from a pre-defined list of options.

- White (default)
- Transparent
- Arizona Red
- Arizona Blue
- Sky
- Oasis
- Azurite
- Midnight
- Bloom
- Chili
- Cool Gray
- Warm Gray
- Leaf
- River
- Silver
- Ash

tadean:
This pull request updates many aspects of the card widget of our custom card field, primarily the introduction of card previews and two-step editor forms so that the entire deck does not open at once.

## Description
The following changes are made to the card type and card widgets:

- Storage of the media ID is sanitized so that it may be NULL
- Corrected issue where the media widget was providing a comma delimited list of Media IDs
- Card field forms are now a two-step AJAX process where an `Edit Card` button is pressed to edit an individual card, as opposed to all field inputs showing simultaneously.
- Card background color selection
- Card background color implementation
- On the edit screen, a rendered preview is provided for collapsed cards.

## Related Issue
#538 
#427 
#390 
#613 
#377 

## How Has This Been Tested?

- Create a page
- Add cards
- Verify that for new cards they do not initially appear collapsed (eg. all fields are accessible for cards newly added to the entity)
- Save page
- Edit page
- Edit cards paragraph
- Verify that each card you added has a visible preview, but that the form fields are hidden
- Click `Edit Card` and verify it expands to make its form fields accessible.
- Make changes
- Verify that an image can be removed from the card (Fixes #538)
- Click `Collapse Card` and verify it reverts back to preview mode with changes reflected. (or skip this step and leave it expanded)
- Save page
- Verify changes were updated
- Verify background colors can be set for cards
- Verify background colors appear for cards.

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
+++ b/modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
@@ -9,7 +9,6 @@ use Symfony\Component\Validator\ConstraintViolationInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\NestedArray;
-use Drupal\file\Entity\File;
 
 /**
  * Defines the 'az_card' field widget.
@@ -39,6 +38,13 @@ class AZCardWidget extends WidgetBase {
   protected $pathValidator;
 
   /**
+   * Drupal\Core\Entity\EntityTypeManagerInterface definition.
+   *
+   * @var \Drupal\Core\Entity\EntityTypeManagerInterface
+   */
+  protected $entityTypeManager;
+
+  /**
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
@@ -51,6 +57,7 @@ class AZCardWidget extends WidgetBase {
 
     $instance->imageFactory = ($container->get('image.factory'));
     $instance->pathValidator = ($container->get('path.validator'));
+    $instance->entityTypeManager = $container->get('entity_type.manager');
     return $instance;
   }
 
@@ -127,7 +134,7 @@ class AZCardWidget extends WidgetBase {
       // Check and see if we can construct a valid image to preview.
       $media_hint = $items[$delta]->media ?? NULL;
       if ($media_hint) {
-        $file = File::load($media_hint);
+        $file = $this->entityTypeManager->getStorage('file')->load($media_hint);
         if ($file) {
           $image = $this->imageFactory->get($file->getFileUri());
           if ($image && $image->isValid()) {
--- a/modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
+++ b/modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
@@ -9,6 +9,7 @@ use Symfony\Component\Validator\ConstraintViolationInterface;
 use Symfony\Component\DependencyInjection\ContainerInterface;
 use Drupal\Component\Utility\Html;
 use Drupal\Component\Utility\NestedArray;
+use Drupal\file\Entity\File;
 
 /**
  * Defines the 'az_card' field widget.
@@ -31,6 +32,13 @@ class AZCardWidget extends WidgetBase {
   protected $imageFactory;
 
   /**
+   * Drupal\Core\Path\PathValidator definition.
+   *
+   * @var \Drupal\Core\Path\PathValidator
+   */
+  protected $pathValidator;
+
+  /**
    * {@inheritdoc}
    */
   public static function create(ContainerInterface $container, array $configuration, $plugin_id, $plugin_definition) {
@@ -42,6 +50,7 @@ class AZCardWidget extends WidgetBase {
     );
 
     $instance->imageFactory = ($container->get('image.factory'));
+    $instance->pathValidator = ($container->get('path.validator'));
     return $instance;
   }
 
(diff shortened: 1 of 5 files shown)


tadean:
image

image

image

tadean:
The following issue still affects cards which has pending resolution in Drupal core. It is not covered by this PR:

- [Allow for deletion of a single value of a multiple value field - General Field API issue](https://www.drupal.org/project/drupal/issues/1038316)

smayden (approved):
Tested as a content editor and verified that all the improvements outlined for this PR is working for as expected. Super impressed with the QS 2 implementation for cards!

I'll make a few notes as a separate comment for possible improvements down the line but none of them are really in the scope of this PR.

smayden:
A few things I noticed while testing this PR that might be improvements for cards down the line (not necessarily for this ticket):
-When I click edit on cards (actually all paragraphs) it seems to jump me down to the bottom of the edit area. I think this will confuse content editors
-We might consider making the edit/collapse buttons for each card in the upper right-hand corner to be consistent with where these buttons appear elsewhere on the edit screen.
-Is there a way to make the whole card clickable? I couldn't figure out how to do that (maybe we decided not to do that?)
-Should we consider being able to change the color of the link button? I made a card with red background and button blends with background (could button switch to reverse similar to how text does based on color?)
-Is there a reason we default to basic html rather than the standard text format? (I probably missed or forgot the conversation...) I was trying to make a card with my own button in the card body and was unable to do so until I switched to standard (and agreed to the scary warning).

danahertzberg:
Follow up tasks are documented here: https://github.com/az-digital/az_quickstart/issues/377

danahertzberg (dismissed):

danahertzberg:
@tadean Will make easy updates from #377 before merging this.

geijutsuka (commented):
Works for me :)
There was an error when entering after uploading the media, but appears to be a Media/Core problem? This is all I could find relating to the error:
- https://www.drupal.org/forum/general/general-discussion/2020-10-20/the-specified-ajax-callback-is-empty-or-not-callable
- https://www.drupal.org/project/drupal/issues/2866330
Error:
`Symfony\Component\HttpKernel\Exception\HttpException: The specified #ajax callback is empty or not callable. in Drupal\Core\Form\FormAjaxResponseBuilder->buildResponse() (line 67 of /src/az-quickstart-scaffolding/web/core/lib/Drupal/Core/Form/FormAjaxResponseBuilder.php).`

geijutsuka (dismissed):

joeparsons:
@tadean I'm seeing a strange issue when editing an existing card.  It seems when I edit an existing cards paragraph on an existing page, when the collapsed version of the card is rendered in its preview mode, the wrong image is used.  Here is a short screencast of the behavior on the Probo site:

https://user-images.githubusercontent.com/471936/113449802-78cc0e00-93b3-11eb-867e-bb33debf92b5.mp4

tadean:
> @tadean I'm seeing a strange issue when editing an existing card. It seems when I edit an existing cards paragraph on an existing page, when the collapsed version of the card is rendered in its preview mode, the wrong image is used. Here is a short screencast of the behavior on the Probo site:
> 
>  card-image-issue.mp4

Good find @joeparsons ! I suspect there's probably something we need to do in order to tell the renderer that the card image needs to be regenerated. Most fields probably handle that automatically and it may be something our compound field needs to do too. I'll check into that.

joeparsons:
Also, for what it's worth, I tried going through the steps of adding a card with Javascript disabled in my browser and found that clicking the "Add media" button for the media form element doesn't work _at all_.  I imagine that is proably an issue with the Media Library Form Element module we're using 😞.

tadean:
> Also, for what it's worth, I tried going through the steps of adding a card with Javascript disabled in my browser and found that clicking the "Add media" button for the media form element doesn't work _at all_. I imagine that is proably an issue with the Media Library Form Element module we're using .

Bummer :frowning_face: 

Maybe when that FAPI media library widget gets incorporated into core it'll be a little more complete. I don't think a modal could work without javascript, but there's no reason they couldn't make it a multistep form element.

joeparsons:
> Maybe when that FAPI media library widget gets incorporated into core it'll be a little more complete. I don't think a modal could work without javascript, but there's no reason they couldn't make it a multistep form element.

Actually it looks like it could be a limitation of the media_library module in core:
https://www.drupal.org/project/drupal/issues/2987675

modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
```diff
@@ -20,11 +23,149 @@
...
+
+    // Generate a preview if we need one.
+    if (!$status) {
+
+      // Bootstrap wrapper.
+      $element['preview_container'] = [
+        '#type' => 'container',
+        '#attributes' => [
+          'class' => ['col-12', 'col-sm-6', 'col-md-6', 'col-lg-4', 'card-preview',
+          ],
```
joeparsons, line 120: Nitpick: is this closing bracket intentionally on a new line?

tadean: I'll take a look at tweaking that - that one required some tinkering to get something `phpcs` was happy with. I do think it'd look better on the same line too.

modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
```diff
@@ -20,11 +23,149 @@
...
+    $instance->entityTypeManager = $container->get('entity_type.manager');
+    return $instance;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function form(FieldItemListInterface $items, array &$form, FormStateInterface $form_state, $get_delta = NULL) {
+
+    // Create shared settings for widget elemeents.
```
trackleft, line 69:
```suggestion
    // Create shared settings for widget elements.
```

joeparsons:
> Good find @joeparsons ! I suspect there's probably something we need to do in order to tell the renderer's caching that the card image needs to be regenerated. Most fields probably handle that automatically and it may be something our compound field needs to do too. I'll check into that.

FWIW, it looks like the `AZCardDefaultFormatter::generateImageRenderArray()` method is already doing something to add cache dependencies via the renderer.  Perhaps that method could be moved into a service that could be used by both the Formatter and Widget plugins?

modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
```diff
@@ -20,11 +23,149 @@
...
+    // This is necessary because wigets have to be AJAX replaced together,
```
trackleft, line 70:
```suggestion
    // This is necessary because widgets have to be AJAX replaced together,
```

trackleft:
> > Good find @joeparsons ! I suspect there's probably something we need to do in order to tell the renderer's caching that the card image needs to be regenerated. Most fields probably handle that automatically and it may be something our compound field needs to do too. I'll check into that.
> 
> FWIW, it looks like the `AZCardDefaultFormatter::generateImageRenderArray()` method is already doing something to add cache dependencies via the renderer. Perhaps that method could be moved into a service that could be used by both the Formatter and Widget plugins?


Some more context that may or may not be relevant: image only appears when card is collapsed within the edit form, also, the image that does appear is a temporary image https://take.ms/YZLbC

trackleft:
Could be related to that image that Rebecca received an error about

modules/custom/az_card/src/Plugin/Field/FieldWidget/AZCardWidget.php
tadean: I think this should be resolved now.

tadean:
Updated the following:

- Fixed some documentation typos (thanks @trackleft !)
- Reformatted an array initialization to be more readable (thanks @joeparsons !)
- Created class constant for default text format.

From #377 
- Use standard format for body by default
- Move background color selection to top
- Move media subfield to second top
- Change link_uri label to Card Link URL

tadean:
Pending tasks:

- [ ] Refactor card image generation to service

tadean:
Refactored to use a shared service between the formatter and widget to provide card image rendering.

joeparsons (approved):
All the recent changes look good to me.  I no longer see the issue with the wrong card image being rendered in the preview mode like I was before.  Nice work @tadean!

barrettbaffert (approved):
This is great! Very enjoyable editing experience. 

https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-634.probo.build/barrett-card-test

🃏🃏🃏

smayden (approved):
Tested as content editor and all working as expected. Images are rendering correctly. Minor tweaks to the order of fields make things easier. Thanks!

danahertzberg (approved):

trackleft (approved):

danahertzberg:
> Works for me :)
> There was an error when entering after uploading the media, but appears to be a Media/Core problem? This is all I could find relating to the error:
> 
> * https://www.drupal.org/forum/general/general-discussion/2020-10-20/the-specified-ajax-callback-is-empty-or-not-callable
> * https://www.drupal.org/project/drupal/issues/2866330
>   Error:
>   `Symfony\Component\HttpKernel\Exception\HttpException: The specified #ajax callback is empty or not callable. in Drupal\Core\Form\FormAjaxResponseBuilder->buildResponse() (line 67 of /src/az-quickstart-scaffolding/web/core/lib/Drupal/Core/Form/FormAjaxResponseBuilder.php).`

Follow up ticket: https://github.com/az-digital/az_quickstart/issues/638

PR #669, opened 2021-04-21
Fixes #127, Fixes #662, Fixes #663, Fixes #701: Added migration testing ProboCI workflow and D7 migrations for Users, Person nodes, Files, and Media.

### LINKED ISSUE #127: Person content type migration
Develop migration path for Person content type. Map fields from QS1 to QS2.

### LINKED ISSUE #662: User migration
## Motivation
Migrating users from Quickstart 1.0 sites to Quickstart 2.0 sites will be a critical part of migration. 

Drupal 9's core migrate functionality already migrates users, but there are some limitations for our use case:

- All Quickstart sites use CAS as the login solution instead of the standard Drupal logins. This is not handled successfully by Core Migrate.

- Quickstart 1.0 did not come with standard Roles and Permissions, but Quickstart 2.0 does. For most use cases, we **do not** want to carry over the roles and permissions from QS1 sites, and instead would prefer to allow mapping during the migration. Note: for sites that have intentionally added more roles, we want to allow them to be migrated.

CAS: 
The users' netids should be listed as the cas username, and the "Allow user to log in via CAS" checkbox should be checked.
Screen Shot 2021-04-09 at 3 42 50 PM

Roles & Permissions:
This is a harder question and is probably worth some group discussion.

- We can map any users with only "Authenticated User" to the same role in QS2, but there aren't many of these users.
- Any higher permissioned users could be migrated to the "Content Editor" role by the principal of least privilege, and then expect any higher roles to be explicitly granted on a case by case basis.

If there is an easy way to do this, we could provide an interactive component (either via UI or drush) that would:
- display the roles in the QS1 site
- allow you to map those roles to QS2 roles (Note: Quickstart 2 roles are additive, so it is expected that someone with the Content Administer role will also have the Content Editor role or they won't be able to do much)

Note: There are also many cases in QS1, where site builders never created any roles beyond the default and just assigned the admin role to all users. _This should not be carried over to new sites_.

- Anonymous user
- Authenticated user
- Content editor
- Content administrator
- Administrator

### LINKED ISSUE #663: Investigate Files and Media migrations
We need to verify that we are using Media consistently between each of our content types in Quickstart 1 and 2. If we are not, we might need to convert some of our files from Quickstart 1 to media entities in Quickstart 2 during our migrations.

## Conditions of satisfaction
- [ ] The method of adding images to News, Event and Person content types is documented for both Quickstart 1 and 2.
- [ ] A recommended next step is documented
- [ ] An estimate of time and effort is provided for the recommendation

### LINKED ISSUE #701: Migration testing pipeline
## Motivation
We need a way to easily test that our migrations are working.

## Proposed Resolution
Create a migration testing pipeline that can be run on every pull request.

#### Source options
- Flat file database export
- Pantheon sandbox site
- Generate D7 site as part of pipeline?

#### Destination site options
- Pantheon sandbox site
- ProboCI PR environment
- LAMP stack within GitHub Actions

#### Source DB configuration settings
- settings.php template with environment variables?
- Populate environment variables at runtime (using connection info from terminus?)
- .env file? (assuming .env files get loaded before settings.php)
- At runtime as drush command arguments?

#### Ideal workflow
- Build Quickstart codebase for destination site with composer (based on commit that triggered pipeline)
- Deploy codebase artifact to destination hosting environment
- Get source DB credentials
- Update settings.php in destination site with source DB credentials ?
- Install Quickstart on destination site
- Run drush migrate commands

smitahantzios:
This Migration module consists of person Content Type Migration Plugin

## How Has This Been Tested?
Please refer to the readme file to see testing instructions

## Types of changes

## Checklist:


modules/custom/az_migration/config/install/migrate_plus.migration.az_node_person.yml
```diff
@@ -0,0 +1,150 @@
+id: az_node_person
+label: Person Node Content
+migration_group: az_migration
+status: true
+
+source:
+  plugin: az_person
```
joeparsons, line 7: @smitahantzios Did you try using just the standard `d7_node` source plugin here rather than implementing your own?  There is nothing too special about the D7 Person content type so I was pretty surprised to see all of the code in your custom Person source plugin.  I don't really think all of that custom code should be necessary to get the source data from the D7 person content type.

joeparsons: (Something like this?)

```source:
  plugin: d7_node
  node_type: uaqs_person
```

modules/custom/az_migration/config/install/migrate_plus.migration.az_node_person.yml
```diff
@@ -0,0 +1,150 @@
...
+        default_value: az_standard
+
+  field_az_body:
+    plugin: sub_process
+    source: person_biography
+    process:
+      delta: delta
+      value: value
+      summary: summary
+      format:
```
trackleft, line 89: We'll probably need another process for all fields that use az_standard to convert embedded entities here eventually.

modules/custom/az_migration/src/Plugin/migrate/process/MediaGenerate.php
```diff
@@ -0,0 +1,78 @@
+<?php
```
trackleft, line 1: For this process, I wonder if it would be better to migrate media first and then reference it later, rather than generating media on the fly?

https://www.drupal.org/project/media_migration

tadean: I agree with this. I think Migration of media entities should be a separate migration that the person migration depends on, with resolution via `migration_lookup` plugin.

modules/custom/az_migration/src/EventSubscriber/MigrationSubscriber.php
```diff
@@ -0,0 +1,141 @@
...
+    }
+  }
+
+  /**
+   * Remove the media entities for that field and type.
+   *
+   * @param string $field_name
+   *   The field name we are checking.
+   */
+  private function removeMediaEntities($field_name) {
```
tadean, line 111: There should be cases where other fields or content types should reference the same piece of media content as media can be shared and reused. Trying to manually compute a rollback in an event listener doesn't seem necessary.

I'd recommend migration of media be its own migration that takes place before the person migration; rolling back person doesn't necessarily mean media needs to be deleted. That would let the media migration keep track of which content came from the source system via its own migration map table.

modules/custom/az_migration/config/install/migrate_plus.migration.az_person_categories.yml
```diff
@@ -0,0 +1,41 @@
+id: az_person_categories
+label: Person Categories
+migration_group: az_migration
+status: true
+
+source:
+  plugin: az_d7_taxonomy_term
```
tadean, line 7: The standard `d7_taxonomy_term` plugin seems like it could be used here.

modules/custom/az_migration/config/install/migrate_plus.migration.az_node_person.yml
tadean: Agreed - I think it'd be desirable to use the standard `d7_` plugins where possible.

smitahantzios: D7 person content type - 15 fields. Out of 15 only two having multi value fields. So while fetching and mapping need to loop only for this two fields, to make it more faster migration process. "d7_node" plugin use function "getFieldValues" - also executes the same database query in generic way and fetch all values. To make the migration process more faster in large number of records, wrote the custom Plugin. Please let me know your thoughts. Thanks.

modules/custom/az_migration/config/install/migrate_plus.migration.az_node_person.yml
smitahantzios: ok. for now, as per the discussion, tagged directly to "az_standard" format as default. Since the direct format mapping from d7 is not present.

modules/custom/az_migration/src/Plugin/migrate/process/MediaGenerate.php
smitahantzios: Migrate all media entity is dumping all unwanted media into the migrated site. It is good to fetch only the wanted media that relevant/wanted for the content and map it. Thanks

modules/custom/az_migration/config/install/migrate_plus.migration.az_person_categories.yml
smitahantzios: yes, I agree. I will update it. Thanks

modules/custom/az_migration/src/EventSubscriber/MigrationSubscriber.php
smitahantzios: Tried to fetch the wanted media as per the content, using the existing file entities. So It is fully dependent on content migration mapping, that is why removing the media in rollback. Using the existing file entity won't duplicate the files, and also here not deleting the files in this rollback. Thanks.

modules/custom/az_migration/config/install/az_migration.settings.yml
```diff
@@ -0,0 +1,5 @@
+migrate_d7_protocol: "https"
+migrate_d7_filebasepath: "dev-shantzios-d7.pantheonsite.io/"
```
camikazegreen, line 2: Can this be generic?

smitahantzios: This default value, We can update using the drush command as mentioned in readme file. Thanks

modules/custom/az_migration/config/install/migrate_plus.migration.az_node_person.yml
tadean: In my opinion using the core `d7_node` plugin is preferable to adding a custom `az_person` plugin from the standpoint of code maintainability. Additional speed is a good feature to have, but migrations will generally be run only once per site.

smitahantzios (commented):
@joeparsons @trackleft @tadean @camikazegreen  User Migration code has been pushed. Please review and comment!

joeparsons:
I think we should try using these two media_migration patches together to see if it clears up the issues you've been seeing with media:
https://www.drupal.org/project/media_migration/issues/3181174
https://www.drupal.org/project/media_migration/issues/3209875

modules/custom/az_migration/config/install/migrate_plus.migration.az_media.yml
```diff
@@ -0,0 +1,31 @@
+id: az_media
+label: All Media (media_migration)
+migration_group: az_migration
+audit: true
+migration_tags:
+  - Drupal 7
+  - Content
+deriver: Drupal\media_migration\Plugin\migrate\D7FileEntityDeriver
+source:
+  plugin: d7_file_entity_item
```
tadean, line 10:
```suggestion
source:
  plugin: d7_file_entity_item
  constants:
    bundle_prefix: az
```

modules/custom/az_migration/config/install/migrate_plus.migration.az_media.yml
```diff
@@ -0,0 +1,31 @@
...
+process:
+  uuid:
+    -
+      plugin: media_migrate_uuid
+      source: fid
+    -
+      plugin: skip_on_empty
+      method: process
+  mid: fid
```
tadean, line 19:
```suggestion
```

tadean: I don't think this should be present. `mid` and `fid` are not the same identifiers.

modules/custom/az_migration/config/install/migrate_plus.migration.az_media.yml
```diff
@@ -0,0 +1,31 @@
...
+  uid: uid
+  bundle: bundle
```
tadean, line 21:
```suggestion
  # Bundles must exist, previous machine names were image, document, etc.
  # Note constants/bundle_prefix evaluates to 'az'
  bundle:
    plugin: concat
    source:
      - constants/bundle_prefix
      - bundle
    delimiter: '_'
  # Lookup fid through migration lookup. Should not depend on fid not changing.
  mfid:
    plugin: migration_lookup
    migration: az_files
    source: fid
  # Bundles have different field names for their target reference.
  field_media_az_audio_file/target_id: '@mfid'
  field_media_az_document/target_id: '@mfid'
  field_media_az_image: '@mfid'
  field_media_az_video_file: '@mfid'
```

modules/custom/az_migration/config/install/migrate_plus.migration.az_media.yml
tadean: Storing this constant for later to add with the bundle names.

tadean (commented):
Made a few suggestions about az_media` that allowed it to work for me:

image

tadean:
We should likely consider moving these under `migrations` rather than `config`:
[Migrations are plugins instead of configuration entities](https://www.drupal.org/node/2668742)

I noticed this when migration definition weren't updating during cache clear as expected. The canonical place for migrations is in `MODULENAME/migrations` now. See the `az_demo` module for an example, which also has a migration.

tadean:
It looks like this module needs to list `media_migration` as one of its dependencies. I had to enable it separately to run the migrations.

.probo.yaml
```diff
@@ -50,8 +53,35 @@ steps:
...
+      - cp example.sites.php sites.php
+      - echo "\$sites['$BUILD_ID--site-migration.probo.build'] = 'migration';" >> sites.php
+      - cp ${ASSET_DIR}/migration.settings.php migration/settings.php
+      - cd migration
+      - mkdir files && chown www-data:www-data -R ./files
+      - drush site:install az_quickstart --site-name='Quickstart Migration' --account-name=azadmin --account-pass=azadmin --verbose --yes -n
+      - terminus env:wake ${MIGRATION_SOURCE_SITE_NAME}.dev
+      - drush -y en az_migration
+      - drush -y cset az_migration.settings migrate_d7_filebasepath " "
+      - drush -y cset az_migration.settings migrate_d7_public_path "sites/default/files/migrate_file"
```
joeparsons, line 82: These lines appear to be redundant.
```suggestion
```

joeparsons:
## 2021-05-10 Next steps

- [ ] Merge `main` branch into PR branch or rebase PR branch on `main` + resolve any conflicts
- [ ] Remove `hook_install()` implementation from `az_migration.install`
- [ ] Add `// @phpstan-ignore-next-line` comments before each static global `\Drupal` method call so that PHPstan code quality checks can pass
- [ ] Remove redundant `drush cset` lines from `.probo.yaml`
- [ ] Resolve file permission issues in Probo migration destination site if they still exist after above steps

.probo.yaml
```diff
@@ -54,6 +57,29 @@ steps:
       - chown www-data:www-data -R /var/www/html/sites/default/files/simpletest
       - source $ASSET_DIR/credentials.sh
```
joeparsons, line 58: This needs to be `test-credentials.sh`
```suggestion
      - source $ASSET_DIR/test-credentials.sh
```

.probo.yaml
```diff
@@ -52,8 +55,32 @@ steps:
...
+      - terminus connection:info ${MIGRATION_SOURCE_SITE_NAME}.dev --fields 'mysql_host,mysql_password,mysql_port,mysql_database,mysql_username' --format=json > /tmp/migration_config.json
+      - mkdir -p /var/www/html/sites/migration && cd /var/www/html/sites
+      - cp example.sites.php sites.php
+      - echo "\$sites['$BUILD_ID--site-migration.probo.build'] = 'migration';" >> sites.php
+      - cp ${ASSET_DIR}/migration.settings.php migration/settings.php
+      - cd migration
+      - mkdir files && chown www-data:www-data -R ./files
+      - drush site:install az_quickstart --site-name='Quickstart Migration' --account-name=azadmin --account-pass=azadmin --verbose --yes -n
+      - terminus env:wake ${MIGRATION_SOURCE_SITE_NAME}.dev
+      - drush user:role:add az_content_admin azcontentadmin
```
joeparsons, line 80:
```suggestion
```

.probo.yaml
```diff
@@ -52,8 +55,32 @@ steps:
...
+      - drush -y en az_migration
+      - drush -y cset az_migration.settings migrate_d7_protocol "https"
+      - drush -y cset az_migration.settings migrate_d7_filebasepath "dev-${MIGRATION_SOURCE_SITE_NAME}.pantheonsite.io/"
+      - drush -y cset az_migration.settings migrate_d7_public_path "sites/default/files"
+      - drush mim --group=az_migration
```
joeparsons, line 85: Let's move the static user account creation steps here:

```suggestion
      - drush mim --group=az_migration
      - drush user:create azadmin --password="azadmin"
      - drush user:role:add administrator azadmin
```

.probo.yaml
```diff
@@ -52,8 +55,33 @@ steps:
...
+      - source $ASSET_DIR/test-credentials.sh
       - export SIMPLETEST_BASE_URL="http://localhost"
-      - export SIMPLETEST_DB="mysql://${DATABASE_USER}:${DATABASE_PASS}@localhost/${DATABASE_NAME}"
+      - export SIMPLETEST_DB="mysql://${TEST_DATABASE_USER}:${TEST_DATABASE_PASS}@localhost/${TEST_DATABASE_NAME}"
       - cd /var/www/html/profiles/custom/az_quickstart
       - sudo -u www-data -E find $SRC_DIR/az-quickstart-scaffolding/web/profiles/custom/az_quickstart/ -name "*Test.php" | sudo -u www-data -E $SRC_DIR/az-quickstart-scaffolding/vendor/bin/fastest -vvv "$SRC_DIR/az-quickstart-scaffolding/vendor/bin/phpunit {} --colors="always" --verbose;"
+  - name: Log in to terminus
+    command: 'source ${ASSET_DIR}/terminus.sh; terminus auth:login --machine-token=${TERMINUS_TOKEN}'
+  # - name: Set up migration data
+  #   command: 'source ${ASSET_DIR}/migration.sh'
```
joeparsons, line 66: I think we can delete this commented-out step.
```suggestion
```

modules/custom/az_migration/az_migration.info.yml
```diff
@@ -0,0 +1,11 @@
...
+description: 'Provides migrations for the site.'
+core_version_requirement: ^8.8 || ^9
+type: module
+package: The University of Arizona
+dependencies:
+ - drupal:migrate
+ - drupal:migrate_plus
+ - drupal:migrate_tools
+ - drupal:migrate_drupal
+ - drupal:media_migration
```
joeparsons, line 11: Should we declare any of the modules that provide our destination entity bundles/fields as dependencies here? (e.g. `az_person`, `az_cas`, etc)

smitahantzios: I think, .info file should be good place to declare it, I will make that change and push it.

modules/custom/az_migration/migrations/migrate_plus.migration_group.az_migration.yml
```diff
@@ -0,0 +1,25 @@
+# The machine name of the group, by which it is referenced in individual migrations.
+id: az_migration
```
tadean, line 2: This file might be able to remain in `config/install` since it's a migration group, not a migration.

modules/custom/az_migration/az_migration.info.yml
tadean: I agree with that, I think declaring the `az_person` etc. modules here would be good.

modules/custom/az_migration/migrations/migrate_plus.migration_group.az_migration.yml
smitahantzios: Sounds good. I will make that change and commit

modules/custom/az_migration/migrations/migrate_plus.migration.az_media.yml
```diff
@@ -0,0 +1,52 @@
...
+migration_tags:
+  - Drupal 7
+  - Content
+  - Quickstart Content Migration
+source:
+  plugin: d7_file_entity_item
+  constants:
+    bundle_prefix: az
+process:
+  uid: uid
```
tadean, line 14: A question for PR review group discussion: this process pattern of `uid:uid` is enforcing that the UIDs of migrated users (and other types of entities where this pattern is used) are using the same id they were on the original site. This can sometimes cause content to get overwritten if they shared the same id as on the previous site. Is that behavior what we want to enforce?

For example, if the site already has any existing users, they will likely be overwritten by this migration.

The alternative is to not specifically list an ID mapping, in which case migrate will choose a new free id. It's still possible to know that ID later (through `migration_lookup` result). This also allows running the migration later if content was missed, or if it's needed to migrate two sites content to a single site.

modules/custom/az_migration/migrations/migrate_plus.migration_group.az_migration.yml
```diff
@@ -0,0 +1,25 @@
...
+ source:
+   key: migrate
+
+# We add dependencies just to make sure everything we need will be available
+dependencies:
+ enforced:
+   module:
+     - az_migration
+     - migrate_drupal
+     - migrate_tools
```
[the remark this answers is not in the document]
joeparsons: After taking a closer look at this I'm thinking it might be best to put the module dependencies in the specific migration plugins:
- `az_node_person` and `az_person_categories` migrations should depend on the `az_person` module
- `az_cas_user` migration should depend on the `az_cas` module
- `az_media` migration should depend on the `az_media` module

joeparsons:
## 2021-05-11 Next Steps

- [ ] [Move `migrate_plus.migration_group.az_migration.yml` back into `config/install`](#discussion_r630573893)
- [ ] [Remove default `migrate_d7_filebasepath` value from `az_migration.settings.yml`](#discussion_r630507502)
- [ ] [Add module dependencies to migration plugins](#discussion_r630587005)
- [ ] [Remove ID mappings from migration plugins (e.g. `uid: uid`) (pending discussion in PR meeting tomorrow)](#discussion_r630584164)

trackleft (approved):

modules/custom/az_migration/migrations/migrate_plus.migration.az_person_categories.yml
```diff
@@ -0,0 +1,49 @@
...
+source:
+  plugin: d7_taxonomy_term
+  bundle: uaqs_person_category
+
+destination:
+  plugin: entity:taxonomy_term
+  default_bundle: az_person_categories
+
+process:
+  tid: tid
```
trackleft, line 17:
```suggestion
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_person.yml
```diff
@@ -0,0 +1,179 @@
...
+  plugin: entity:node
+  bundle: az_person
+  migration_dependencies:
+    requireed:
+      - az_person_categories
+      - az_person_categories_secondary
+
+process:
+  nid: nid
+  vid: vid
```
trackleft, line 24:
```suggestion
```

tadean, line 24: This causes a similar issue as in the case of user migrations - attempting to preserve the node ID overwrites existing nodes on the destination site.

modules/custom/az_migration/migrations/migrate_plus.migration.az_files.yml
```diff
@@ -0,0 +1,51 @@
...
+        - uri
+
+  filemime: filemime
+  # filesize is dynamically computed when file entities are saved, so there is
+  # no point in migrating it.
+  # filesize: filesize
+  status: status
+  created: timestamp
+  changed: timestamp
+  fid: fid
```
trackleft, line 38:
```suggestion
```

trackleft (approved):

tadean (approved):

jwmoore1 (approved):

kevcooper (approved):

mmunro-ltrr (approved):

camikazegreen (approved):

PR #752, opened 2021-05-17
Closes #135: Migration for D7 event content type.

### LINKED ISSUE #135: Create event migration

smitahantzios:
Migration for Events Content Type

## How Has This Been Tested?
Please check the Probo Link under Migration Tests to see this working

## Types of changes

## Checklist:


modules/custom/az_migration/migrations/migrate_plus.migration.az_event_categories.yml
```diff
@@ -41,3 +41,9 @@ process:
     source: '@parent_id'
 
   changed: timestamp
+
```
joeparsons, line 44: Looks like we actually merged the event_categories migration as part of your last PR (?).  Did we actually want to preserve term IDs for event categories (line 16-17 of this file)?
```
process:
  tid: tid
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,190 @@
...
+destination:
+  plugin: entity:node
+  bundle: az_event
+  migration_dependencies:
+    requireed:
+      - az_event_categories
+
+process:
+  nid: nid
+  vid: vid
```
joeparsons, line 23: I don't think we actually want/need to preserve event node IDs or revision IDs.

smitahantzios: took care of this one..

modules/custom/az_migration/migrations/migrate_plus.migration.az_event_categories.yml
smitahantzios: We dint. I have removed this line

modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_contact.yml
```diff
@@ -0,0 +1,46 @@
...
+destination:
+  plugin: 'entity_reference_revisions:paragraph'
+  default_bundle: az_contact
+
+process:
+
+  field_az_email:
+    plugin: sub_process
+    source: field_uaqs_contact_email
+    revision_id: revision_id
```
joeparsons, line 21: I think we want to omit the revision_id mappings in these migrations as discussed in our PR meeting yesterday.
```suggestion
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_contact.yml
```diff
@@ -0,0 +1,46 @@
...
+      delta: delta
+      value: value
+
+  field_az_title:
+    plugin: sub_process
+    source: field_uaqs_contact_name
+    process:
+      delta: delta
+      value: value
+    revision_id: revision_id
```
joeparsons, line 40: I think we want to omit the revision_id mappings in these migrations as discussed in our PR meeting yesterday.
```suggestion
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,188 @@
...
+        plugin: default_value
+        default_value: az_standard
+  
+  field_az_summary:
+    plugin: sub_process
+    source: field_uaqs_summary
+    process:
+      delta: delta
+      value: value
+      summary: summary
```
tadean, line 81: Field `field_az_summary` does not enable the summary subfield.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,188 @@
...
+      uri: url
+      title: title
+
+  field_az_body:
+    plugin: sub_process
+    source: field_uaqs_body
+    process:
+      delta: delta
+      value: value
+      summary: summary
```
tadean, line 70: Field `field_az_body` does not enable the summary subfield.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,188 @@
...
+      format:  
+        plugin: default_value
+        default_value: plain_text
+
+  field_az_attachments:
+    plugin: sub_process
+    source: field_uaqs_attachments
+    process:
+      delta: delta
+      target_id: fid
```
tadean, line 91: This `target_id` should be a result of a `migration_lookup` I think.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,188 @@
...
+      value: value
+      format: format
+
+  field_az_link:
+    plugin: sub_process
+    source: field_uaqs_link
+    process:
+      delta: delta
+      uri: url
+      title: title
```
joeparsons, line 54: The default value for `url` on the source D7 `field_uaqs_link` field is the `[node:url]` token, which doesn't work in D9.  I think we may want to add something similar to what you did for Carousel to replace `[node:url]` values either with an empty string or `<none>`.

danahertzberg:
I triggered a rebuild since it got reaped.

danahertzberg:
The times don't appear to match the source
- https://dc509246-c69b-4bdf-aeb7-4a773c18184b--site-migration.probo.build/calendar
- https://kitten.ltrr.arizona.edu/calendar

And we don't have a short title field on D9 yet, right?

mmunro-ltrr:
With the non-matching times, one issue that came up with the old kitten site was getting the timezone set right everywhere (there was a PHP configuration issue at one point). I've not yet looked at what is happening here.

joeparsons:
> With the non-matching times, one issue that came up with the old kitten site was getting the timezone set right everywhere (there was a PHP configuration issue at one point). I've not yet looked at what is happening here.

Looks like the D7 Kitten site on Pantheon has its default timezone configured as Arizona but the default timezone on the Probo Migration site is UTC.  We probably need to add something to our ProboCI config to change the default timezone.

joeparsons:
I just changed the default timezone on the Probo migration site for this PR to Phoenix.  @danahertzberg Do you want to compare the dates again?

danahertzberg:
> I just changed the default timezone on the Probo migration site for this PR to Phoenix. @danahertzberg Do you want to compare the dates again?

@joeparsons Do I need to clear cache or anything? I'm not seeing the times updated on this page. https://dc509246-c69b-4bdf-aeb7-4a773c18184b--site-migration.probo.build/calendar

mmunro-ltrr:
The strange aspect was that the times on the Probo build looked correct, but the ones on kitten were wrong (events spanning days & talks scheduled in the small hours of the morning). Looks like they're 7 hours too early there, which would fit in with the time zone offset, but not in the expected place.

tadean:
Since the migration mapping is using `strtotime` callback to handle its conversion to timestamp, and strtotime isn't really timezone aware, I'm curious if there's some extra layer of time zone change taking place unintentionally beyond site timezones ( though it might not show up in cases where the source set is UTC)

```
    process:
      delta: delta
      value:
        plugin: callback
        callable: strtotime
        source: value
 ```

File: modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml (hunk 5 above)
Comment (joeparsons) [inline modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml]:
```suggestion
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
joeparsons:
```suggestion
```

joeparsons:
> Since the migration mapping is using `strtotime` callback to handle its conversion to timestamp, and strtotime isn't really timezone aware, I'm curious if there's some extra layer of time zone change taking place unintentionally beyond site timezones ( though it might not show up in cases where the source set is UTC)

@tadean Are you thinking PHP ini settings?

joeparsons:
@smitahantzios There are still some unaddressed comments/suggestions on this PR (beyond the time zone issues we're looking into).

joeparsons:
Here is a comparison of the various PHP/Drupal timezone settings between the source & destination sites:

Setting | D7 Kitten Site (Pantheon) | Migration Test Site (Probo)
------------ | ------------- | -------------
PHP default timezone `date_default_timezone_get()` | America/Phoenix | UTC
PHP INI `date.timezone` | UTC | No value
Drupal default timezone | America/Phoenix | UTC

tadean:
> > Since the migration mapping is using `strtotime` callback to handle its conversion to timestamp, and strtotime isn't really timezone aware, I'm curious if there's some extra layer of time zone change taking place unintentionally beyond site timezones ( though it might not show up in cases where the source set is UTC)
> 
> @tadean Are you thinking PHP ini settings?

I believe Drupal stores all dates in UTC in the database; however I think `strtotime` uses the default PHP time zone. So, in cases where the default PHP timezone is not UTC on the destination machine this might result in the wrong times being migrated. E.g. it's not necessarily aware the output timezone needs to be UTC I think.

The source DB can be assumed to be UTC however, since that's how Drupal should have the columns stored.

camikazegreen:
Remaining tasks on this PR:

- [ ] Remove Summary Fields (per Joe's comments)
- [ ] Resolve UTC Time issues
- [ ] Update the target id for attachments to use a migration lookup.

kevcooper:
https://www.php.net/manual/en/datetime.formats.php
https://www.php.net/manual/en/datetime.formats.time.php

Looks like you should be able to specify timezone or timezone correction as part of the time format

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
joeparsons: @smitahantzios This comment has still not been addressed.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,210 @@
...
+    plugin: sub_process
+    source: field_uaqs_date
+    process:
+      delta: delta
+      value:
+        -
+          plugin: format_date
+          from_format: 'Y-m-d H:i:s'
+          to_format: 'Y-m-d\TH:i:s'
+          from_timezone: 'America/Phoenix'
```
tadean, line 151: The source should already be `UTC` from a Drupal database, I think the only change necessary is the need to display the timezone in the date so that `strtotime` uses it.

This is to prevent `strtotime` from outputting the wrong timestamp if its default timezone is not UTC.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
```diff
@@ -0,0 +1,210 @@
...
+          source: value
+        -
+          plugin: callback
+          callable: strtotime
+      end_value:
+        -
+          plugin: format_date
+          from_format: 'Y-m-d H:i:s'
+          to_format: 'Y-m-d\TH:i:s'
+          from_timezone: 'America/Phoenix'
```
tadean, line 162: Same issue as above.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_event.yml
smitahantzios: @tadean I will work on it as per your comment and commit somthing soon. Thanks

mmunro-ltrr (approved):

joeparsons (approved):

tadean (approved):

PR #755, opened 2021-05-17
Feature/734 - Carousel Migration

smitahantzios:
## Description
Carousel Type Migration

## How Has This Been Tested?
You can test this in Probo build

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_migration/az_migration.info.yml
+++ b/modules/custom/az_migration/az_migration.info.yml
@@ -9,4 +9,3 @@ dependencies:
  - drupal:migrate_tools
  - drupal:migrate_drupal
  - drupal:media_migration
- - az_carousel
--- a/modules/custom/az_migration/az_migration.info.yml
+++ b/modules/custom/az_migration/az_migration.info.yml
@@ -9,3 +9,4 @@ dependencies:
  - drupal:migrate_tools
  - drupal:migrate_drupal
  - drupal:media_migration
+ - az_carousel
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
@@ -39,7 +39,12 @@ process:
     source: field_uaqs_link
     process:
       delta: delta
-      uri: url
+      uri: 
+        -
+          plugin: str_replace
+          search: '[node:url]'
+          replace: ''
+          source: url
       title: title
 
   field_az_carousel_short_summary:
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
@@ -92,12 +92,6 @@ process:
       - constants/slash
       - path
 
-dependencies:
-  enforced:
-    module:
-      - az_migration
-      - az_carousel
-
 migration_dependencies:
   required:
     - az_user
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
@@ -1,0 +1,100 @@
+id: az_node_carousel
+label: AZ Carousel Item
+migration_group: az_migration
+migration_tags:
+  - Quickstart Content Migration
+status: true
+
+source:
+  plugin: d7_node
+  node_type: uaqs_carousel_item
+  constants:
+    slash: '/'
+
+destination:
+  plugin: entity:node
+  bundle: az_carousel_item
+
+process:
+  type: 
+    plugin: default_value
+    default_value: az_carousel_item
+  langcode:
+    plugin: static_map
+    bypass: true
+    source: language
+    map:
+      und: en
+
+  title:
+    plugin: sub_process
+    source: title_field
+    process:
+      delta: delta
+      value: value
+      format: format
+
+  field_az_link:
+    plugin: sub_process
+    source: field_uaqs_link
+    process:
+      delta: delta
+      uri: url
+      title: title
+
+  field_az_carousel_short_summary:
+    plugin: sub_process
+    source: field_uaqs_summary_short
+    process:
+      delta: delta
+      value: value
+      format:  
+        plugin: default_value
+        default_value: az_standard
+
+  field_az_media_image:
+    plugin: sub_process
+    source: field_uaqs_photo
+    process:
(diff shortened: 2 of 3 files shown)


joeparsons:
The ProboCI migration site is having a fatal error on the front page.  I looked in the watchdog logs and found this error:
```
InvalidArgumentException: The URI '[node:url]' is invalid. You must use a valid URI scheme. in Drupal\Core\Url::fromUri() (line 290 of /src/az-quickstart-scaffolding/web/core/lib/Drupal/Core/Url.php).
```
I think this is happening because one of our D7 demo content carousel items uses the `[node:url]` token as its URL value but [the Drupal 8/9 version of the Token module does not (yet) support link fields](https://www.drupal.org/project/token/issues/3112449).

camikazegreen:
@joeparsons Wasn't there conversation about not supporting [node:url] links on the carousel in QS2? 

It was either that or styling the node page, since we hadn't in QS1.

akslay:
> @joeparsons Wasn't there conversation about not supporting [node:url] links on the carousel in QS2?
> 
> It was either that or styling the node page, since we hadn't in QS1.

Yes, we decided to "do away" with the node display page for Carousel items. It doesn't provide any additional information to users and gets confusing since you're seeing it out of context from the slideshow/carousel. In D7, you were routed to the node page whenever you didn't provide a link. now, we're just not including a link at all if the editor doesn't provide one. And are also excluding the Carousel Item content type from search results so that people don't land on them.

joeparsons:
> Yes, we decided to "do away" with the node display page for Carousel items. It doesn't provide any additional information to users and gets confusing since you're seeing it out of context from the slideshow/carousel. In D7, you were routed to the node page whenever you didn't provide a link. now, we're just not including a link at all if the editor doesn't provide one. And are also excluding the Carousel Item content type from search results so that people don't land on them.

@akslay Thanks.  Do you have any recommendations for what we should do with existing D7 carousel items that are using `[node:url]` as their link value during automated migrations?  Maybe just replace them with `<front>`?

akslay:
> Do you have any recommendations for what we should do with existing D7 carousel items that are using `[node:url]` as their link value during automated migrations? Maybe just replace them with `<front>` or empty?

Is it possible to just not migrate the link if the value is [node:url]? (not familiar with the migration process). I thought `<front>` might be a good option at first, but then realized we would basically be sending users in a loop. If they are on the home page viewing the carousel, click the link on the carousel item, then get routed to the home page, it'll seem like the link is broken or stuck in a loop.

smitahantzios:
> > Do you have any recommendations for what we should do with existing D7 carousel items that are using `[node:url]` as their link value during automated migrations? Maybe just replace them with `<front>` or empty?
> 
> Is it possible to just not migrate the link if the value is `[node:url]`? (not familiar with the migration process). I thought `<front>` might be a good option at first, but then realized we would basically be sending users in a loop. If they are on the home page viewing the carousel, click the link on the carousel item, then get routed to the home page, it'll seem like the link is broken or stuck in a loop.

Yes, we updated the migrations as the comment. If url is with` [node:url] `, not migration the link for the carousel item.

camikazegreen (approved):

camikazegreen:
Post migration, there are two pieces of manual work that will still need to be done:
- Block placement settings (which pages are the carousel displayed on)
- Carousel item order

modules/custom/az_migration/az_migration.info.yml
```diff
@@ -9,3 +9,4 @@ dependencies:
  - drupal:migrate_tools
  - drupal:migrate_drupal
  - drupal:media_migration
+ - az_carousel
```
joeparsons, line 12: Maybe we should omit this (for consistency) and just declare our destination module dependencies in the migrations themselves.
```suggestion
```

joeparsons (dismissed):
This looks good to me with the exception of possibly removing `az_carousel` as a module dependency from the az_migration.info.yml file.

camikazegreen (approved):

joeparsons (dismissed):

trackleft:
I think I am seeing an issue with this PR
image

I am seeing this here: https://b87ff29a-1077-4f73-a6ea-6156a69a0eee--site-migration.probo.build/admin/content

camikazegreen:
I still approve this, but does this in effect mean that the carousel will be enabled for all sites that we migrate?

joeparsons:
> I still approve this, but does this in effect mean that the carousel will be enabled for all sites that we migrate?

@camikazegreen I think that will be the case when @smitahantzios reverts the latest commit in this PR branch but she hasn't done so yet.

modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml
```diff
@@ -0,0 +1,105 @@
...
+    plugin: concat
+    source:
+      - constants/slash
+      - path
+
+dependencies:
+  enforced:
+    module:
+      - az_migration
+      - az_carousel
```
joeparsons, line 99: Since the module dependencies defined in the migrations don't seem to do anything maybe we should just remove them?

smitahantzios: @joeparsons Last time I removed az_carousel dependency, we couldn't see the title for carousel in the content area, so I put that back in.

joeparsons: > @joeparsons Last time I removed az_carousel dependency, we couldn't see the title for carousel in the content area, so I put that back in.

@smitahantzios I'm talking about the module dependencies defined in `modules/custom/az_migration/migrations/migrate_plus.migration.az_node_carousel.yml`.  It seemed like those didn't do anything while `az_carousel` was removed as a dependency from `az_migration.info.yml` (as you said) so I'm wondering if there is any value having them in the migration file at all?

danahertzberg:
Looks like the content all migrated as expected. But the call to action button styling is not working as expected. Is that caused by this PR or is it a separate issue?

camikazegreen:
@danahertzberg I opened the first carousel item and saved it without making any changes and it is now appearing correctly. 
https://e3cb9113-1be8-4637-a94c-754932926582--site-migration.probo.build/node/1

Is there a function that runs when saving that converts it from a link to a button? The contents of the field definitely migrated correctly.

camikazegreen (approved):

tadean (approved):

danahertzberg:
Link styling will be addressed separately.

Ready to merge after security update.

joeparsons (approved):

PR #761, opened 2021-05-19
Closes #760 Use library discovery service to find AZ Barrio theme css.

### LINKED ISSUE #760: Update CKEditor to use site's version of Arizona Bootstrap
Follow up ticket to this PR: https://github.com/az-digital/az_quickstart/pull/754

trackleft:
## Description
Using the library.discovery Service from drupal core
https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Asset%21LibraryDiscovery.php/class/LibraryDiscovery/8.2.x

## Related Issue
#760 

## How Has This Been Tested?
Locally

## Types of changes
- [ x ] New feature (non-breaking change which adds functionality)

## Checklist:
- [ x ] I've added this Pull Request to the AZ Quickstart Github project and set it to "Review in progress".
- [ x ] My code follows the code style of this project.

### DIFF
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -11,6 +11,7 @@ use Drupal\pathauto\PathautoPatternInterface;
 use Drupal\block\Entity\Block;
 use Drupal\Core\Access\AccessResult;
 use Drupal\Core\Session\AccountInterface;
+use Drupal\editor\Entity\Editor;
 
 /**
  * Implements hook_help().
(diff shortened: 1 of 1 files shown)


trackleft:
If az_barrio doesn't exist this will throw an error in the log 

User warning: The following theme is missing from the file system: az_barrio in drupal_get_filename() (line 188 of /app/web/core/includes/bootstrap.inc)

joeparsons:
Looks like AZ Barrio's `hook_library_info_alter()` implementation isn't being evaluated (probably because the admin theme is being used when CKeditor is rendered) so the default local path specified for Arizona Bootstrap in `az_barrio.libraries.yml` is loaded in the CKeditor iframe(s):
```
<link type="text/css" rel="stylesheet" href="/libraries/arizona-bootstrap/css/arizona-bootstrap.min.css?qtdf21=">
```

So this is sort of the inverse (?) of the problem with the fixed asset path used in the `az_barrio.info.yml` file since it always uses the local copy included with Quickstart even though AZ Barrio might be configured to use a different CDN version.

joeparsons:
FWIW, I tried doing very similar to this locally and saw similar results.  I also tried using the `LayoutDiscoveryParser` an `LayoutDiscoveryCollector` services instead since it looks like `hook_library_info_alter()` implementations should be considered within the context of the methods those services provide but the same problem still existed.

trackleft:
> Looks like AZ Barrio's `hook_library_info_alter()` implementation isn't being evaluated (probably because the admin theme is being used when CKeditor is rendered) so the default local path specified for Arizona Bootstrap in `az_barrio.libraries.yml` is loaded in the CKeditor iframe(s):
> 
> ```
> <link type="text/css" rel="stylesheet" href="/libraries/arizona-bootstrap/css/arizona-bootstrap.min.css?qtdf21=">
> ```
> 
> So this is sort of the inverse (?) of the problem with the fixed asset path used in the `az_barrio.info.yml` file since it always uses the local copy included with Quickstart even though AZ Barrio might be configured to use a different CDN version.

Good point.

PR #801, opened 2021-05-24
Wake env before doing site install

kevcooper:
Attempt to resolve kitten install issues by waking the environment and sleeping before running drush site-install

### DIFF
--- a/.github/workflows/review-site.yml
+++ b/.github/workflows/review-site.yml
@@ -97,6 +97,11 @@ jobs:
         run: |
           terminus -y -n auth:login --machine-token ${{ secrets.TERMINUS_TOKEN }}
 
+      - name: Wake env
+        run: |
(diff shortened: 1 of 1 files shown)


joeparsons (approved):
This seems OK to me.  The most recent failure we saw last week happened during the execution of the drush site install command (after the terminus command that that enables SFTP mode) within the "Install site" GH actions step.  It's not clear to me whether the files in the site codebase weren't fully ready/available to drush due to the SFTP mode change not actually being all the way completed or due to the side effects of an earlier GitHub action step not being fully in place but I'd be OK with merging this and then if we still see problems the next step might be try to add a sleep or env:wake command between those two commands within the `Install site` step.

mmunro-ltrr (approved):
Should be harmless to try, & can always get removed again if it doesn't help.

trackleft (approved):

tadean (approved):

PR #820, opened 2021-06-01
Configure PHP default timezone in ProboCI migration site (and pulled main branch).

joeparsons:
## Description
PR against #752 that configures default timezone for Probo migration site (and pulled latest changes from main).

## Related Issue
#135 

## How Has This Been Tested?

Needs to be tested on ProboCI


camikazegreen:
Merged #891 to fix this issue.

PR #836, opened 2021-06-11
Fix files view to use @count instead of 1

### LINKED ISSUE #585: Files list shows incorrect number of “used in” instances (patch)
_SASG found original bug._

Files list shows incorrect number of “used in” instances--always defaulting to “1 place” even when it’s not actually in use.

This is actually a bug in the Files view. There’s no patch available that I found, but there is a simple fix. Click on the “used in” field in the View, and change “1 place” to “@count place”.

https://www.drupal.org/project/drupal/issues/2887988

Related core issue regarding file deletion of unused files: https://www.drupal.org/project/drupal/issues/2821423

akslay:
## Description
Fixes incorrect usage count by using `@count` instead of `1 place`, for items that have a 0 usage count.

Review site: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-836.probo.build/

## Related Issues
- Closes #585
- https://www.drupal.org/project/drupal/issues/2887988
- https://www.drupal.org/project/drupal/issues/2821423

## How Has This Been Tested?
Tested locally and on Probo

## Types of changes

## Checklist:


smayden:
@akslay Is the expected behavior when testing to have an image that is added to site, but not placed on any pages show "zero"?

I added a file here: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-836.probo.build/media/add and did not place on any pages. But then when I go to: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-836.probo.build/admin/content/files it shows as used in two places.

image

akslay:
@smayden That would normally be the behavior, but unfortunately there's a separate Drupal issue(s) about Drupal having the wrong number of file "used in" instances period. Which is why you're still seeing 2 instead of 0. This PR just fixes the view so that items that would normally have a 0 use count don't display `1 place`, because `1 place` was hard-coded in the view settings, rather than using `@count` (which could be 0 or 1 once the other Drupal issues are fixed).
https://www.drupal.org/project/drupal/issues/2887988
https://www.drupal.org/project/drupal/issues/2821423

smayden (dismissed):

akslay:
To-do:

- [ ] Add "place" to the singular result

trackleft (dismissed):

smayden (approved):

camikazegreen (approved):

trackleft (approved):

joeparsons:
@akslay Could you write up the steps to reproduce this problem and add them to the Drupal.org issue?  If this is a separate issue from the more generic file usage issue in core then the Drupal.org issue for this probably needs to be re-opened and we should probably consider contributing a patch.

joeparsons:
If a core patch could be created, I'd also be generally more in favor of adding that to Quickstart instead of maintaining a mostly duplicated version of the default files view...

joeparsons:
After taking a closer look at this issue, I don't think that using `@count` in the singular string for the usage field in the view is actually something we want to do.  The documentation for the [formatPlural() method ](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21StringTranslation%21TranslationManager.php/function/TranslationManager%3A%3AformatPlural/9.2.x) that is actually used by views under the hood states pretty specifically that `@count` should _not_ be used in singular format strings (and that `1` should be used).  I'm afraid this problem is just another symptom/issue being caused by the underlying [file usage calculation issue in core](https://www.drupal.org/project/drupal/issues/2821423) 😞.

akslay:
Are we good to close this issue/PR in favor of waiting for the overall core patch to file count?

akslay:
Closing in favor of waiting for overall core fix.

PR #841, opened 2021-06-14
Closes #736: Adding patch to fix claro dialog with ckeditor_bs_grid plugin

### LINKED ISSUE #736: Improve UX for creating/editing grids
When creating content on the new admissions site, content editors had lots of trouble when it came to using grids

-Difficult to see which column they are editing within the ckeditor without viewing source code (visual of grid outlines only appears on hover and then the columns are stacked not side by side)
-Need default settings for the column width (perhaps auto or equal width?). Numerous editors did not edit the default selection of "None" and then didn't understand why the columns did not appear side by side
-Make column width settings appear with in the dialog box (this is already addressed in #389)
-Possibly reduce the amount of choices or add help text - Content editors did not understand the different options for small, medium, large screen sizes or how best to select those options; It would be great if defaults were similar to QS 1 and then more custom options were part of something similar to advanced settings

image

## Conditions of satisfaction
- [ ] Content editors can easily add a grid (similar to QS1) with default settings
- [ ] Content editors easily know which part of the grid they are editing content for
- [ ] Grid displays as expected on different screen sizes (with default settings)
- [ ] Grid options all appear on screen

trackleft:
## Description
Added patch to Drupal.org for issue https://www.drupal.org/project/ckeditor_bs_grid/issues/3218760
Have not tested if this breaks core, and not sure if we should worry about that.

There is also some discussion about setting equal width as the default option 
https://www.drupal.org/project/ckeditor_bs_grid/issues/3208113
which has been blocked by https://www.drupal.org/project/ckeditor_bs_grid/issues/3193368

## Related Issue
https://github.com/az-digital/az_quickstart/issues/736

## How Has This Been Tested?
Probo and local.

## Types of changes

## Checklist:

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -48,6 +48,7 @@
         "drupal/config_snapshot": "1.0-rc2",
         "drupal/config_sync": "2.0-beta7",
         "drupal/config_update": "1.7",
+        "drupal/console": "^1.9",
         "drupal/core-recommended": "9.1.9",
         "drupal/crop": "2.1",
         "drupal/ctools": "3.6",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -48,6 +48,7 @@
         "drupal/config_snapshot": "1.0-rc2",
         "drupal/config_sync": "2.0-beta7",
         "drupal/config_update": "1.7",
+        "drupal/console": "^1.9",
```
trackleft, line 51:
```suggestion
```

joeparsons (approved):

smayden:
Dana and I discussed this and have a few clarifications on requirements

If you are able to set defaults, please set
- xs, sm, md, and xl to 'none'
- lg to 'equal width'

Is there anyway to make the Large tab be open when content editors add a grid? We think this is where they will need to update settings the most.

Other than that, the fix on the dialog box looks good (kind of squished but works for now). Thanks!

mmunro-ltrr (approved):

danahertzberg (approved):

PR #851, opened 2021-06-18
Closes #783: Migration for D7 views paragraph type.

### LINKED ISSUE #783: Migrate Views paragraph type
See migration mapping doc for details: https://docs.google.com/spreadsheets/d/134UMiYtE8f1YAKYgSukL2JxFzDJF6-TgCAvYFC-vxqQ/edit#gid=0

smitahantzios:
Paragraph Chunks View to az_paragraphs_view Migration

## Description
Paragraph Chunks View to az_paragraphs_view Migration

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_chucks_view.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_chucks_view.yml
@@ -1,5 +1,5 @@
 id: az_paragraph_chucks_view
-label: AZ Pragrapgh Chucks View
+label: AZ Paragraph Chunks View
 migration_group: az_migration
 migration_tags:
   - Quickstart Content Migration
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_chucks_view.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_chucks_view.yml
@@ -1,0 +1,28 @@
+id: az_paragraph_chucks_view
+label: AZ Pragrapgh Chucks View
+migration_group: az_migration
+migration_tags:
+  - Quickstart Content Migration
+status: true
+
+source:
+  plugin: az_paragraphs_item
+  bundle: uaqs_content_chunks_view
+
+destination:
+  plugin: 'entity_reference_revisions:paragraph'
+  default_bundle: az_view_reference
+
+process:
+
+  field_az_view_reference:
+    plugin: paragraphs_chunks_view_display_mapping
+    source: field_uaqs_view
+
+  behavior_settings:
+    plugin: paragraphs_behavior_settings
+
+dependencies:
+  enforced:
+    module:
+      - az_migration
--- a/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
+++ b/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
@@ -20,8 +20,16 @@ class ParagraphsChunksViewDisplayMapping extends ProcessPluginBase {
    */
   public function transform($value, MigrateExecutableInterface $migrate_executable, Row $row, $destination_property) {
 
+    // Collecting the view data field values.
+    $view_data = [];
+
     // Getting view name and display name.
     $view_display = explode("|", $value['vname']);
+
+    // View argument mapping.
+    if ($value['vargs'] != "") {
+      $view_data['argument'] = $value['vargs'];
+    }
 
     $value = [];
     $uaqs_events = [
@@ -81,8 +89,10 @@ class ParagraphsChunksViewDisplayMapping extends ProcessPluginBase {
 
     // Setting Items per page: 6 for 3 Column news block.
     if ($view_display[0] === 'uaqs_news' && $view_display[1] === 'three_col_news_block') {
-      $value['data'] = serialize(['limit' => 6]);
+      $view_data['limit'] = 6;
     }
+
+    $value['data'] = serialize($view_data);
 
     return $value;
   }
--- a/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
+++ b/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
@@ -7,7 +7,7 @@ use Drupal\migrate\ProcessPluginBase;
 use Drupal\migrate\Row;
 
 /**
- * Process Plugin to field merge for Extra Info paragraphs.
+ * Process Plugin to map view and display for Chunks View paragraphs.
  *
  * @MigrateProcessPlugin(
  *   id = "paragraphs_chunks_view_display_mapping"
--- a/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
+++ b/modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
@@ -1,0 +1,84 @@
+<?php
+
+namespace Drupal\az_migration\Plugin\migrate\process;
+
+use Drupal\migrate\MigrateExecutableInterface;
+use Drupal\migrate\ProcessPluginBase;
+use Drupal\migrate\Row;
+
+/**
+ * Process Plugin to field merge for Extra Info paragraphs.
+ *
+ * @MigrateProcessPlugin(
+ *   id = "paragraphs_chunks_view_display_mapping"
+ * )
+ */
+class ParagraphsChunksViewDisplayMapping extends ProcessPluginBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  public function transform($value, MigrateExecutableInterface $migrate_executable, Row $row, $destination_property) {
+
+    // Getting view name and display name.
+    $view_display = explode("|", $value['vname']);
+
+    $value['markup'] = [];
+    $uaqs_events = [
+      'view' => 'az_events',
+      'display' => [
+        'default' => 'page_1',
+        'page' => 'page_1',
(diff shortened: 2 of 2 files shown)


modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
```diff
@@ -0,0 +1,84 @@
...
+
+    $value['markup']['target_id'] = $view_mapping[$view_display[0]]['view'];
+    $value['markup']['display_id'] = $view_mapping[$view_display[0]]['display'][$view_display[1]];
+
+    // Setting Items per page: 6 for 3 Column news block.
+    if ($view_display[0] === 'uaqs_news' && $view_display[1] === 'three_col_news_block') {
+      $value['markup']['data'] = serialize(['limit' => 6]);
+    }
+
+    return $value['markup'];
```
joeparsons, line 81: Seems like the  `['markup']` key isn't really used in the $value variable returned by this function.
```suggestion
    $value = [];
    $uaqs_events = [
      'view' => 'az_events',
      'display' => [
        'default' => 'page_1',
        'page' => 'page_1',
        'list_block' => 'page_1',
        'card_group_block' => 'az_grid',
        'block_1' => 'az_sidebar',
      ],
    ];
    $uaqs_news = [
      'view' => 'az_news',
      'display' => [
        'default' => 'az_grid',
        'three_col_news_block_3' => 'az_grid',
        'three_col_news_block' => 'az_grid',
        'sidebar_promoted_news' => 'az_sidebar',
        'uaqs_teaser_list_page' => 'az_teaser_grid',
        'uaqs_media_list_page' => 'az_paged_row',
        'recent_news_marquee' => 'marquee',
        'recent_news_medium_media_list' => 'az_paged_row',
      ],
    ];
    $uaqs_person_directory = [
      'view' => 'az_person',
      'display' => [
        'default' => 'grid',
        'page' => 'grid',
        'page_1' => 'row',
      ],
    ];
    $uaqs_content_chunks_views_page_by_category = [
      'view' => 'az_page_by_category',
      'display' => [
        'default' => 'grid',
        'page' => 'grid',
        'page_1' => 'row',
      ],
    ];
    $view_mapping = [
      'uaqs_events' => $uaqs_events,
      'uaqs_news' => $uaqs_news,
      'uaqs_person_directory' => $uaqs_person_directory,
      'uaqs_content_chunks_views_page_by_category' => $uaqs_content_chunks_views_page_by_category,
    ];

    $value['target_id'] = $view_mapping[$view_display[0]]['view'];
    $value['display_id'] = $view_mapping[$view_display[0]]['display'][$view_display[1]];

    // Setting Items per page: 6 for 3 Column news block.
    if ($view_display[0] === 'uaqs_news' && $view_display[1] === 'three_col_news_block') {
      $value['data'] = serialize(['limit' => 6]);
    }

    return $value;
```

joeparsons:
As is, this only accounts for D7 paragraphs that reference Quickstart views.  I'm wondering if we should just pass-through the view name and display name for paragraphs referencing a non-Quickstart view or if we should just skip migrating those altogether?

Any thoughts @danahertzberg @tadean?

modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
```diff
@@ -0,0 +1,84 @@
...
+    ];
+    $view_mapping = [
+      'uaqs_events' => $uaqs_events,
+      'uaqs_news' => $uaqs_news,
+      'uaqs_person_directory' => $uaqs_person_directory,
+      'uaqs_content_chunks_views_page_by_category' => $uaqs_content_chunks_views_page_by_category,
+    ];
+
+    $value['target_id'] = $view_mapping[$view_display[0]]['view'];
+    $value['display_id'] = $view_mapping[$view_display[0]]['display'][$view_display[1]];
```
joeparsons, line 74: Maybe something like this to handle non-Quickstart views?

```suggestion
    if (isset($view_mapping[$view_display[0]])) {
      $value['target_id'] = $view_mapping[$view_display[0]]['view'];
      $value['display_id'] = $view_mapping[$view_display[0]]['display'][$view_display[1]];
    }
    else {
      $value['target_id'] = $view_display[0];
      $value['display_id'] = $view_display[1];
    }
```

danahertzberg:
> As is, this only accounts for D7 paragraphs that reference Quickstart views. I'm wondering if we should just pass-through the view name and display name for paragraphs referencing a non-Quickstart view or if we should just skip migrating those altogether?
> 
> Any thoughts @danahertzberg @tadean?

Personally, I see the migrations developed by Arizona Digital as being for Quickstart only. We already aren't accounting for custom content types and the views themselves. I think a firmer line is easier to communicate, but that's just me.

joeparsons:
Do we also need to handle views arguments?

danahertzberg:
> Do we also need to handle views arguments?

I forgot about that.. Are taxonomy terms able to be migrated?

trackleft:
> > Do we also need to handle views arguments?
> 
> I forgot about that.. Are taxonomy terms able to be migrated?

Luckily the arguments field is not an entity reference field so we only need to migrate the text.

joeparsons:
> I forgot about that.. Are taxonomy terms able to be migrated?

I think they should already be (?).  They definitely can be if not.

danahertzberg:
Okay then yes. If terms are migrated with the same ID, then we should definitely migrate the argument field.

trackleft:
For the view display issue, I wonder if we need something like this migrate reports module https://www.drupal.org/project/migrate_report to track all the things that have trouble migrating.

joeparsons:
> Okay then yes. If terms are migrated with the same ID, then we should definitely migrate the argument field.

We probably actually need to do a migration lookup to get the migrated term ID(s)...  @smitahantzios Does that sounds like something you can add?

smitahantzios:
> > Okay then yes. If terms are migrated with the same ID, then we should definitely migrate the argument field.
> 
> We probably actually need to do a migration lookup to get the migrated term ID(s)... @smitahantzios Does that sounds like something you can add?

@joeparsons  I will look into it for sure

smitahantzios:
> > > Okay then yes. If terms are migrated with the same ID, then we should definitely migrate the argument field.
> > 
> > 
> > We probably actually need to do a migration lookup to get the migrated term ID(s)... @smitahantzios Does that sounds like something you can add?
> 
> @joeparsons I will look into it for sure

@joeparsons - We are not keeping the same `tid` that in Drupal 7 database, as per the initial discussions. So for all migrations, we are using the process plugin `migration_lookup` to map. Example as below from person content type migration. Please let me know, if you want more details. Thanks.

```
field_az_person_category:
  - 
    plugin: migration_lookup
    source: field_uaqs_person_category
    migration:
      - az_person_categories
```

modules/custom/az_migration/src/Plugin/migrate/process/ParagraphsChunksViewDisplayMapping.php
```diff
@@ -0,0 +1,90 @@
...
+        'page' => 'grid',
+        'page_1' => 'row',
+      ],
+    ];
+    $uaqs_content_chunks_views_page_by_category = [
+      'view' => 'az_page_by_category',
+      'display' => [
+        'default' => 'grid',
+        'page' => 'grid',
+        'page_1' => 'row',
```
danahertzberg, line 63:
```suggestion
        'default' => 'row',
        'page' => 'row',
        'page_1' => 'grid',
```

trackleft:
@smitahantzios Here is the db view of the view field in QS1

image

smitahantzios:
> @smitahantzios Here is the db view of the view field in QS1
> 
> image

@trackleft - I have gone through the view paragraph got the view argument field. Currently, migrated the view argument as same. Because we can't do the migration_lookup on the migration import now since there is no information about the entity type or information about vocabulary name. 

To lookup the migration table we need `"source_id" `and `"migration name or entity name to type"` but then second information is not available here and that will differ for each views. Please advise.

Other option is that do the migration look up while rendering the view. ( know the entity that is exposing as argument)

trackleft:
Make sense @smitahantzios , most sites don't have as many view paragraphs with view args like the one shown in the database screenshot.  Shouldn't be too difficult for the migration team to figure out after the fact, if they have the source site and the destination site in front of them.  

I think we can move on from this one, and if something dawns on us in the future, we can implement it.

To get the entity reference type, or info, you would have to look in the view contextual filters setttings.

trackleft (dismissed):

smitahantzios:
> Make sense @smitahantzios , most sites don't have as many view paragraphs with view args like the one shown in the database screenshot. Shouldn't be too difficult for the migration team to figure out after the fact, if they have the source site and the destination site in front of them.
> 
> I think we can move on from this one, and if something dawns on us in the future, we can implement it.
> 
> To get the entity reference type, or info, you would have to look in the view contextual filters setttings.

@trackleft  can you provide the DB dump of the Database you are showing in the screenshot above. I need database or instance url  . Views - uanews_stories , covid_19_news uaqs_person_directory & uanews_galleries etc...

modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_chucks_view.yml
```diff
@@ -0,0 +1,28 @@
+id: az_paragraph_chucks_view
+label: AZ Pragrapgh Chucks View
```
tadean, line 2:
```suggestion
label: AZ Paragraph Chunks View
```

danahertzberg:
Follow up ticket for TID work: https://github.com/az-digital/az_quickstart/issues/880

danahertzberg (dismissed):

trackleft (approved):

joeparsons (approved):

danahertzberg (approved):

PR #875, opened 2021-06-23
Migration for Paragraph content HTML to AZ Text

smitahantzios:
Migration for Paragraph content HTML to AZ Text

## Description
Migration for Paragraph content HTML to AZ Text

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_html.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_paragraph_html.yml
@@ -1,0 +1,34 @@
+id: az_paragraph_html
+label: AZ Paragraph HTML
+migration_group: az_migration
+migration_tags:
+  - Quickstart Content Migration
+status: true
+
+source:
+  plugin: az_paragraphs_item
+  bundle: uaqs_content_chunks_html
+
+destination:
+  plugin: 'entity_reference_revisions:paragraph'
+  default_bundle: az_text
+
+process:
+
+  field_az_text_area:
+    plugin: sub_process
+    source: field_uaqs_html
+    process:
+      delta: delta
+      value: value
+      format:  
+        plugin: default_value
+        default_value: az_standard
+  
+  behavior_settings:
+    plugin: paragraphs_behavior_settings
+
+dependencies:
+  enforced:
+    module:
+      - az_migration


smitahantzios:
As per the mapping excel sheet condition, migration for this paragraph having condition to check the source code for each row and dynamic switch between the destination "az_html" or "az_text" paragraph. 

But we can do only single destination for a migration yml. as per that I have mapped currently to "az_text".

Please guide me, if there is option to switch dynamically the destination as per the condition check for each row from the source content.

thanks

tadean (approved):

danahertzberg:
Some pieces get migrated, but don't appear in the entity reference field. Needs further troubleshooting.

tadean:
- Added new text format detection based upon change in parsed html (to better account for whitespace, comments, etc)
- `az_html` paragraph is no longer required. If the module is not enabled, it will fall back on `az_text`

tadean:
My revisions add  a `text_format_recognizer` plugin that attempts to render markup in specified formats and then reparses the html to detect if the original rendering caused changes (e.g. the text format causes data loss). This is used to detect whether or not markup really needs to be full html or not.

**`az_html` is an optional paragraph. If the relevant module is not enabled, it will continue to use `az_text`.**

```
tadean@polaris:~/az_quickstart$ lando drush mim az_paragraph_html
 [notice] Processed 2 items (2 created, 0 updated, 0 failed, 0 ignored) - done with 'az_paragraph_html'
 
tadean@polaris:~/az_quickstart$ lando mysql drupal8
Welcome to the MySQL monitor.  Commands end with ; or \g.
Your MySQL connection id is 172
Server version: 5.7.29 MySQL Community Server (GPL)

mysql> select * from paragraphs_item;
+----+-------------+---------+--------------------------------------+----------+
| id | revision_id | type    | uuid                                 | langcode |
+----+-------------+---------+--------------------------------------+----------+
|  9 |           9 | az_text | 1724e8e9-96c3-4bd4-8074-9a982b889bf1 | en       |
| 10 |          10 | az_html | a13951f0-4c72-4041-89b8-445cc8cd7092 | en       |
+----+-------------+---------+--------------------------------------+----------+
2 rows in set (0.00 sec)
```

**Paragraphs are migrated as `az_html` if they cannot be represented as text format `az_standard` without loss of information. Otherwise, they are migrated as `az_text`.**

tadean:
Ready for review.

PR #899, opened 2021-07-01
Add view_unpublished module

### LINKED ISSUE #844: View unpublished content by other authors
## Problem/Motivation
At the moment, the only way to allow users access to all unpublished content in Drupal (without using a contributed module), is to enable the permission "Bypass content access control". This permission must be used with extreme caution as it has far more consequences than just allowing access to unpublished content. Allowing this permission is not a feasible option.

### Describe the bug
Users without the "Bypass content access control" permission can create unpublished content but other users (except those with the permission) cannot view the node(s) in the content overview page nor the node page itself—they get an access denied. This becomes a problem when more than one user needs to work on unpublished content.

### To Reproduce
Get two users without "Bypass content access control" as a permission
1. Go to '/node/add/page'
2. Create a page and make sure it is unpublished
3. Save the page
4. Have the other user without the permission try to find the page in content overview or go directly to the node page to try to access
5. Result: node will not show for them in the content overview page and they will get an access denied if trying to view directly.

## Proposed resolution
This is a problem with Drupal core, but there may be a contributed module which will help close the gap such as https://www.drupal.org/project/view_unpublished

### Expected behavior
Editors on a site should be able to collaborate on unpublished nodes.

## Additional context
- https://www.drupal.org/docs/7/managing-users/viewing-unpublished-content#s-drupal-7 (says D7, but still a problem)

akslay:
## Description
Adds the View Unpublished module so that all editors can see unpublished content.

### Testing Instructions
Review site: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-899.probo.build/

1. Log in as a `content admin` add a new page that's unpublished, and save it
2. Log out
3. Log in as `content editor`
4. Navigate to the Content list and verify you can see the unpublished page that was just added

## Related Issue
Closes #844 

Related core issue: https://www.drupal.org/project/drupal/issues/2971902

## How Has This Been Tested?
Tested locally and on probo

## Types of changes

## Checklist:

### DIFF
--- a/az_quickstart.info.yml
+++ b/az_quickstart.info.yml
@@ -58,6 +58,7 @@ install:
   - drupal:field_ui
   - drupal:file
   - drupal:rdf
+  - drupal:view_unpublished
   - drupal:views
   - drupal:views_ui
   - drupal:tour
--- a/composer.json
+++ b/composer.json
@@ -80,6 +80,7 @@
         "drupal/smart_title": "1.0-beta1",
         "drupal/smtp": "1.0",
         "drupal/token": "1.9",
+        "drupal/view_unpublished": "1.0",
         "drupal/views_bootstrap": "4.3",
         "drupal/viewsreference": "2.0-beta2",
         "drupal/webform": "6.0.4",


smayden (approved):
Tested as described in this PR and worked as expected. Thanks!

I noticed a few things while testing that might need separate tickets (if they don't exist already):
- Content editors can now view unpublished content created by admin/content admins, but they don't have any option to publish it for page, news, events, and person
- Similarly content editors can create content, but it is published by default for all four content types (no publish/unpublish checkbox appears for content editors)
- When viewing unpublished content there is no indicator at all that the page is unpublished.

trackleft (approved):

tadean:
- [ ] Add unpublished view permissions to install profile
- [ ] This module enables per-node access permissions. Are there any similar modules that might not take this step?

akslay:
We will use "Content Moderation" in core instead--it adds the necessary permissions without changing permissions to a per-node basis. Make a new branch for this change, but use the same issue.

PR #903, opened 2021-07-02
Add Content Moderation to install profile

### LINKED ISSUE #844: View unpublished content by other authors

akslay:
## Description
Adds the Content Moderation (in core) module to the install profile in order to address content permissions issue. By enabling the module, you grant "view unpublished" permissions to editors without modifying permissions to a per node basis. This method keeps permissions on a per node type basis. Enabling the module does not enable content workflows--you must manually enable workflows to specific content types.

## Related Issue
Closes #844 

## How Has This Been Tested?
Locally and on Probo

## Types of changes

## Checklist:

### DIFF
--- a/az_quickstart.info.yml
+++ b/az_quickstart.info.yml
@@ -31,6 +31,7 @@ install:
   - drupal:ckeditor
   - drupal:color
   - drupal:config
+  - drupal:content_moderation
   - drupal:contextual
   - drupal:menu_link_content
   - drupal:datetime


camikazegreen (approved):
This is so simple!
It resolves the issue with no other identified changes to moderation states or other types of regressions.

smayden (approved):
yay!

Should we rebuild to try and take care of the php unit test failing?

PR #915, opened 2021-07-09
Closes #595: Add az_security module

### LINKED ISSUE #595: Investigate Security Recommendations for Webforms.
Webforms are one of the most common areas for security vulnerabilities in Drupal sites. The purpose of this spike is to investigate:
- What is the current state of security concerns for the Drupal 9 version of webform?
- Which types of configuration should be put in to place to mitigate any risks?
- How should we implement adding this configuration to Quickstart?

## Conditions of satisfaction
- [ ] The considered options are documented
- [ ] A recommended next step is documented
- [ ] An estimate of time and effort is provided for the recommendation

kevcooper:
## Description
This pr adds a custom modue, az_security, with config for honeypot and seckit

## Related Issue
#595 

## Changes
- Add `honeypot` module and default config
  - Form expiration: 10 minutes
  - Time limit (minimum time before submission allowed) 5 seconds
- Add `seckit` module and default config
  - Add `Strict-Transport-Security` with max-age of 30 days
  - Add `Referrer-Policy` header with value of `strict-origin-when-cross-origin`.
  - Add `Content-Security-Policy` with value of `default-src *; report-uri https://api.web.arizona.edu/v1/azqs/report; report-to azqs`
  - Add `Report-To` with a value of `{"group":"azqs","max_age":2592000,"endpoints":[{"url":"https://api.web.arizona.edu/v1/azqs/report","include_subdomains":false}]}`

## Header information
The notable security header changes included in this PR are as follows:
- Establish an initial default `Content-Security-Policy-Report-Only` header
  - The proposed default will be `default-src *` which will **disallow** the use of any inline scripts, styles, fonts, media, and the javascript `eval` function, but will **allow** remote loading of assets from any domain. The purpose of this is to mitigate most cases of XSS where some user input was not properly sanitized. We may want to restrict domains in the future as improper sanitation could still lead to XSS via a malicious asset being loaded remotely.
  - The `report-uri` defined in this default will receive reports of any instance of this policy being violated. At the start we will be using the `-Report-Only` variant of Content Security Policy, and as such the browser will not actually prevent policy violations and instead will only report them. Adding `report-to azqs` to the end of our `report-uri` header will tell browsers to switch to the endpoint defined in the `azqs` group of the `Report-To` header as browsers start to adopt the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API).
- Establish an initial value for the proposed `Report-To` header (see [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API)). While this header is experimental and not in use for the most part, the intent is for it to replace CSP's `report-uri` as well as receiving other interesting reports generated by browsers such as crashes, deprecated feature usage, and various other browser interventions.
- Establish a standard `Referrer-Policy` of `strict-origin-when-cross-origin`. This is [Chrome's new default](https://web.dev/referrer-best-practices/) so any concerns about loss of analytics information will have to contend with the fact that a significant percentage of users will be using this policy even if we don't define it ourselves.

## Checklist:

## TODO
- [x] Remove webform as dependency
- [x] Change module name (az_security?)
- [x] Add seckit
- [ ] Check calendar widget functionality - this should no longer be a concern as we've removed antibot
- [x] Disable honeypot for authenticated users
- [ ] Investigate what information will be lost with `strict-origin-when-cross-origin`
- [ ] Investigate what happens when HSTS is send behind a load balancer
- [ ] Investigate what kinda of information may be disclosed via CSP and other reports


mmunro-ltrr (dismissed):
To fully see the implications of this, the only way would be to enable the module on some live sites (then follow up as needed, but keep an eye on it initially).

modules/custom/az_security/src/EventSubscriber/AZSecuritySubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+    $headers = $this->config->get('headers') ?? [];
+
+    if ($headers['report_to']['enabled'] ?? FALSE) {
+      $response->headers->set(
+        'Report-To',
+        trim(json_encode($headers['report_to']['objects'], JSON_UNESCAPED_SLASHES), '[]')
+      );
+    }
+
+    if ($headers['other']['enabled'] ?? FALSE) {
```
trackleft, line 47: What does this mean?

trackleft: if empty, false?

trackleft (dismissed):

tadean (approved):

joeparsons (approved):

PR #942, opened 2021-07-16
Closes #941 No margin between nodes in az-row view mode when on flexible pages or if a listing of flexible pages. 

### LINKED ISSUE #941: No margin between nodes in az-row view mode when on flexible pages or if a listing of flexible pages.
### Describe the bug
https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/css/style.css#L205-L207 is supposed to cover all non-flexible pages, but there are times when we embed nodes in flexible pages, or list flexible pages in alternate view modes from full page.

### To Reproduce
https://live-azs-demo.pantheonsite.io/events/annual-events
http://live-azs-quickstart.pantheonsite.io/live-training
https://live-azs-demo.pantheonsite.io/page-list

## Proposed resolution
I'd suggest either scoping margin to the view mode

```
.node--view-mode-az-row .node__content {
  margin-bottom: 1rem;
}
```

Or, editing each view mode that needs it, to add an `mb-2` class on node.

trackleft:
## Description
made no margin fix specific to full page
## Related Issue
#941

### DIFF
--- a/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
+++ b/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
@@ -2,7 +2,7 @@
 *  Page-specific overrides to the theme defaults.
 */
 
-.node--type-az-flexible-page .node--view-mode-full .node__content {
+.node--type-az-flexible-page article.node--view-mode-full:first-child {
   /* Remove extra space at the top of content. */
   margin-top: 0;
 }
--- a/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
+++ b/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
@@ -2,7 +2,7 @@
 *  Page-specific overrides to the theme defaults.
 */
 
-.node--type-az-flexible-page .node__content {
+.node--type-az-flexible-page .node--view-mode-full .node__content {
   /* Remove extra space at the top of content. */
   margin-top: 0;
 }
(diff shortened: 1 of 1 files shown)


modules/custom/az_flexible_page/css/az_flexible_page.theme.css
```diff
@@ -2,7 +2,7 @@
 *  Page-specific overrides to the theme defaults.
 */
 
-.node--type-az-flexible-page .node__content {
+.node--type-az-flexible-page article.node--view-mode-full:first-child {
```
trackleft, line 5: This isn't specific enough for this page: https://live-azs-demo.pantheonsite.io/pages/accordions since the node__content class is overriding it.

trackleft: Try 

.node--type-az-flexible-page article.node--view-mode-full .node__content:nth-of-type(1)

mmunro-ltrr: Any reason not to simply use the child combinator `>` & keep node__content in the rule?

trackleft: The flexible page with the list of flexible pages 

structured like 

-.node--type-az-flexible-page
--article.node--view-mode-full
--- .node__content
---- node__content
---- node__content
---- node__content

mmunro-ltrr: Just checked that the https://live-azs-demo.pantheonsite.io/pages/accordions case was unfortunately indistinguishable from the one that should get the override (so much for that idea).

trackleft:
Closing in favor of #945

Issue #916, opened 2021-07-09
The fix for Issue #852 (Extra spacing above body content) removed margins where they are needed.


mmunro-ltrr:
## Problem/Motivation
A necessary bugfix, removing `margin_top` where it wasn't required, also removed it where it was required. 

### Describe the bug
The deleted Bootstrap Barrio node.css stylesheet provided a `node__content` CSS class used in several places.

### To Reproduce

1. Create several Events.
2. Display the `/calendar` view showing them.
3. Note that the events run together with no spacing between them (see https://github.com/az-digital/az_quickstart/pull/910#issuecomment-875960824 for the initial report).

## Proposed resolution
Remove all usage of the `node__content` class, replace it with an equivalent where a margin is required.

### Expected behavior

Restore the behavior before https://github.com/az-digital/az_quickstart/pull/865

Issue #852, opened 2021-06-18
Extra spacing above body content


danahertzberg:
## Motivation
I want my content with colored or media background to touch the main menu if it is the first element on the page.

## Example
![image.png](https://images.zenhubusercontent.com/5eb9d10a2b1f2c51836b60e3/83fcda89-40be-4305-807a-31c94f48230d)

danahertzberg:
Potential issue is `node__content` class adding margin-top

mmunro-ltrr:
PR #921 re-introduced this bug.

PR #945, opened 2021-07-19
Closes #941: Make the selectors in the CSS rule applying the margin-top override very specific.

### LINKED ISSUE #941: No margin between nodes in az-row view mode when on flexible pages or if a listing of flexible pages.

mmunro-ltrr:
## Description
Attempts to remove the margin-top from elements with the class `node__content` only when they are at the top of the main content area of a Page.

## Related Issue
Another iteration in the series of issues #852, #916 ...

## How Has This Been Tested?
On a local Lando build, created some az_flexible_page nodes, verified that there was no margin between the navigation menu and the top of a full-width text on media paragraph at the top of the page content, verified that margins did appear between entries in the `/page-list` view.

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
+++ b/modules/custom/az_flexible_page/css/az_flexible_page.theme.css
@@ -2,7 +2,7 @@
 *  Page-specific overrides to the theme defaults.
 */
 
-.node--type-az-flexible-page .node__content {
+#block-az-barrio-content > .content > .node--type-az-flexible-page > .node__content {
   /* Remove extra space at the top of content. */
   margin-top: 0;
 }


smayden (approved):
Tested pages with text on media and text with background at top of page and no longer see the gap between the menu and content described in #852.

Also tested things like page-list and event listings (on /calendar and full calendar view on a page) and am no longer seeing the spacing between items reviewed.

So looks good to me!

joeparsons (approved):

trackleft (approved):
Looks good to me, the `>` just means it has to be a child of a specific parent and not just a descendent, right?

Tested with this structure
```
- #block-az-barrio-content
-- .content
--- .node--type-az-flexible-page
---- .node__content
----- .node__content
```
And it seems to work, and not removing the margin-top from the last node__content.  Here is my test https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-945.probo.build/test, but could probably also be tested by embedding a node in a flexible page via text-area.

Issue #966, opened 2021-08-02
Tab Crashes in Probo Build Chromedriver


tadean:
### Describe the bug
The following crash has been seen to happen in new PRs:
```
WebDriver\Exception\UnknownError: unknown error: session deleted because of page crash
from unknown error: cannot determine loading status
from tab crashed
  (Session info: headless chrome=90.0.4430.93)
  (Driver info: chromedriver=90.0.4430.24 (4c6d850f087da467d926e8eddb76550aed655991-refs/branch-heads/4430@{#429}),platform=Linux 4.15.0-51-generic x86_64)
```
This seems to be related to chromedriver version, or possible some resource exhaustion in the container.

### To Reproduce
Probo build

PR #967, opened 2021-08-02
WIP Probo Chromedrive Tab Crash Test

tadean:
## Description
This is a WIP in progress PR to investigate chromedriver troubles in probo builds as seen in #966 

## Related Issue
#966 

## Types of changes

## Checklist:

### DIFF
--- a/.probo.yaml
+++ b/.probo.yaml
@@ -56,6 +56,7 @@ steps:
       - chown www-data:www-data -R /var/www/html/sites/simpletest
       - mkdir -p /var/www/html/sites/default/files/simpletest
       - chown www-data:www-data -R /var/www/html/sites/default/files/simpletest
+      - echo $(/usr/bin/google-chrome --product-version | cut -d . -f 1-3)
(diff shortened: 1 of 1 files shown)


trackleft (approved):

joeparsons (approved):
Nice work. Thanks @tadean!

tadean:
The implication of this PR is that chromedriver will be using disk space as its source of memory rather than actual RAM in the docker image, as it appears that our page environment has gotten to the point where it is beyond the limit of shared memory for the docker image from the host. This is sometimes a pretty small amount, and chromedriver was crashing in Probo because it ran out of memory to render tabs.

The tradeoff (as @trackleft pointed out) is that this usage is much slower. However, given that we're not likely to have a huge number of javascript tests compared to regular browser tests, perhaps that's acceptable.

joeparsons:
> The tradeoff (as @trackleft pointed out) is that this usage is much slower. However, given that we're not likely to have a huge number of javascript tests compared to regular browser tests, perhaps that's acceptable.

Seems like an acceptable compromise to me, especially for Probo.

mmunro-ltrr (approved):
It would be interesting to see how the change benchmarks (times relative to old runs).

Issue #964, opened 2021-08-02
Update manager module being installed on kitten review site


joeparsons:
## Problem/Motivation
One of the arguments passed to the drush command used in our GitHub action for updating the Kitten review site is meant to prevent the core update manager module from being installed but it is not working as expected.

### Describe the bug
The drush command we're using to install Quickstart on the kitten review site passes an argument for core's install_configure_form but the value passed is based on the D7 form, not the D9 version.

PR #969, opened 2021-08-04
#964: Fix broken drush site:install command (again).

joeparsons:
## Description
The fix for #964 that we merged this morning added an extra line to the multi-line shell command without adding the necessary `\` to the preceding line so now the `drush site:install` command is totally broken 🤦🏻 .

## Related Issue
#964


trackleft (approved):

mmunro-ltrr (approved):
The only way to test it....

joeparsons:
Actually the `drush site:install` command is working but the tries running dangling `--verbose` as its own step and so the GH actions workflow step fails.

Issue #990, opened 2021-08-18
Add responsive image styles to be used in the text with media paragraphs.


trackleft:
## Motivation
As a content creator/editor I would like to be able to use different renditions of the same image depending on how my content is being consumed. For example, if someone is browsing my site on a mobile device with a small screen and limited bandwidth, i don't want to send them a giant image meant for desktop viewing and vice versa.

## Proposed Resolution
I propose we enable the picture module.
Starting with the image style being used for the text on media paragraph item, and the bottom display mode, allow 4 image styles attached to a background image picture group.

Image styles:
full-width-xs -- scaled to 575px
full-width-sm --  scaled to 767px
full-width-md -- scaled to 991px
full-width-lg -- scaled to 1199px
Picture group
full-width-bg

Issue #693, opened 2021-04-27
Image focal point module


danahertzberg:
## Motivation
1. I want to keep a specific part of the image in view when using the Text on Media paragraph type.
2. I want to keep the person's face in frame when switching displays of people.

## Proposed Resolution
Could look into focal point module: https://www.drupal.org/project/focal_point

danahertzberg:
Related to #656

akslay:
I looked in to this, and the Focal Point module utilizes a field widget to make focal points available to editors. However, we are already utilizing the ImageWidget Crop module (which is also a widget) to expose cropping to editors. Seeing as we can only select one widget to use for an image field, it would have to be one or the other.

Will have to see if there's other ways to offer both levels of functionality.

PR #992, opened 2021-08-19
Closes #989 edit media from the parent referencing form.

### LINKED ISSUE #989: Add ability to edit media from the parent/referencing entity form.
## Motivation
As a content editor, I would like to edit media on the page or paragraph that uses it.
As it stands, I can only edit media by tracking it down somewhere in the `/admin/content/media` interface.

## Proposed Resolution
Add the https://www.drupal.org/project/media_library_edit module

trackleft:
Adding drupal.org/project/media_library_edit module
Testing can be done here:
https://app.probo.ci/#/dashboard/build/65e02ce3-e071-4d7f-a80b-a6d9680976de/task/5ee0fee94d400004
Media items now look like this: 
image

## Related Issue
Closes #989
Blocks #693 
Blocks #990

## How Has This Been Tested?
Locally ran `lando install`

## Types of changes

## Checklist:


smayden (commented):
Tested successfully as admin and content admin, but don't get the edit option when testing a content editor. Permission issue? I'm assuming content editors need access too.

trackleft:
Thanks for testing @smayden, It is probably a permission issue, would the solution be to add the appropriate permissions to just the content editor role?

There are two options that might be the culprit, but my suspicion is that it is `Edit any media`.  
image

Could also be the `Update any media` permission, but only the administrator has that permission currently.

smayden:
Agree that it is probably that "edit any media" permission. Might make sense to address as a separate issue? I would think we'd want content editors to be able to edit any media for all media types.

trackleft:
FYI there is this patch that we might consider for editing media on node create https://www.drupal.org/project/media_library_edit/issues/3189250

trackleft:
Additionally, this patch says the module only checks permissions on the first image, which isn't a problem until we try to edit photo galleries and someone else uploaded the first image. https://www.drupal.org/project/media_library_edit/issues/3201011

smayden (approved):

trackleft:
FYI, I did some looking into the `hide_date: true` setting that appeared by accident within this pull request when I pulled configuration out of my local site, and found that we have a discrepancy between the config for form display for event  in the quickstart folder for az_seo, and the config in az_event.

In az_seo the setting is:  `hide_date: true`
https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_seo/config/quickstart/core.entity_form_display.node.az_event.default.yml#L102

In az_event the setting is `hide_date: false`
https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_event/config/install/core.entity_form_display.node.az_event.default.yml#L92

image


I believe we haven't noticed this before because az_seo is enabled by default on most of our sites and test sites, but the correct setting is what appears in az_seo. `hide_date: true`

trackleft:
This is ready for review.

joeparsons (approved):

smayden (approved):

PR #1038, opened 2021-09-22
Closes #446: Summary link field(s) cannot use tokens.

### LINKED ISSUE #446: Summary link field cannot use tokens
In D7 several content types have the ability to add a "summary link" that is used to point to a different site/url when that item appears in different view modes (e.g. takes a news story outside the site). By default the link field points to [node:url] to take users to the node.

For the news content type, we were unable to create the summary link field since in D9 you cannot add a token for the default on the field. In the 1/13/21 we decided to remove from news content type and explore solutions as a separate issue.

trackleft:
## Description
This PR is the result of a group workshop on Wednesday Sept 22 2021

Bypass need for using tokens in summary link field in favor of adding our own token `[az-canonical-url]` that checks if the field is empty, and if it is use the node url, otherwise use the field value.  This token is then used in the Field Group Link formatter (custom url setting) on each display mode for each content type that needs to house internal/external content.

Still need to reexport all the display modes and add the field_az_link field to each content type that needs it.

Relevant links https://git.drupalcode.org/project/field_group_link/-/blob/8.x-3.x/src/Plugin/field_group/FieldGroupFormatter/Link.php#L203-213

## Related Issue
#446

## How Has This Been Tested?
Locally

## Types of changes
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -5,6 +5,7 @@
  * Contains az_core.module.
  */
 
+use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Url;
 use Drupal\pathauto\PathautoPatternInterface;
@@ -109,8 +110,8 @@ function az_core_token_info() {
     'tokens' => [
       'node' => [
         'az-canonical-url' => [
-          'name' => t("Canonical URL"),
-          'description' => t("Switches between field_az_link and node:url."),
+          'name' => t("AZ Canonical URL"),
+          'description' => t("Returns the URI value of field_az_link if present on the node or the default node URI if not."),
         ],
       ],
     ],
@@ -120,9 +121,9 @@ function az_core_token_info() {
 /**
  * Implements hook_token().
  */
-function az_core_tokens($type, $tokens, array $data, array $options, \Drupal\Core\Render\BubbleableMetadata $bubbleable_metadata) {
+function az_core_tokens($type, $tokens, array $data, array $options, BubbleableMetadata $bubbleable_metadata) {
   $replacements = [];
-  if ($type == 'node') {
+  if ($type === 'node') {
     foreach ($tokens as $name => $original) {
       // Find the desired token by name.
       switch ($name) {
@@ -137,13 +138,23 @@ function az_core_tokens($type, $tokens, array $data, array $options, \Drupal\Cor
 }
 
 /**
- * @TODO internal paths don't work.
+ * Token replacement callback providing the canonical URI for a node.
+ *
+ * Returns the URI value of field_az_link if present on the node or the default
+ * node URI if not.
+ *
+ * @param array $data
+ *   Node data from hook_tokens().
+ *
+ * @return string
+ *   The node's canononical URI.
  */
-function az_core_canonical_url($data) {
+function az_core_canonical_url(array $data) {
   /** @var \Drupal\node\NodeInterface $node */
   $node = $data['node'];
-  $url = \Drupal\Core\Url::fromRoute('entity.node.canonical', ['node' => $node->id()]);
-  $uri = $url->getUri();
+  /** @var \Drupal\Core\Url $url */
+  $url = $node->toUrl();
+  $uri = $url->toUriString();
 
   if (!empty($node->field_az_link[0]->uri)) {
     $uri = $node->field_az_link[0]->uri;
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   }
   return AccessResult::neutral();
 }
+
+
+/**
+ * Implements hook_token_info().
+ */
+
+function az_core_token_info() {
+  return [
+    'tokens' => [
+      'node' => [
+        'az-canonical-url' => [
+          'name' => t("Canonical URL"),
+          'description' => t("Switches between field_az_link and node:url."),
+        ],
+      ],
+    ],
+  ];
+}
+
+/**
+ * Implements hook_token().
+ */
+
+function az_core_tokens($type, $tokens, array $data, array $options, \Drupal\Core\Render\BubbleableMetadata $bubbleable_metadata) {
+  $replacements = [];
+  if ($type == 'node') {
+
+    foreach ($tokens as $name => $original) {
+      // Find the desired token by name.
+      switch ($name) {
+        case 'az-canonical-url':
+          $replacements[$original] = az_core_canonical_url($data);
+          break;
+      }
+    }
+  }
+  return $replacements;
+
+}
+/**
+ * @TODO internal paths don't work.
+ */
+function az_core_canonical_url($data) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $data['node'];
+  $options = ['absolute' => TRUE];
+  $url = \Drupal\Core\Url::fromRoute('entity.node.canonical', ['node' => $node->id()], $options);
+  $default_url = $url->toString();
+  $url = $default_url;
+
+  if (!empty($node->field_az_link[0]->uri)){
+    try {
+      $url = Url::fromUri($node->field_az_link[0]->uri);
+      $url = $url->toString();
+      } catch (\InvalidArgumentException $e) {
+        $url = $default_url;
+      }
+  }
+
+  return $url;
+}
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -101,11 +101,9 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   return AccessResult::neutral();
 }
 
-
 /**
  * Implements hook_token_info().
  */
-
 function az_core_token_info() {
   return [
     'tokens' => [
@@ -122,11 +120,9 @@ function az_core_token_info() {
 /**
  * Implements hook_token().
  */
-
 function az_core_tokens($type, $tokens, array $data, array $options, \Drupal\Core\Render\BubbleableMetadata $bubbleable_metadata) {
   $replacements = [];
   if ($type == 'node') {
-
     foreach ($tokens as $name => $original) {
       // Find the desired token by name.
       switch ($name) {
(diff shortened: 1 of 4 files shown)


modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+        ],
+      ],
+    ],
+  ];
+}
+
+/**
+ * Implements hook_token().
+ */
+
```
joeparsons, line 125:
```suggestion
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   }
   return AccessResult::neutral();
 }
+
+
+/**
+ * Implements hook_token_info().
+ */
+
```
joeparsons, line 108:
```suggestion
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+        case 'az-canonical-url':
+          $replacements[$original] = az_core_canonical_url($data);
+          break;
+      }
+    }
+  }
+  return $replacements;
+
+}
+/**
```
joeparsons, line 142:
```suggestion

/**
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+    try {
+      $url = Url::fromUri($node->field_az_link[0]->uri);
+      $url = $url->toString();
+      } catch (\InvalidArgumentException $e) {
+        $url = $default_url;
+      }
+  }
+
+  return $url;
+}
```
joeparsons, line 163:
```suggestion
}

```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+  $default_url = $url->toString();
+  $url = $default_url;
+
+  if (!empty($node->field_az_link[0]->uri)){
+    try {
+      $url = Url::fromUri($node->field_az_link[0]->uri);
+      $url = $url->toString();
+      } catch (\InvalidArgumentException $e) {
+        $url = $default_url;
+      }
```
joeparsons, line 159:
```suggestion
    } catch (\InvalidArgumentException $e) {
      $url = $default_url;
    }
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+}
+/**
+ * @TODO internal paths don't work.
+ */
+function az_core_canonical_url($data) {
+  /** @var \Drupal\node\NodeInterface $node */
+  $node = $data['node'];
+  $options = ['absolute' => TRUE];
+  $url = \Drupal\Core\Url::fromRoute('entity.node.canonical', ['node' => $node->id()], $options);
+  $default_url = $url->toString();
```
joeparsons, line 150: Since the field formatter provided by Field Group Link is [actually expecting a URI](https://git.drupalcode.org/project/field_group_link/-/blob/8.x-3.x/src/Plugin/field_group/FieldGroupFormatter/Link.php#L205-211) that it will generate a URL from, I think we might want to do something like this here:
```suggestion
  $default_url = $url->getUri();
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+  $url = $default_url;
+
+  if (!empty($node->field_az_link[0]->uri)){
+    try {
```
joeparsons, line 154: Is this try/catch block even needed?  Seems the Field Group Link formatter is [already doing this](https://git.drupalcode.org/project/field_group_link/-/blob/8.x-3.x/src/Plugin/field_group/FieldGroupFormatter/Link.php#L205-211).

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
   }
   return AccessResult::neutral();
 }
+
```
joeparsons, line 103:
```suggestion
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+}
+
+/**
+ * Implements hook_token().
+ */
+
+function az_core_tokens($type, $tokens, array $data, array $options, \Drupal\Core\Render\BubbleableMetadata $bubbleable_metadata) {
+  $replacements = [];
+  if ($type == 'node') {
+
```
joeparsons, line 129:
```suggestion
```

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,64 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+      // Find the desired token by name.
+      switch ($name) {
+        case 'az-canonical-url':
+          $replacements[$original] = az_core_canonical_url($data);
+          break;
+      }
+    }
+  }
+  return $replacements;
+
```
joeparsons, line 140:
```suggestion

    return $replacements;
```

joeparsons (changes requested):

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,61 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+  if (!empty($node->field_az_link[0]->uri)){
+    try {
+      $url = Url::fromUri($node->field_az_link[0]->uri);
+      $url = $url->toString();
+    } catch (\InvalidArgumentException $e) {
+      $url = $default_url;
+    }
+  }
+
+  return $url;
```
joeparsons, line 159: Since the field formatter provided by Field Group Link is [actually expecting a URI](https://git.drupalcode.org/project/field_group_link/-/blob/8.x-3.x/src/Plugin/field_group/FieldGroupFormatter/Link.php#L205-211) that it will generate a URL from, I think we might want to do something like this here:
```suggestion
  $url = \Drupal\Core\Url::fromRoute('entity.node.canonical', ['node' => $node->id()]);
  $uri = $url->getUri();

  if (!empty($node->field_az_link[0]->uri)) {
    $uri = $node->field_az_link[0]->uri;
  }

  return $uri;
```

joeparsons: @tadean Does this change seem reasonable to you?

modules/custom/az_core/az_core.module
```diff
@@ -100,3 +100,61 @@ function az_core_block_access(Block $block, $operation, AccountInterface $accoun
...
+      // Find the desired token by name.
+      switch ($name) {
+        case 'az-canonical-url':
+          $replacements[$original] = az_core_canonical_url($data);
+          break;
+      }
+    }
+  }
+
+    return $replacements;
```
joeparsons, line 136:
```suggestion
  return $replacements;
```

modules/custom/az_core/az_core.module
tadean: @joeparsons Yes, looks good to me. The field group does seem to do its own exception handling so the try/catch shouldn't be needed here. This is much cleaner.

joeparsons:
This is ready for review now.  There are some remaining tasks that maybe(?) should be in a separate ticket/PR:
- [ ] Add link field to news content type
- [ ] Add field group link to news content type

Also, PSA: I had to manually omit some changes to the `az_event` field display configs that were coming from `az_seo` being enabled by default.

joeparsons (approved):

trackleft (commented):
Nice job remembering the demo content!

smayden:
This could be done as a separate task later, but I'm wondering if the Link text field for Link to Event has to be required when a url is present? In testing I was required to enter text even though it isn't used anywhere on the events.

image

We also should probably edit the help text. It still refers to the token and could be made clearer.

trackleft:
@smayden do these look like the proper settings? 
image

smayden:
For events, I think disabled might make the most sense @trackleft since the text isn't used in any of the views as far as i can tell.

I need to investigate whether that text is used for news or not (hoping not, if so we would probably need a default value)

trackleft:
The current help text is 
```If you want to link to this event on [site:name], leave the token in the URL field, if you want to link to the event hosted elsewhere, such as on the master calendar, replace the token with a link.```

Maybe?
```If you want listings of this event that appear on [site:name], leave this URL field blank, if you want to list this event but link to it hosted elsewhere, such as on the master calendar, fill the URL field with an absolute URL.```

trackleft:
> For events, I think disabled might make the most sense @trackleft since the text isn't used in any of the views as far as i can tell.
> 
> I need to investigate whether that text is used for news or not (hoping not, if so we would probably need a default value)

FYI I believe the link text could be used for accessible aria-label attributes.

smayden:
> The current help text is
> If you want to link to this event on [site:name], leave the token in the URL field, if you want to link to the event hosted elsewhere, such as on the master calendar, replace the token with a link.
> 
> Maybe?
> If you want listings of this event that appear on [site:name] to link to this node's content, leave this URL field blank, if you want to list this event but link to it hosted elsewhere, such as on the master calendar, fill the URL field with an absolute URL.

How about:
In most cases, leave this field blank. Enter an external URL if you want the calendar listings for this event to link directly to event information on another website (e.g. the master calendar).

We could consider relabeling the field "Link to External Event"

trackleft:
That's way better than what I said haha

smayden:
> FYI I believe the link text could be used for accessible aria-label attributes. Unless we are adding an accessible link text field. Though I'm not sure what the best way to handle link text is when their is metadata about event date in the calendar listing that should probably be read aloud when viewing the calendar instead of requiring people who use screen readers to always click through to events to find out when it is.

Wouldn't we want the Event title to be the accessible aria-label (just as I assume it would be for cases where the URL field is left blank)?

trackleft:
Sure, that could work, I was conflating event cards with clickable cards with no event titles.

mmunro-ltrr (approved):
Tried some quick messing around with the Event links on a local Lando build: it definitely needs some follow-up issues addressed (like requiring the link text even if it's unused), but would be good to merge as a staring point.

smayden (approved):
I'm fine moving forward with this and addressing issues related to help text and disabling/making url field optional in a separate ticket when we do news

Issue #1005, opened 2021-08-25
News default/page display renders media entity via Large view mode, causing duplicate captions.


trackleft:
## Problem/Motivation
Captions can appear twice under certain conditions.

### Describe the bug
While building the news migration, Z found that the news content type is using the large display mode for media, meaning that if media is referenced from the media library and already has a caption on it, you will see the caption on the media entity itself and the caption field on the news content type.

### To Reproduce
Create an image media entity
add a caption
Create news content
reference created media entity
Add Caption to news for the feature image.
publish news
view both captions.

## Proposed resolution
Add media view mode to be referenced by news items exclusively, or call it az_large_no_meta.

### Expected behavior

Feature media caption should only be settable on the create news page itself

PR #1108, opened 2021-10-27
Closes: #1005 Refactoring captions on news to use image caption

smayden:
## Description
Based on our earlier discussion about what to do with the caption field on news, this PR refactors the news content type to use the caption field from the image instead.

- Removes the the caption from the news default entity view display
- Keeps the caption from the media entity view display in place (note any styling is controlled via figcaption in boostrap)
- Updates the demo content by removing az_caption data for news and adding image 

Still to do (depending on further discussion)

- remove az_caption from the news form display
- remove az_caption from the fields on news content type
- update starter content
- update migrations
- database update (maybe?)

## Related Issue
#1005 

## How Has This Been Tested?
Locally and in Probo

## Types of changes

- [X] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [X] I've added this Pull Request to the AZ Quickstart Github project and set it to "Review in progress".


smayden:
I've deleted the captions field from the news content type and associated entity forms.

I think this is ready for the database update @joeparsons  and migration piece @trackleft

joeparsons:
@smayden #1116 Should be ready to merge into this PR's branch with the DB update(s).

joeparsons:
Question:  Do we also want to remove the `field_az_caption` node field storage config?  I'm thinking probably not because doing so would definitely be riskier than just deleting the field config from `az_news` as its possible that custom content types might be using the `field_az_caption` storage config.  That being said, `az_news` was the only Quickstart content type using this field (I believe) and we could have the database update only delete the field storage config if no other content types are using it.

trackleft:
> Question: Do we also want to remove the `field_az_caption` node field storage config? I'm thinking probably not because doing so would definitely be riskier than just deleting the field config from `az_news` as its possible that custom content types might be using the `field_az_caption` storage config. That being said, `az_news` was the only Quickstart content type using this field (I believe) and we could have the database update only delete the field storage config if no other content types are using it.

Actually I think you are right, since media will have its own storage config

joeparsons:
> Actually I think you are right, since media will have its own storage config

If we do want to remove it, we'll need to update the DB update(s) PR as well.

trackleft:
Let's remove it. I wish the update descriptions could be longer, probably best to add an additional update that says we are deleting that.

smayden:
My assumption was that captions on media used the same field storage config. Should it have?

Happy to delete it on my PR if we want.

trackleft:
@smayden I think there is a separate storage config for each type of media even if the field is named the same.

trackleft:
This might work to list field instances 

```
$field_config = \Drupal\field\Entity\FieldStorageConfig::loadByName('node', 'field_machine_name');
```

joeparsons (approved):

camikazegreen:
This one turned into a monster! Glad it's looking like it should be ready to merge on Friday.

trackleft (approved):

camikazegreen (approved):
Looks good on the front end.

PR #1109, opened 2021-10-27
Closes #880: Add new ViewsReferenceMapping process plugin with views argument lookup handling.

### LINKED ISSUE #880: Migrate Views argument field
As a follow up to https://github.com/az-digital/az_quickstart/pull/851, we need to migrate the content in the argument field on the QS1 views paragraph type.

**Additional context:**
Taxonomy term IDs are not preserved when migrated from QS1 to QS2. The views argument field relies on these term IDs. So we will need a way to look up the new TIDs and apply them to the QS2 argument field in the same format (preserve commas, quotes, etc.)

joeparsons:
## Description
Adds new ViewsReferenceMapping migrate process plugin and deprecates old ParagraphsChunksViewDisplayMapping plugin.

New plugin performs migrate lookups on source viewsfield argument values if present.  Provides built-in mappings for default QS1 views but custom views can be used via `view_mapping` migration configuration key.

## Related Issue
Closes #880 

## How Has This Been Tested?

Pretty extensive local testing with lando.

## Types of changes

**Arizona Quickstart** (install profile, custom modules, custom theme)
- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [x] New feature
- [ ] Other or unknown

**Drupal core**
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

**Drupal contrib projects**
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist:

### DIFF
--- a/modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
+++ b/modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
@@ -1,0 +1,208 @@
+<?php
+
+namespace Drupal\az_migration\Plugin\migrate\process;
+
+use Drupal\Core\Plugin\ContainerFactoryPluginInterface;
+use Drupal\migrate\MigrateExecutableInterface;
+use Drupal\migrate\ProcessPluginBase;
+use Drupal\migrate\Row;
+use Drupal\views\Plugin\views\HandlerBase as ViewsHandler;
+use Symfony\Component\DependencyInjection\ContainerInterface;
+
+/**
+ * Process plugin that maps QS1 viewfield fields QS2 viewsreference fields.
+ *
+ * Used to map view and display names referenced in QS1 field_uaqs_view source
+ * field values into viewsreference field values suitable for use in
+ * field_az_view_reference destination fields in QS2.  Also handles migrating
+ * viewsfield argument values using a the specified taxonomy term migrations.
+ *
+ * Expects a field_uaqs_view source field value.
+ *
+ * Available configuration keys (optional):
+ * - views_mapping : May be used to specify how custom source view and display
+ *   names are mapped to destination views as well as configure which migrations
+ *   are used to perform argument ID migration lookups on.  Default mappings for
+ *   QS1 views and displays are built-in but can be overridden using this
+ *   configuration key.
+ *
+ * Examples:
+ *
+ * Consider a paragraph item migration, where you want to map the views
+ * references in the source field_uaqs_view to QS2 view and display names and
+ * also map additional custom source view and display names to the proper
+ * destination view and display names and also migrate the argument values using
+ * a custom taxonomy term migration.
+ * @code
+ * process:
+ *   field_az_view_reference:
+ *     source: field_uaqs_view
+ *     plugin: az_views_reference_mapping
+ *     view_mapping:
+ *       neuroscience_faculty_directory:
+ *         view: neuroscience_faculty_directory
+ *         display:
+ *           page: page
+ *         argument_migrations:
+ *           - az_person_categories
+ *           - neuroscience_person_categories
+ * @endcode
+ *
+ * @MigrateProcessPlugin(
+ *   id = "az_views_reference_mapping"
+ * )
+ */
+class ViewsReferenceMapping extends ProcessPluginBase implements ContainerFactoryPluginInterface {
+
+  /**
+   * The migrate lookup service.
+   *
+   * @var \Drupal\migrate\MigrateLookupInterface
+   */
+  protected $migrateLookup;
+
+  /**
+   * Provides Quickstart 1 to Quickstart 2 view mapping data.
+   *
+   * @var array
+   */
+  const VIEW_MAPPING = [
+    'uaqs_events' => [
+      'view' => 'az_events',
+      'display' => [
+        'default' => 'page_1',
+        'page' => 'page_1',
+        'list_block' => 'page_1',
+        'card_group_block' => 'az_grid',
+        'block_1' => 'az_sidebar',
+      ],
+      'argument_migrations' => [
+        'az_event_categories',
+      ],
+    ],
+    'uaqs_news' => [
+      'view' => 'az_news',
+      'display' => [
+        'default' => 'az_grid',
+        'three_col_news_block_3' => 'az_grid',
+        'three_col_news_block' => 'az_grid',
+        'sidebar_promoted_news' => 'az_sidebar',
+        'uaqs_teaser_list_page' => 'az_teaser_grid',
+        'uaqs_media_list_page' => 'az_paged_row',
+        'recent_news_marquee' => 'marquee',
+        'recent_news_medium_media_list' => 'az_paged_row',
+      ],
+      'argument_migrations' => [
+        'az_news_tags',
+      ],
+    ],
+    'uaqs_person_directory' => [
+      'view' => 'az_person',
+      'display' => [
+        'default' => 'grid',
+        'page' => 'grid',
+        'page_1' => 'row',
+      ],
+      'argument_migrations' => [
+        'az_person_categories',
+        'az_person_categories_secondary',
+      ],
+    ],
+    'uaqs_content_chunks_views_page_by_category' => [
+      'view' => 'az_page_by_category',
+      'display' => [
+        'default' => 'row',
+        'page' => 'row',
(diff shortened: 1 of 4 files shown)


modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
```diff
@@ -0,0 +1,208 @@
...
+    if (isset($viewMapping[$view])) {
+      $transformedValue['target_id'] = $viewMapping[$view]['view'];
+      $transformedValue['display_id'] = $viewMapping[$view]['display'][$display];
+      $argumentMigrations = $viewMapping[$view]['argument_migrations'];
+    }
+    else {
+      $transformedValue['target_id'] = $view;
+      $transformedValue['display_id'] = $display;
+      $argumentMigrations = NULL;
+    }
```
joeparsons, line 164: Question for team:  If a mapping does not exist for a source view reference, should we skip the row/entity containing the viewsfield, e.g. the Views Paragraph?  If not, the source values are copied directly so we end up with a broken views reference.

Another possible option: create a migrate stub intead?

trackleft: I think we can skip the row and add a migrate message like the skip_on_empty plugin allows for.
https://api.drupal.org/api/drupal/core%21modules%21migrate%21src%21Plugin%21migrate%21process%21SkipOnEmpty.php/class/SkipOnEmpty/9.3.x

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
```diff
@@ -0,0 +1,208 @@
...
+
+    if (!empty($value['vargs']) && !empty($argumentMigrations)) {
+      // Parse view arguments.
+      $arguments = [];
+      $rawArguments = explode('/', $value['vargs']);
+      foreach ($rawArguments as $rawArgument) {
+        /** @var stdClass $parsed */
+        $parsedArgument = ViewsHandler::breakString($rawArgument);
```
trackleft, line 172: Nice

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
trackleft: This would allow rerunning the migration with the --update flag after creating the view on the destination site, and mapping it in their migration.

smitahantzios (commented):
@joeparsons  Looks good to me!

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
trackleft: I do love validation. You might have to create your own custom constraint for this since exiting ones seem to require permissions https://www.berramou.com/blog/drupal-8-create-custom-validation-constraint

https://gitlab.si.umich.edu/root/test/-/blob/8.5.x/core/lib/Drupal/Core/Entity/Plugin/Validation/Constraint/ValidReferenceConstraint.php

trackleft:
Might add az_paragraphs_views as an unenforced module dependency in the migration yml file.

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
joeparsons: I was just thinking we could try loading the view with the EntityTypeManager service for validation.

trackleft: The constraint I linked uses the entityTypeManager

joeparsons:
> Might add az_paragraphs_views as an unenforced module dependency in the migration yml file.

I guess we could although this is designed to be usable with _any_ D7 `viewfield` source field and D9 `viewsreference` destination field on any entity type/bundle (although the only existing example in Qucikstart is in the view paragraph type).

trackleft:
> > Might add az_paragraphs_views as an unenforced module dependency in the migration yml file.
> 
> I guess we could although this is designed to be usable with _any_ D7 `viewfield` source field and D9 `viewsreference` destination field on any entity type/bundle (although the only existing example in Qucikstart is in the view paragraph type).

I was meaning only in the `migrate_plus.migration.az_paragraph_chucks_view.yml` file

trackleft:
It might also be worth adding this as a plugin in the viewsreference project on drupal.org.

trackleft:
I am seeing this on missing views 
image

I am seeing this on an existing view with one argument 

image

Says 9 but should be 44

trackleft:
If I run `lando drush mim az_paragraph_chucks_view --update` after all other migrations run I see the correct value.

trackleft:
To fix this in our migration, we should add migration dependencies.

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
trackleft: FYI an alternative approach would be to add the validation flag to node migrations, which would run field validation for everything, but that would require that we migrate users with a higher level permission unfortunately, since node validation takes permissions into account.

trackleft:
I think the dependencies should look like this

```
dependencies:
  module:
    az_paragraphs_view
  enforced:
    module:
      - az_migration

migration_dependencies:
  optional:
    - az_event_categories
    - az_flexible_page_categories
    - az_news_tags
    - az_person_categories
    - az_person_categories_secondary
```

trackleft:
Other than that, I think this looks fantastic!

trackleft (approved):

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
joeparsons: Looks like since the destination field type extends core's `EntityReferenceItem` field type which already uses the `ValidReferenceConstraint` so I think adding the validation flag to the migration file would effectively force that constraint to be used.

joeparsons:
> It might also be worth adding this as a plugin in the viewsreference project on drupal.org.

I thought about that, too, but the module we use in D7 ([Viewfield](https://www.drupal.org/project/viewfield)) actually has D8/D9 releases now so I'm not sure how common our use case of wanting to migrate D7 viewsfield values into D9 [Views Reference Field](https://www.drupal.org/project/viewsreference) values would be...

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
tadean: In the case where a mapping doesn't exist, I think it would be sufficient to skip the row with a migrate message notifying about the failed mapping value.

tadean (dismissed):

trackleft:
Example of migration config validation https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_paragraphs/src/Plugin/migrate/process/ParagraphsBehaviorSettings.php#L134-L140

modules/custom/az_migration/src/Plugin/migrate/process/ViewsReferenceMapping.php
joeparsons: Thanks all.

joeparsons:
## Remaining tasks (2022-04-08 Friday meeting discussion)

### Before merging
- [ ] Add note about deprecation to AZ Migration module README

### After merging (#1463)
- Add validation for config (similar to what @trackleft added to the [az_paragraphs_behavior_settings plugin](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_paragraphs/src/Plugin/migrate/process/ParagraphsBehaviorSettings.php#L134-L139))
- Log migrate messages when referenced view does not match any mappings

joeparsons:
README updated.  This should be ready for final approval/merging.  Created issue for followups (#1463)

trackleft (approved):

camikazegreen (approved):
Confirmed in Probo migration that the existing Views migrations are still functioning as expected.

tadean (approved):

PR #1140, opened 2021-11-17
Closes #1139: Fixes horizontal scrollbars on full-width paragraphs.

### LINKED ISSUE #1139: Text on media/Text with background horizontal scrollbar
Using the demo content as a basis, the Text on Media and Text with Background sections produce a horizontal scrollbar. The horizontal scrollbar appeared to show on a couple machines in Chrome, Firefox and Edge that I tested.

It looks like .full-width-background sets the left/right margin to `calc(50% - 50vw)` which is probably causing it. The vw unit doesn't account for scrollbars, so certain settings that work with it on the edge can create a horizontal scrollbar when a vertical scrollbar is present (and a vertical scrollbar is always present with the current sticky footer setup).

A few ideas to fix it:
- Setting body to use `overflow-x: hidden;` (Reference: https://stackoverflow.com/a/31391593)
- Adding javascript to calculate scrollbar width and have it create a CSS variable. (Reference: https://stackoverflow.com/a/57748743)
- Go back to the Quickstart 1.0 method of closing the container, putting in the marquee content like with the .paragraphs-item-uaqs-full-width-media-row setting of `width: 100%;`, and re-opening the container after.

image
image
image

joshuasosa:
## Description

This PR uses the first suggestion in #1139 to add CSS that hides horizontally overflowing content on body. Doing so resolves the horizontal scrollbar issue found when using Text with Background and Text on Media paragraph types.

## Related Issue

#1139 

## How Has This Been Tested?

Tested locally with DDEV

## Types of changes

## Checklist:


danahertzberg:
Reviewed in today's PR meeting. Hiding overflow for all of the body could cause inadvertent changes. More exploration into a solution for just full-width paragraphs needed. Maybe look into `width: calc(100vw - (100vw - 100%));`

joshuasosa:
Marked ready but not yet ready with the comment above.

> Maybe look into `width: calc(100vw - (100vw - 100%));`
Tested. Setting width, whether it's on paragraph's container or body, does not change the problem with the margins pushing the edge further and creating the scrollbar.

If overflow-x: hidden isn't ideal, I'll try the second idea of calculating scrollbar width with Javascript.

trackleft:
Thanks, @joshuasosa here are instructions for compiling the JS, https://github.com/az-digital/az_quickstart/blob/main/CONTRIBUTING.md#compiling-javascript-in-local-development

And here is an example
 https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_paragraphs/az_paragraphs_text_media/js/az_paragraphs_az_text_media_youtube.es6.js

joshuasosa:
Updated PR implements solution from https://stackoverflow.com/a/66726233
Demo: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-1140.probo.build/

trackleft (approved):

camikazegreen (dismissed):
Looks good in my testing on pages with full width media and background elements.

joeparsons:
Remaning tasks 
- [ ] Add core jQuery as library dependency (?)
- [ ] Add .es6.js version of Javascript file
- [ ] Generate .js file using babel

trackleft:
> Remaning tasks
> 
> * [ ]  Add core jQuery as library dependency (?)
> * [ ]  Add .es6.js version of Javascript file
> * [ ]  Generate .js file using babel

I can do this.

modules/custom/az_paragraphs/js/az_paragraphs_full_width.js
```diff
@@ -0,0 +1,20 @@
+/**
+* DO NOT EDIT THIS FILE.
+* See the following change record for more information,
+* https://www.drupal.org/node/2815083
+* @preserve
+**/
+
+(function ($, Drupal) {
+  Drupal.behaviors.azParagraphsFullWidth = {
+    attach: function attach() {
```
joeparsons, line 10: Shouldn't this be an anonymous function here?

camikazegreen:
I'm not sure what is causing this, but now on Probo, when I navigate to the homepage most of the menu items are gone...

Overall, the horizontal scrollbars are present during page load, and then go away once it is complete.

trackleft:
> I'm not sure what is causing this, but now on Probo, when I navigate to the homepage most of the menu items are gone...
> 
> Overall, the horizontal scrollbars are present during page load, and then go away once it is complete.

I loaded the pages before the demo content was fully added, I cleared cache, please try again
https://14005782-4f60-4d3c-bae8-b4ee7e4e0199.probo.build/pages/text-background

camikazegreen (dismissed):
I agree that we can clean up the jank on page load with a follow up.

joeparsons (approved):

trackleft (dismissed):

joshuasosa:
Thanks for reviewing! It's not the most ideal solution, but it works to resolve things the way they are. It'd probably be better to use the template suggestion approach like in UAQS, but I'm not sure if there's a reason it wasn't done that way with AZQS.

trackleft:
We did it this way because in QS1 you couldn't have a sidebar over the full width bg wrapper.

modules/custom/az_paragraphs/css/az_paragraphs_full_width.css
```diff
@@ -4,7 +4,7 @@
 .region-content .full-width-background {
     margin-left: calc(50% - 50vw);
     margin-right: calc(50% - 50vw);
-    max-width: calc(100vw - var(--scrollbar-width) / 2);
+    max-width: calc(100vw - var(--scrollbar-width));
```
joshuasosa, line 7: Why was this removed? It's required to get it to the edge appropriately. Now it doesn't sit flesh with the right edge.

trackleft: Hmmm, it wasn't working, but I can add it back.

joshuasosa: Yeah that's the limitation I currently have, unable to test on Mac.

On Windows at least...

Here's with the /2:
edge_with

And currently without the /2:
edge_without

It's not terrible, but doesn't achieve full-width.

trackleft: OK, I've reapplied the fix locally, and it looks like it works.

trackleft: Turns out the problem I was having was sidebar implementation specific.  Your solution works for most use cases.

trackleft:
Please test here for sidebar: https://d77f80bd-3dda-40a6-91cb-04af559e4e2d.probo.build/pages/text-background
And for normal usage:  https://d77f80bd-3dda-40a6-91cb-04af559e4e2d.probo.build

camikazegreen:
On my mac with Firefox, the alignment is perfect without sidebars, and ever so slightly off on a page with sidebars. 

Screen Shot 2021-11-19 at 1 55 00 PM

camikazegreen:
But this solution does get rid of the horizontal scroll bar. If this is holding up the release, I would suggest that we can clean it up after.

trackleft:
There is a tiny white line when there is a left sidebar, which is a fairly rare use case, at any rate, no-sidebars is working great, except for the jank, and one sidebar needs work, though it is better than it was IMHO. RTBC again lol

trackleft:
We do have plans to continuously fine-tune these full-width backgrounds.

Since we are using JS for this, we could probably just set the max-width with js and not worry about finding the scrollbar width in the future.

joshuasosa:
Could try this on Mac to see if there's any difference?
```
.region-content .full-width-background {
    margin-left: calc(50% - (50vw - var(--scrollbar-width)/2))
    margin-right: calc(50% - (50vw - var(--scrollbar-width)/2))
    max-width: 100vw;
    overflow: hidden;
}
```

camikazegreen:
@joshuasosa I pasted that in my inspector and it did not change anything. 

I think I'm ok merging as is with the tiny white line (which now seems to be present in all browsers at least on Mac) and then continuing to refine.

trackleft:
I also see the line, but only on full-width with sidebar, just need approvals before I can merge.

trackleft (approved):

camikazegreen (dismissed):

modules/custom/az_paragraphs/css/az_paragraphs_full_width.css
```diff
@@ -1,10 +1,18 @@
 .nav-link, .nav-pills .nav-link, .nav-tabs .nav-link {
   background-color: white;
 }
+.full-width-background {
+    max-width: calc(100vw - var(--scrollbar-width) / 2);
+}
+
+.layout-sidebar-first .full-width-background,
+.layout-sidebar-second .full-width-background {
+    max-width: calc(100vw - var(--scrollbar-width));
```
camikazegreen, line 10:
```suggestion
    max-width: calc(100vw - var(--scrollbar-width)/2);
```
This fixes it for me.

modules/custom/az_paragraphs/css/az_paragraphs_full_width.css
```diff
@@ -1,10 +1,18 @@
...
+}
+
```
trackleft, line 12:
```suggestion
```

trackleft: What about that ☝️

trackleft (dismissed):

danahertzberg (approved):
Reviewed on the front-end in Firefox on MacOS. No horizontal scrollbars!

jwmoore1 (approved):

trackleft (approved):
Address Jank in https://github.com/az-digital/az_quickstart/issues/1151

joeparsons (approved):

PR #1155, opened 2021-11-23
Closes #1154: Remove problematic role mapping from az_user migration.

### LINKED ISSUE #1154: az_user migration fails when source site has users without assigned roles
## Problem/Motivation
When running the migrations provided by our custom `az_migration` module using a source site containing users with no assigned roles, the `az_user` migration fails.

### Describe the bug
Errors from console when running `drush migrate:import`:
```
[error]  InvalidArgumentException: Unable to remove item at non-existing index. in Drupal\Core\TypedData\Plugin\DataType\ItemList->removeItem() (line 138 of /app/web/core/lib/Drupal/Core/TypedData/Plugin/DataType/ItemList.php). 
[error]  Unable to remove item at non-existing index. (/app/web/core/lib/Drupal/Core/Entity/Sql/SqlContentEntityStorage.php:810)
```

### To Reproduce

Try importing `az_migration` migrations using a source site containing user accounts with no assigned roles.

## Proposed resolution
Currently we have a `static_map` process step in our `az_user` migration which is intended to map all source role assignments to the `authenticated user` role.  This is unnecessary (and problematic) since `authenticated user` is _not_ an assignable role.

If we want all migrated users to only have the authenticated user role, we should simply remove the roles mapping from our `az_user` migration.

### Expected behavior
`az_user` should be able to migrate user accounts regardless of whether or not they have assigned roles.

## Additional context

User entity class method implementing `ItemList::removeItem()` (via `ItemList::offsetUnset()`):
https://git.drupalcode.org/project/drupal/-/blob/9.2.x/core/modules/user/src/Entity/User.php#L92-111

joeparsons:
## Description
The `authenticated user` role is automatically added to all users and not explicitly assignable.  Since we are want all migrated users to only have the `authenticated user` role, we can simply omit our existing (problematic) role mapping which is causing migration failures for source sites having users with no assigned roles.

This also includes some minor cleanup changes and makes the `access` (last access) field value optional (since it isn't populated on the source site if a user has never logged in). 

## Related Issue

#1154 

## How Has This Been Tested?

Locally with lando

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
@@ -19,11 +19,9 @@ process:
   pass: pass
   mail: mail
   created: created
-  access:
-    plugin: skip_on_empty
-    source: access
-    method: process
+  access: access
   login: login
+  # Skip blocked users.
   status:
     plugin: skip_on_value
     equals: true
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
@@ -19,17 +19,18 @@ process:
   pass: pass
   mail: mail
   created: created
-  access: access
+  access:
+    plugin: skip_on_empty
+    source: access
+    method: process
   login: login
-  status: 
-    -
-      plugin: skip_on_value
-      equals: true
-      method: row
-      source: status
-      value: 
-        - 0
-
+  status:
+    plugin: skip_on_value
+    equals: true
+    method: row
+    source: status
+    value:
+      - 0
   timezone: timezone
   langcode:
     plugin: user_langcode
@@ -44,15 +45,6 @@ process:
     source: language
     fallback_to_site_default: true
   init: init
-  roles:
-    plugin: static_map
-    bypass: true
-    source: roles
-    map:
-      3: authenticated
-      2: authenticated
-      4: authenticated
-      5: authenticated
   user_picture:
     -
       plugin: default_value
--- a/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
+++ b/modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
@@ -21,7 +21,7 @@ process:
   created: created
   access: access
   login: login
-  # Skip blocked users.
+  # Skip blocked users.  If you want to maintain a blocked user's history, you should not use this approach.
   status:
     plugin: skip_on_value
     equals: true


trackleft (dismissed):
This doesn't seem to skip_on_empty if they haven't logged in before.  Did you mean to use method: row?
Maybe that is what you meant by making the access field optional.

joeparsons:
> This doesn't seem to skip_on_empty if they haven't logged in before. Did you mean to use method: row? Maybe that is what you meant by making the access field optional.

Yeah, the intention there was to skip that field if it is empty, not the whole row (user).  I guess It might still work without it.

danahertzberg:
So this makes it such that all users migrated from QS1 regardless of role will become "authenticated user" by default?

trackleft:
It currently works as intended, but could probably also work without the skip plugin. Either way is fine with me

joeparsons:
> So this makes it such that all users migrated from QS1 regardless of role will become "authenticated user" by default?

Yes.  That is already what is happening (before this PR) but if the source site has users with no assigned roles, the current migration fails.

modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
```diff
@@ -19,17 +19,18 @@ process:
   pass: pass
   mail: mail
   created: created
-  access: access
+  access:
+    plugin: skip_on_empty
+    source: access
+    method: process
```
trackleft, line 25:
```suggestion
  access: access
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
```diff
@@ -19,17 +19,18 @@ process:
...
-  status: 
-    -
-      plugin: skip_on_value
-      equals: true
-      method: row
-      source: status
-      value: 
-        - 0
-
+  status:
```
joeparsons, line 27:
```suggestion
  # Skip blocked users.
  status:
```

modules/custom/az_migration/migrations/migrate_plus.migration.az_user.yml
```diff
@@ -21,15 +21,14 @@ process:
...
-  status: 
-    -
-      plugin: skip_on_value
-      equals: true
-      method: row
-      source: status
-      value: 
-        - 0
-
+  # Skip blocked users.
```
trackleft, line 24:
```suggestion
  # Skip blocked users.  If you want to maintain a blocked user's history, you should not use this approach.
```

trackleft (approved):

danahertzberg (approved):

PR #1173, opened 2021-12-01
Fixes #1170: Added setting for disabling default user login form.

### LINKED ISSUE #1170: Provide setting (or default config) for disabling default user login form
## Motivation

**Question from @michaelhagedon in `azdigital-general` slack channel 2021-11-30**
>Mike Hagedon  5:13 PM
>Question: CAS is the main authentication mechanism in Quickstart. Is there a reason Quickstart doesn't also disable the user login form entirely?

The CAS module setting we used for this in Quickstart 1 no longer exists in the D9 versions of the CAS module.  We'll probably need to go back to  old way we did this in Quickstart 1 which involved modifying the user/login route.

## Proposed Resolution
Add config setting to AZ CAS (similar to the `disable_password_reset_link` setting) and alter the `user.login` route in a RouteSubscriber and set its access to always be `FALSE` whenever that setting is enabled.

- The D7 CAS module setting we used in Quickstart 1 actually redirected requests to `/user/login` to `/cas`.  Doing this would be more complicated in D9 due to the way Symfony routes/controllers work.
- It may also be possible to achieve the same redirect functionality by utilizing the CAS module's `forced login paths` setting (see comment from CAS module maintainer below).  @geijutsuka also suggested this in slack.

- Administrator : would be able to disable/enable the default user login form

## Additional context
Comment from CAS module maintainer (https://www.drupal.org/project/cas/issues/2839480#comment-11849022):
>...you can achieve this by adding user/login as a forced login path in the CAS settings. You can do the same thing if you want to set some other arbitrary path to force a CAS login.
>
>We made a deliberate decision in the D8 port to NOT make any major modifications to the standard user login form. Instead we just provide an option to include a link on the login form to authenticate using CAS. Site themers can modify the user login form to style it as they'd like,

joeparsons:
## Description
Adds setting to AZ CAS module for disabling default Drupal user login form.  When enabled, this setting makes `/user/login` always return  Access Denied.  
### Notes
- The new setting (which disables the default login page) is currently disabled by default because when it is enabled by default it breaks _a lot_ of tests (any test that uses `\Drupal\Tests\UiHelperTrait::drupalLogin()`).
- My attempts to come up with a solution that redirects requests to `/user/login` to `/cas` (instead of making `user/login` always return access denied) haven't worked so far and generally just cause redirect loops.

## Related Issue
#1170 

## How Has This Been Tested?
Locally with lando and on ProboCI while PR in draft.

## Types of changes

## Checklist:


.probo.yaml
```diff
@@ -53,6 +53,7 @@ steps:
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_editor azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_admin azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_user_admin azuseradmin
+      - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush -y -n config:set az_cas.settings disable_login_form false
```
joeparsons, line 56: If the setting stays disabled by default (as it is now), this step won't actually be needed.

michaelhagedon: Disabled by default makes sense to me.

michaelhagedon (approved):
This works on a fresh local lando build, and the tests pass. The code makes sense to me (though I'm not an expert in this area).

The redirect to `/cas` would be cool -- it would make admin button on Pantheon envs work -- but it sounds difficult / impossible and probably a different ticket.

Thanks! Looks good. If we decide to keep it disabled by default, of course we should delete the ProboCI step as you say.

trackleft (dismissed):

.probo.yaml
joeparsons, line 56:
```suggestion
```

trackleft (approved):

michaelhagedon (approved):

PR #1249, opened 2022-01-24
Closes #1248: Fix undefined index notices generated for existing cards.

### LINKED ISSUE #1248: Clickable Card PHP Notice (undefined index)
### Describe the bug
After updating a Quickstart 2 site to v2.0.15, the following PHP notice is printed in the site's logs for every card on the page you are viewing (that is not clickable). For example, if the page you are viewing has 3 non-clickable cards, you will see 3 notices in the log.

```
Notice: Undefined index: card_clickable in Drupal\az_card\Plugin\Field\FieldFormatter\AZCardDefaultFormatter->viewElements() (line 139 of /code/web/profiles/custom/az_quickstart/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php)
```

The error might be stemming from non-clickable card decks that existed before the clickable card functionality was added? 
Chris thinks we might need to add this to the [formatter](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php#L143):
```if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable'] === 1) {```

### PR DESCRIPTION
**Pull request by joeparsons**
## Description
Should fix undefined index notices being generated on sites with existing cards after updating to 2.0.15.

## Related Issue
Closes #1248 
Issue introduced with #1220

## How Has This Been Tested?
Needs to be tested.

## Types of changes

## Checklist:

### DIFF
--- a/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
+++ b/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
@@ -140,7 +140,7 @@ class AZCardDefaultFormatter extends FormatterBase implements ContainerFactoryPl
           if (!empty($parent_config['az_cards_paragraph_behavior'])) {
             $card_defaults = $parent_config['az_cards_paragraph_behavior'];
             // Is the card clickable?
-            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable'] != FALSE) {
+            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable']) {
               $link_render_array['#attributes']['class'][] = 'stretched-link';
             }
 
--- a/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
+++ b/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
@@ -140,7 +140,7 @@ class AZCardDefaultFormatter extends FormatterBase implements ContainerFactoryPl
           if (!empty($parent_config['az_cards_paragraph_behavior'])) {
             $card_defaults = $parent_config['az_cards_paragraph_behavior'];
             // Is the card clickable?
-            if (isset($card_defaults['card_clickable'] && $card_defaults['card_clickable'] === TRUE) {
+            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable'] === TRUE) {
               $link_render_array['#attributes']['class'][] = 'stretched-link';
             }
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php [hunk 1]
```diff
@@ -140,7 +140,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
           if (!empty($parent_config['az_cards_paragraph_behavior'])) {
             $card_defaults = $parent_config['az_cards_paragraph_behavior'];
             // Is the card clickable?
-            if ($card_defaults['card_clickable']) {
+            if (isset($card_defaults['card_clickable'] && $card_defaults['card_clickable'] === TRUE) {
```
trackleft, line 143:
```suggestion
            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable'] === TRUE) {
```

trackleft: I thought checkboxes always returned a `1` or `0` unless there was a schema

joeparsons: Oops.  Thanks!

akslay:
I created a multi-dev to test the patch against the site where I was noticing the problem. And it looks like the patch fixes the PHP notices but also introduces a new problem. Once you apply the patch, the "clickable" functionality stops working and you can only click on the physical button/link that's inside the card.

trackleft:
Here is what it looks like in the DB 
image

trackleft:
FYI @akslay now that you've confirmed that this PR stops the unidentified index notices, you shouldn't have to test on an existing codebase with the updated code.

akslay:
Tested the updated patch on clickable and non-clickable cards and everything looks good to go this time. (:

akslay (approved):

trackleft (approved):

trackleft (approved):

tadean (approved):

modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
```diff
@@ -140,7 +140,7 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
           if (!empty($parent_config['az_cards_paragraph_behavior'])) {
             $card_defaults = $parent_config['az_cards_paragraph_behavior'];
             // Is the card clickable?
-            if ($card_defaults['card_clickable']) {
+            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable'] != FALSE) {
```
joeparsons, line 143: Changing this back to the way it was for truthy testing to remove the PHPCS warning since @akslay checked the PHP docs:
>When converting to bool, the following values are considered false:
>the empty string, and the string "0"
https://www.php.net/manual/en/types.comparisons.php

```suggestion
            if (isset($card_defaults['card_clickable']) && $card_defaults['card_clickable']) {
```

trackleft (approved):

tadean (approved):

akslay (approved):

PR #1259, opened 2022-01-28
Closes #576 Add dropdown select menu block

### LINKED ISSUE #576: Create Dropdown select menu block type
Match Quickstart 1 "Select Menu" block type that was used in the header, AKA "I am a" menu. This should be available in Content > Blocks.

- Should allow selection of menu
- Should allow customization of leader text
- Should be organizable in Structure > Blocks

trackleft:
## Types of changes

## Checklist:


composer.json
```diff
@@ -46,6 +46,7 @@
         "drupal/ckeditor_bs_grid": "2.0.7",
         "drupal/coffee": "1.2",
         "drupal/config_distro": "1.0-alpha4",
+        "drupal/config_inspector": "^2.0",
```
joeparsons, line 49: Is this something you meant to add or just something to use for local development?

modules/custom/az_select_menu/az_select_menu.module
```diff
@@ -0,0 +1,67 @@
...
+ * Provides configurable blocks of menu links.
+ */
+
+/**
+ * Implements hook_theme().
+ */
+function az_select_menu_theme() {
+  return [
+    'az_select_menu' => [
+      'base hook' => 'az_select_menu',
```
joeparsons, line 14: Shouldn't this be `menu_block` or some other theme hook?

composer.json
trackleft, line 49:
```suggestion
```

modules/custom/az_select_menu/az_select_menu.module
trackleft: menu_block uses `menu`
but I couldn't make my module find my theme template without it.

trackleft:
I removed most of the jQuery from the javascript, though .popover requires jQuery so, we are still dependent here.

trackleft:
To Test,  enable the module, create a menu, or use an existing menu that isn't already using some theme template.
Place block, and find your menu
Set some options in the Quickstart Select Menu accordion.

modules/custom/az_select_menu/az_select_menu.module
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * Implements hook_theme_suggestions_HOOK() for "menu".
+ */
+function az_select_menu_theme_suggestions_menu(array $variables) {
+  $suggestions = [];
+
+  // The AzSelectMenu plugin's build() method populates this variable.
+  if (!empty($variables['menu_block_configuration'])) {
+    if (!empty($variables['menu_block_configuration']['provider'])) {
+      if ($variables['menu_block_configuration']['provider'] === 'az_select_menu') {
```
trackleft, line 39: Only adds these suggestions to blocks that use this provider.

modules/custom/az_select_menu/src/Plugin/Block/AzSelectMenu.php
```diff
@@ -0,0 +1,179 @@
...
+      '#default_value' => $config['az_select_menu']['empty_option'],
+    ];
+
+    $form['az_select_menu']['empty_option_label'] = [
+      '#type' => 'textfield',
+      '#title' => t('Empty option label.'),
+      '#default_value' => $config['az_select_menu']['empty_option_label'],
+      '#states' => [
+        'visible' => [
+          ':input[name="settings][az_select_menu][empty_option"]' => ['checked' => TRUE],
```
joeparsons, line 61: Did you already happen to try it like this?

```suggestion
          ':input[name="settings[az_select_menu][empty_option]"]' => ['checked' => TRUE],
```

trackleft:
@joeparsons your latest commit suggestion did resolve the issue with the form controls.

trackleft:
I think we should look at the menu block form and decide what to remove from the quickstart select menu options, or at least disable.
image
image

danahertzberg (dismissed):

smayden (approved):
tested front end, looking good

modules/custom/az_select_menu/src/Plugin/Block/AzSelectMenu.php
```diff
@@ -0,0 +1,179 @@
...
+
+    $form = parent::blockForm($form, $form_state);
+    $form['menu_levels']['depth']['#attributes'] = ['disabled' => 'disabled'];
+    $form['menu_levels']['depth']['#default_value'] = 1;
+    $form['menu_levels']['depth']['#description'] = 'This maximum number includes the initial level. This has to be set to 1 for Quickstart Select Menus.';
+    $form['menu_levels']['expand_all_items']['#default_value'] = 0;
+    $form['menu_levels']['expand_all_items']['#attributes'] = ['disabled' => 'disabled'];
+    $form['menu_levels']['expand_all_items']['#states']['visible'] = [
+      ':select[name="settings[depth]"]' => ['!value' => 1],
+    ];
```
joeparsons, line 50: Could we just set `#access` to `FALSE` for both of these?
```suggestion
    $form['menu_levels']['depth']['#access'] = FALSE;
    $form['menu_levels']['expand_all_items']['#access'] = FALSE;
```

danahertzberg (approved):

joeparsons:
Follow-up task:

Change JS to support more than one instance or explicitly indicate that only one instance is supported.

camikazegreen (approved):

smayden (approved):

joeparsons (approved):

PR #1266, opened 2022-01-31
2.1.x branch-specific changes for CI configs.

joeparsons:
## Description
2.1.x branch specific changes needed for CI and local development tool configs. This should only be merged into the `2.1.x` branch.

### DIFF
--- a/.ddev/commands/web/install
+++ b/.ddev/commands/web/install
@@ -5,7 +5,7 @@
 ## Example: "ddev install"
 
 rm -rf /var/www/html/*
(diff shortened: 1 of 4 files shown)


joeparsons:
This needs to be merged before we can create the `2.1.0-alpha1` release.

trackleft:
Do we want to add a version to the .info.yml file? 

https://github.com/az-digital/az_quickstart/blob/main/az_quickstart.info.yml

trackleft (approved):

kevcooper (approved):

joeparsons:
> Do we want to add a version to the .info.yml file?
> 
> https://github.com/az-digital/az_quickstart/blob/main/az_quickstart.info.yml

Possibly but I don't think that needs to be a blocker for our `2.1.0-alpha1` release and I think we need a separate ticket / PR for that and possibly even a spike for how to best automate keeping a version reference in `.info.yml` up to date.

PR #1353, opened 2022-02-28
Closes #1208: Added classes to page template depending on contents of off-canvas-nav

### LINKED ISSUE #1208: Persistent main nav borders
When disabling or removing the main nav block from the navigation region (top nav) the borders remain when they shouldn't.

See above image for gray lines

cpoff:
## Description

When the navbar is removed the 1px grey border remained because the border CSS was assigned to the navbar-offcanvas selector, which was still in the DOM. I added logic to the twig page template to assign classes based on whether nav was populated or empty.

## Related Issue

[https://github.com/az-digital/az_quickstart/issues/1208](https://github.com/az-digital/az_quickstart/issues/1208)

## How Has This Been Tested?

I removed the navigation using several methods, most involving the Block Layout tool.

## Types of changes

## Checklist:


themes/custom/az_barrio/templates/layout/page.html.twig
```diff
@@ -130,9 +130,15 @@
...
           <div class="container">
             <div class="row">
-              {% if az_barrio_navbar_offcanvas %}
-              <div class="col-lg">
-                <nav class="navbar-offcanvas" id="navbarOffcanvasDemo">
+                {%
+                  set navbar_classes = [
+                    az_barrio_navbar_offcanvas ? 'navbar-offcanvas',
+                    page.navigation ? 'has-navigation-region' : 'no-navigation-region' ,
+                    page.navigation_offcanvas ? 'has-off-canvas-region' : 'no-off-canvas-region',
```
joeparsons, line 137: I don't think Twig supports trailing commas.  Also a whitespace nitpick.
```suggestion
                    page.navigation ? 'has-navigation-region' : 'no-navigation-region',
                    page.navigation_offcanvas ? 'has-off-canvas-region' : 'no-off-canvas-region'
```

themes/custom/az_barrio/templates/layout/page.html.twig
```diff
@@ -130,9 +130,15 @@
 
           <div class="container">
             <div class="row">
-              {% if az_barrio_navbar_offcanvas %}
```
joeparsons, line 133: I think we still need to this `if` statement to be here.  If we don't, the `endif` on line 157 should also be removed.

themes/custom/az_barrio/templates/layout/page.html.twig
joeparsons: Actually, the trailing comma might be fine.  I think the removal of the if statement above without removing the `endif` is what is causing the syntax error.

trackleft: @joeparsons do we want to move setting of the classes to the top of the template?

trackleft:
I think there is a file missing with the css to remove the border.

So to me the options are, add custom CSS in this PR, or add a `border-0` class instead.

https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/borders/#subtractive

themes/custom/az_barrio/templates/layout/page.html.twig
joeparsons: > @joeparsons do we want to move setting of the classes to the top of the template?

Sounds reasonable to me.

trackleft:
Should we consider this PR a bugfix, and address this:

> ☝️ also looks like there is an existing variable that is not being used `navbar_offcanvas_attributes` 🤦

In a follow-up?

joeparsons:
> Should we consider this PR a bugfix, and address this:
> 
> > ☝️ also looks like there is an existing variable that is not being used `navbar_offcanvas_attributes` 🤦
> 
> In a follow-up? Since there are Google Tag Manager implications?

I think that sounds reasonable.  What are the GTM implications?

trackleft:
> I think that sounds reasonable. What are the GTM implications?

Changing the ID could cause some tracking code to cease working.

themes/custom/az_barrio/templates/layout/page.html.twig
cpoff: Ok. That snippet now moved up below the comments and before the opening page-wrapper.

themes/custom/az_barrio/css/az-barrio-off-canvas-nav.css
```diff
@@ -2,3 +2,7 @@
 header#header_arizona {
   min-height: 50px;
 }
+
+.navbar-offcanvas.no-navigation-region {
+  border: 0;
+}
```
trackleft, line 8:
```suggestion
}

```

cpoff: Done

trackleft (approved):

trackleft (approved):

joeparsons (approved):
Changes look good to me now.  We should be sure to create a ticket for the follow-up tasks before merging.

danahertzberg:
Works with block being disabled, but not if menu links are all disabled (didn't test if menu links are deleted). This is easy enough to communicate and should not block merging.

danahertzberg (approved):

PR #1355, opened 2022-02-28
Fixes #1354 Lando database doesn't play nicely with M1 Mac.

### LINKED ISSUE #1354: Lando database doesn't play nicely with M1 Mac.
## Problem/Motivation
Often while making pull requests and developing for quickstart via lando, the database goes away and in the log, there is a segmentation fault error.

### Describe the bug
The lando configuration for database is currently configured in an incompatible way.
`qemu: uncaught target signal 11 (Segmentation fault) - core dumped`

### To Reproduce
Use M1 Mac
Install quickstart with lando
`lando phpunit` will usually crash the database immediately.

## Proposed resolution
Update the lando config file to use a more recent version of mariadb

### Expected behavior
`lando phpunit` should not crash database

##  Screenshots
If applicable, add screenshots to help explain your problem. (Need a tool to share screenshots or screencasts as URLs? 
image
image

 - OS: MacOS
 - Browser N/A

### Smartphone (please complete the following information):
 - Device: M1 Mac

trackleft:
Update the .lando.yml file to use latest version of mariadb.

## Description
See https://github.com/lando/lando/issues/2688#issuecomment-998570544

## Related Issue
Closes #1354 

## How Has This Been Tested?
Locally

## Types of changes

## Checklist:


joeparsons:
It's worth noting that this essentially replaces the default mariadb Docker image used by lando ([`bitnami/mariadb`](https://hub.docker.com/r/bitnami/mariadb/)) with the official Docker [`mariadb`](https://hub.docker.com/_/mariadb) image that is arm64 (e.g. Apple M1) compatible so lando's buit-in SQL tooling (`db-import` and `db-export`) doesn't work.

trackleft:
Notes:  In the current state of this PR the lando db-export and lando db-import commands do not function.
I did try adding config following this example: https://docs.lando.dev/guides/db-export.html#adding-the-db-export-command
But unfortunately, the script can not be passed the name of the database and therefore can not find it.

Additionally, I attempted to set `MYSQL_DATABASE` and `DATABASE` env vars to the lando file, but was unsuccessful.

joeparsons:
Since this breaks lando's `db-import` `db-export` functionality, I'm thinking we might want to add this as a commented out section that can be optionally enabled by folks who use M1 macs instead of changing the defaults.

.lando.yml
```diff
@@ -4,9 +4,21 @@ config:
   php: '7.4'
   via: apache:2.4
   webroot: web
-  database: mariadb:10.4
   xdebug: false
```
joeparsons, line 7:
```suggestion
  database: mariadb:10.4
  xdebug: false
```
In my testing I  found that this `database` line doesn't actually need to be deleted.

.lando.yml
```diff
@@ -4,9 +4,21 @@ config:
...
 services:
+  database:
```
joeparsons, line 9:
```suggestion
  # Replaces default lando MariaDB image with arm64 (Apple M1) compatible one.
  # TODO: remove database service overrides once lando is updated.
  database:
```

.lando.yml
```diff
@@ -4,9 +4,21 @@ config:
...
+    type: compose
+    services:
+      image: mariadb:latest
+      command: docker-entrypoint.sh mariadbd
+      restart: always
+      ports:
+        - '3306'
+      environment:
+        MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: 'true'
+        MARIADB_DATABASE: drupal9
```
joeparsons, line 19: I found that I was able to use `db-export` and `db-import` after adding this and running `lando rebuild`
```suggestion
        MARIADB_DATABASE: drupal9
        MYSQL_DATABASE: drupal9
```

joeparsons (commented):
Tested this locally.  Found that it worked well on an M1 Mac, especially after the suggested changes.

joeparsons (approved):
I tested this on an M1 mac and it works well for me.

jwmoore1 (dismissed):
Built locally on M1 Mac and worked as expected.

.lando.yml
```diff
@@ -7,6 +7,22 @@ config:
   database: mariadb:10.4
   xdebug: false
 services:
+  # Replaces default lando MariaDB image with arm64 (Apple M1) compatible one.
+  # TODO: remove database service overrides once lando is updated.
+  database:
+    type: compose
+    services:
+      image: mariadb:latest
```
joeparsons, line 15: I found that using `10.4` here also worked.  Maybe that would be a "safer" option?
```suggestion
      image: mariadb:10.4
```

joeparsons:
FWIW, lando's `db-export` and `db-import` tooling works with this PR for me now but I do see a lot of these warnings getting generated each time I use either one:
```
WARNING: Forcing protocol to  TCP  due to option specification. Please explicitly state intended protocol.
```

joeparsons:
I also tested this on an Intel Mac and everything seems to work as expected (aside from the warnings mentioned above).

.lando.yml
trackleft: I will test this

.lando.yml
```diff
@@ -7,6 +7,22 @@ config:
...
+      command: docker-entrypoint.sh mariadbd
+      restart: always
+      ports:
+        - '3306'
+      environment:
+        MARIADB_ALLOW_EMPTY_ROOT_PASSWORD: 'true'
+        MARIADB_DATABASE: drupal9
+        MYSQL_DATABASE: drupal9
+        MARIADB_USER: drupal9
+        MARIADB_PASSWORD: drupal9
```
joeparsons, line 25: I think we could get rid of the `db-export` and `db-import` warnings by adding these.
```suggestion
        MARIADB_PASSWORD: drupal9
        LANDO_EXTRA_DB_EXPORT_ARGS: '--protocol=TCP'
        LANDO_EXTRA_DB_IMPORT_ARGS: '--protocol=TCP'
```

trackleft: Where are you finding these answers?

joeparsons: https://github.com/lando/cli/blob/main/plugins/lando-services/scripts/sql-export.sh
https://github.com/lando/cli/blob/main/plugins/lando-services/scripts/sql-import.sh

.lando.yml
joeparsons: FWIW using `mariadb:10.4` here is _NOT_ the same thing as the `database: mariadb:10.4` line (line 7).  Lando uses [`bitnami/mariadb`](https://hub.docker.com/r/bitnami/mariadb/) docker images by default which don't currently support arm64, which is why it crash on M1 Macs.  `mariadb:10.4` here is actually referencing the [default Docker mariadb image](https://hub.docker.com/_/mariadb), which does support arm64.

.lando.yml
joeparsons: > I think we could get rid of the `db-export` and `db-import` warnings by adding these.

Confirmed that these lines remove the warnings.

joeparsons: Actually, I found these weren't needed when using `mariadb:10.4` on line 15.

.lando.yml
joeparsons: This change also removes the warnings I was seeing.

.lando.yml
```diff
@@ -7,6 +7,22 @@ config:
   database: mariadb:10.4
```
trackleft, line 7:
```suggestion
```

joeparsons (approved):
I think this is ready to go now.  Thanks.

jwmoore1 (approved):

Issue #1403, opened 2022-03-16
Add natural image style 


danahertzberg:
## Problem/Motivation
I want my embedded media to respect alignment selections when using natural size.

### Describe the bug
No image style exists for the natural image size (img-fluid is not being applied to image)

### To Reproduce
1. Embed small media in natural size
2. Select center alignment
3. Save page
4. Image is not centered

## Proposed resolution
Add image style for natural size

### Expected behavior
Embedded media should appear at its original natural size and aligned with selected setting.

cpoff:
The left align seems to only happen because of the display attributes (`display:block`). I think I have this fix in two parts:

1. Add a wrapper div around the figure to set a new class. Media Type > Manage Display Natural Size. I added a field group wrapper called Align Natural Image and assigned the class: `align-natural-image`.
2. Add class CSS to style.css:

```
.align-natural-image {
	display: flex;
}
```
`display: grid` also appears to work just the same as flex, so we can swap if needed.

cpoff:
![Uploading Screen Shot 2022-04-07 at 2.24.42 PM.png…]()

cpoff:
![Screen Shot 2022-04-07 at 2 24 42 PM](https://user-images.githubusercontent.com/730076/162828575-b5bb5e82-3ece-4208-8be4-bd5289736cbb.png)

cpoff:
![Screen Shot 2022-04-11 at 1 41 47 PM](https://user-images.githubusercontent.com/730076/162828671-32b25403-cac3-42d3-85bb-a32543fb64a7.png)

bberndt-uaz:
Would adding the "img-fluid" class to images using the "Natural Size" display mode be within the scope of this issue as well? As @danahertzberg mentioned in the issue description, this class is not added with the "Natural Size" display mode but it is added with the other four display modes. I think the class would be helpful for preventing large images from breaking out of their container.

cpoff:
The sm, med, lg, and square image styles all place a max-width on the figure. Then the actual HxW is applied to the img explicitly. With the Natural Size, I can't really put a max-width on it, and the img-fluid only works on the img and not the figure. I assume the img-fluid class is added by the theme?

trackleft:
This is where the `img-fluid` class is added by the theme
https://github.com/az-digital/az_quickstart/blob/caec8c5750e126107b4f14d981af0302ff57e722/themes/custom/az_barrio/az_barrio.theme#L154-L158

PR #1469, opened 2022-04-11
Closes 1403: Added style for Natural Image behavior

cpoff:
This issue in #1403 was when you embed an image, set the size to Natural, and align center ... the block aligned left. So I added a wrapper class to override that display setting and make it `flex` instead of `block`. Changes should apply only to the Natural Image display. Did a new class because I didn't want to bother existing blocks using that behavior.

It required changes to two files:

1. Structure > Media Type > Image (Edit) > Manage display (Filename: modules/custom/az_media/config/install/core.entity_view_display.media.az_image.az_natural_size.yml) to add new container/wrapper setting the class to align-image-natural.
2. Added those class settings to style.css:

```
.align-natural-image {
  display:flex;
}
```

### DIFF
--- a/themes/custom/az_barrio/css/style.css
+++ b/themes/custom/az_barrio/css/style.css
@@ -6,6 +6,11 @@
  * of these @include files will be combined into a single file.
  */
 
+/* Class for Natural Image block alignment issue */
+.align-natural-image {
+  display:flex;
+}
+
 /* For structuring header utilities*/
 header#header_arizona {
   min-height: 10px;
--- a/themes/custom/az_barrio/css/style.css
+++ b/themes/custom/az_barrio/css/style.css
@@ -6,7 +6,6 @@
  * of these @include files will be combined into a single file.
  */
 
-
 /* For structuring header utilities*/
 header#header_arizona {
   min-height: 10px;
--- a/themes/custom/az_barrio/css/style.css
+++ b/themes/custom/az_barrio/css/style.css
@@ -6,18 +6,10 @@
  * of these @include files will be combined into a single file.
  */
 
-/* Class for Natural Image block alignment issue */
-.align-natural-image {
-  display:flex;
-}
 
 /* For structuring header utilities*/
 header#header_arizona {
   min-height: 10px;
-}
-
-.align-natural-image {
-  display:flex;
 }
 
 /* Sticky footer */
--- a/modules/custom/az_media/config/install/core.entity_view_display.media.az_image.az_natural_size.yml
+++ b/modules/custom/az_media/config/install/core.entity_view_display.media.az_image.az_natural_size.yml
@@ -6,7 +6,7 @@ dependencies:
     - field.field.media.az_image.field_az_caption
     - field.field.media.az_image.field_az_credit
     - field.field.media.az_image.field_media_az_image
-    - image.style.natural
+    - image.style.az_natural
     - media.type.az_image
   module:
     - field_group
--- a/modules/custom/az_media/config/install/core.entity_view_display.media.az_image.az_natural_size.yml
+++ b/modules/custom/az_media/config/install/core.entity_view_display.media.az_image.az_natural_size.yml
@@ -6,6 +6,7 @@ dependencies:
     - field.field.media.az_image.field_az_caption
     - field.field.media.az_image.field_az_credit
     - field.field.media.az_image.field_media_az_image
+    - image.style.natural
     - media.type.az_image
   module:
     - field_group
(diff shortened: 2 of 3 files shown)


trackleft:
You need to add this config 
`core.entity_view_display.media.az_image.az_natural_size`

```
160
161In UnmetDependenciesException.php line 100:
162                                                                               
163  [Drupal\Core\Config\UnmetDependenciesException]                              
164  Configuration objects provided by <em class="placeholder">az_media</em> hav  
165  e unmet dependencies: <em class="placeholder">core.entity_view_display.medi  
166  a.az_image.az_natural_size (image.style.natural)</em>                        
```

cpoff:
I added image.style.az_natural.yml, but it has some SecKit formatting that I can't figure out. Here's an example of the Small image config style that I cloned:

```
langcode: en
status: true
name: az_small
label: 'Small (360px)'
effects:
  a5aee9bc-6dd3-4b48-b04e-027cc8017630:
    uuid: a5aee9bc-6dd3-4b48-b04e-027cc8017630
    id: image_scale
    weight: 1
    data:
      width: 360
      height: null
      upscale: true
```

trackleft:
Alternatively we could add another class to set `width: fit-content `if we didn't want to alter the align-center class

akslay:
> Alternatively we could add another class to set `width: fit-content `if we didn't want to alter the align-center class

I like the idea of adding a utility class of something like `.w-fit-content` (to match other width utility classes like `.w-100`). And then have that class set the `width: fit-content` style.

As a side note, the width utility classes are coming from Bootstrap. So it might be a good idea to put it in there rather than in Quickstart, but that would end up delaying this PR. So maybe we can put it in QA CSS for now and have a follow-up to get it added to AZ Bootstrap?

trackleft:
> As a side note, the width utility classes are coming from Bootstrap. So it might be a good idea to put it in there rather than in Quickstart, but that would end up delaying this PR. So maybe we can put it in QS CSS for now and have a follow-up to get it added to AZ Bootstrap?

I think we can do what you described and make a follow-up ticket in arizona-bootstrap.

cpoff:
Just for the record, here's my existing process:

Add new wrapper in `Media Types > Image > Display Type` to add "natural-image-align" class around the entire figure:

```
.natural-image-align {
	display:grid;
}
```

.... added to the custom theme style.css.

This was done with a custom class as to not clobber the `align-center` class, which specifies "display:block" and is used elsewhere around the site.

Plus I added the new image style, which should draw the default `img-fluid` class applied to all images with a defined image style.

trackleft:
FYI another alternative would be to use the style name prefix that gets added to the img element.

```
.az-natural-size.img-fluid {
  width: fit-content'
}
```

trackleft:
FYI https://caniuse.com/mdn-css_properties_width_fit-content

trackleft:
FYI there is no need to add a new class, `d-flex` would change the display

https://digital.arizona.edu/arizona-bootstrap/docs/2.0/utilities/display/

cpoff:
Indeed it does. Thanks Chris!

Would it be acceptable to re-add a line back in the Media Type > Natural Display? And rather than creating a new class/adding definition to style.css just include that `d-flex` class in a standalone div?

cpoff:
Screen Shot 2022-04-14 at 4 08 39 PM

cpoff:
`d-flex` utility class installed. Ready for checkmarks.

themes/custom/az_barrio/css/style.css
```diff
@@ -6,11 +6,20 @@
...
+
 /* For structuring header utilities*/
 header#header_arizona {
   min-height: 10px;
 }
 
+.align-natural-image {
+  display:flex;
+}
+
```
joeparsons, line 22: Should these styles get removed now that you're using the `d-flex` class?

```suggestion
/* For structuring header utilities*/
header#header_arizona {
  min-height: 10px;
}
```

smayden (approved):
tested as content admin. natural style can now be centered. Other image styles are still working.

cpoff:
style.css updated to remove old styles. Ready to merge.

themes/custom/az_barrio/css/style.css
```diff
@@ -6,6 +6,7 @@
  * of these @include files will be combined into a single file.
  */
 
+
```
danahertzberg, line 9: Do we care about the extra added space here? 😬

danahertzberg, line 9:
```suggestion
```

danahertzberg:
```suggestion
```

joeparsons: would be better to remove it, if possible.
```suggestion
```

danahertzberg (approved):

smayden (approved):

PR #1471, opened 2022-04-13
Fixes #1460 Allow for Delayed Configuration Overrides.

### LINKED ISSUE #1460: Config distro import fails with az_security module installed
## Problem/Motivation
While working on a workaround for the Drupal 9.3.0 permissions config changes, @tadean noticed that the changes we merged in #915 actually cause errors to be encountered when running config distro import operations.

#915 added the seckit module to Quickstart and included some "override default settings" (quickstart config) for the seckit module in az_security but did _not_ install seckit by default.

### Describe the bug

Error message:
```
The configuration cannot be imported because it failed validation for the following reasons:
Configuration seckit.settings depends on the Security Kit module that will not be installed after import.
```

### To Reproduce

Run config distro import on a site with az_security installed.

## Proposed resolution
From @tadean:
>It looks like we should add a step to our quickstart provider that makes it only work if the module it's overriding is actually enabled. Since it seems to be the cause of the issue was it cheerfully loading the sec kit settings into active config even though there was no seckit

### Expected behavior
It should be possible to run config distro import without errors when the `az_security` module is installed.

tadean:
This PR adds functionality for quickstart configuration overrides to not install themselves until the module that uses them is ready; this covers the case of `az_security` where it contains overrides for `seckit.settings` but `seckit` is not installed by default.

This means that an attempt to install `seckit.settings` would create problems because it installs configuration for a module that doesn't exist. When a user later attempts to install  `seckit` it would fail, because it would find that its configuration already exists.

This PR attempts to extend the Quickstart configuration provider to wait to install configuration overrides until the dependencies it is based on are present. This creates the notion of a "passive" override that has not be installed yet because a module it depends upon doesn't exist yet.

A database update is included to remove `seckit.settings` on sites where `seckit` is not enabled.

## Related issues
#1460 

## How to test
- Add `config/quickstart` overrides for a module that does not exist yet
- Install the module containing the overrides
- Verify overrides not installed
- Install the module the overrides depend upon
- Verify the overrides are installed

- Patch release changes
   - [x] Bug fix
- Minor release changes
   - [x] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/src/AZConfigOverride.php
+++ b/modules/custom/az_core/src/AZConfigOverride.php
@@ -87,11 +87,14 @@ class AZConfigOverride {
     $module_keys = array_flip($modules);
     $extensions = array_intersect_key($module_list, $module_keys);
 
+    // Previously enabled extensions.
+    $old_extensions = array_diff_key($module_list, $module_keys);
+
     // Ask the override provider for direct overrides available.
     foreach ($providers as $provider) {
       // Only query config for the Quickstart provider.
       if ($provider instanceof QuickstartConfigProvider) {
-        $overrides = $provider->getOverrideConfig($extensions);
+        $overrides = $provider->getOverrideConfig($extensions, $old_extensions);
 
         // Edit active configuration for each explicit override.
         foreach ($overrides as $name => $data) {
--- a/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
+++ b/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
@@ -35,22 +35,64 @@ class QuickstartConfigProvider extends ConfigProviderBase {
    *
    * @param \Drupal\Core\Extension\Extension[] $extensions
    *   An associative array of Extension objects, keyed by extension name.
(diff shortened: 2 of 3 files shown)


joeparsons (approved):
I feel comfortable with this new approach.  I will try to do some actual testing soon.

One question that affects whether this change gets backported / included in a patch release or not: if someone has custom code that invokes `QuickstartConfigProvider::getOverrideConfig()` for some reason (probably very unlikely, will this method still work the same way it did without the `$old_extensions` parameter?

joeparsons:
I already added my checkmark but I have now also performed the following steps to test this PR branch:
- Added exported config overrides for Antibot module as Quickstart config to az_security module (see #1485)
- Installed Quickstart
- Installed Antibot module
- Verified that Antibot config overrides were imported correctly.

Extra approved.

jwmoore1 (approved):

cpoff (approved):

jwmoore1 (approved):

Issue #1486, opened 2022-04-15
Create a Migration Source Plugin for Filtering out Out of Date Content


rmarsh1ua:
## Motivation
While performing Drupal 7 migrations, sometimes sites have a plethora of old content that hasn't been updated in years and is better left behind. Rather than leaving that responsibility to a human, why not add a migration plugin that can filter old content by date.

## Proposed Resolution

Add a source plugin that filters out old content.

### Describe the solution you'd like

The plugin should accept a date and node type as input. The date can be a timestamp or a string formatted date.

### Describe alternatives you've considered
The alternative is that someone manually cleans content before or after a migration takes place. Very tedious.

### Roles and Permissions considerations
None of these roles would be impacted as this would not affect the public facing site or admin interface in any way; it would purely be a new tool available for developers working on a Drupal 7 migration.

PR #1488, opened 2022-04-15
Adds a source migration plugin that filters out old content

rmarsh1ua:
## Description

New source plugin that filters old content that's being migrated

## Related issues
closes https://github.com/az-digital/az_quickstart/issues/1486

- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [X] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
   - [] Major or minor level update
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist

- [X] I have updated the documentation accordingly.


modules/custom/az_migration/src/Plugin/migrate/source/FilterOldContent.php
```diff
@@ -0,0 +1,57 @@
...
+ * a decade. How to easily cut that out of the migration:
+ * @code
+ *  source:
+ *    plugin: filter_old_content
+ *    node_type: uagc_page
+ *    filter_date: "2016-01-01"
+ * @endcode
+ *
+ * @MigrateSource(
+ *   id = "filter_old_content"
```
joeparsons, line 30: Nitpick: we should prefix the plugin ID with `az_`

joeparsons, line 30: Nitpick: we should prefix the plugin ID with `az_`

```suggestion
 *   id = "az_filter_old_content"
```

danahertzberg:
To test: Switch plugin of events migration from D7 node to this plugin

joeparsons:
> What do you think @joeparsons @tadean ?

I like the idea of adding this as optional functionality that can be enabled via config to our existing `az_node` source plugin.

rmarsh1ua:
> > What do you think @joeparsons @tadean ?
> 
> I like the idea of adding this as optional functionality that can be enabled via config to our existing `az_node` source plugin. Would you be OK with that approach @rmarsh1ua?

Yea, I think that makes sense. Originally this was meant to sort of extend az_node in a way I thought was kind of boutique but I agree its probably a really common scenario and it makes to incorporate it directly into the more general plugin, az_node.

modules/custom/az_migration/src/Plugin/migrate/source/AZNode.php
```diff
@@ -44,4 +60,28 @@ public function prepareRow(Row $row) {
...
+      else {
+        $date_cutoff = (new \DateTime($this->configuration['filter_date']))->getTimestamp();
+      }
+
+      $query->condition('n.changed', $date_cutoff, '>');
+    }
+
+    return $query;
+  }
+
```
trackleft, line 86: /**
   * {@inheritdoc}
   */
  public function is_timestamp($timestamp)
{
   return ((string) (int) $timestamp === $timestamp)
   && ($timestamp <= PHP_INT_MAX)
   && ($timestamp >= ~PHP_INT_MAX)
   && (!strtotime($timestamp));
}

modules/custom/az_migration/src/Plugin/migrate/source/AZNode.php
```diff
@@ -44,4 +60,28 @@ public function prepareRow(Row $row) {
...
+   */
+  public function query() {
+    $query = parent::query();
+    /*
+     * Filter date from migration script can be a
+     * string date or a UNIX timestamp so get
+     * it into correct format
+     */
+    if (array_key_exists("filter_date", $this->configuration)) {
+      if (is_int($this->configuration['filter_date'])) {
```
trackleft, line 74:
```suggestion
      if ($this->is_timestamp($this->configuration['filter_date'])) {
```

modules/custom/az_migration/src/Plugin/migrate/source/AZNode.php
trackleft: What do you think about this approach @rmarsh1ua

rmarsh1ua: It blows my mind that this isn't something built into PHP as a general validation function. Looks good, but should it be a private function?

trackleft: Yeah, I went down a rabbit hole looking for a built-in too, ha. Probably should be private

modules/custom/az_migration/src/Plugin/migrate/source/AZNode.php
```diff
@@ -44,4 +60,55 @@ public function prepareRow(Row $row) {
...
+     */
+    if (array_key_exists("filter_date", $this->configuration)) {
+      if ($this->isTimestamp($this->configuration['filter_date'])) {
+        $date_cutoff = $this->configuration['filter_date'];
+      }
+      else {
+        $date_cutoff = (new \DateTime($this->configuration['filter_date']))->getTimestamp();
+      }
+
+      $query->condition('n.changed', $date_cutoff, '>');
```
trackleft, line 81: Possibly should be `>=`

trackleft (approved):

Issue #1340, opened 2022-02-23
Resolve 9.3.x permissions issue(s)


danahertzberg:
~Blocks PR #1181~

Need to determine a way to address permissions in a Drupal core 9.3.0-compatible way

https://www.drupal.org/node/3193348

joeparsons:
## Remaining issues after merging #1181

Various permissions are not being preserved by QuickstartConfigProvider:
- Entity Bundle permissions (e.g. `edit / own X entity`) generated by node, taxonomy, etc modules for entity bundles included in other modules (e.g. webform)
- `Assign X role` permissions for the Role Delegation module
- `use text format full_html` permission

PR #1181, opened 2021-12-07
#1340, Closes #1180: Prepare Quickstart for Drupal core 9.3.x.

joeparsons:
## Description
- Re-exported ALL configuration included with Quickstart to apply sorting changes from Drupal core
  - See https://www.drupal.org/node/3230199
- Updated various core patches (still need to upload final versions of patches in use to Drupal.org)
- Removed some permissions from authenticated and anonymous roles to prevent config dependencies on unused modules (e.g. comment and contact)
  - See https://www.drupal.org/node/3193348 
- Addressed various 9.3.x code deprecations:
  - https://www.drupal.org/node/2940438
  - https://www.drupal.org/node/2940031

### Remaining tasks:
- [x] Switch from 9.3.0-rc1 to 9.3.0 once released (before merging)
- [x] Finalize updated patches (replace drupal.org MR diffs with actual patches)
- [ ] Add DB update for uninstalling quickedit (?)
  - See https://www.drupal.org/node/3227039 
- ~Add DB update for installing access_unpublished (?)~
- [x] Check if optional HTML paragraph module changes user role permissions (and potentially re-export)
- [ ] Peer review & testing!

## Related Issue

- Closes #1180
- #1340 

## How Has This Been Tested?

Locally with lando

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

- [x] I've added this Pull Request to the AZ Quickstart Github project and set it to "Review in progress".
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.

PR #1578, opened 2022-05-18
#1340 Allow Import of Dependent Permissions

tadean:
This PR alters the 9.3 permission changes to allow for two kinds of permissions to be assigned on the fly; provided permissions and dependent permissions. Previously there was no issue with provided permissions, but in the case of dependent permissions they were not being assigned (eg. `edit any book content` being provided by `node`, but actually existing because of `book`). In this PR, we attempt to additionally detect if a module contains a piece of configuration the permission depends on, which we take to assume that it should have a chance for quickstart to set its permissiions.

Additionally, we alllow the profile a chance to install permissions for its dependent modules.

## Related issues
#1340 
#1181 

## How to test
Test that bundle-based permissions are now enabled correctly.

- Patch release changes
   - [x] Bug fix
- Minor release changes
   - [x] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist


modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
```diff
@@ -167,12 +195,9 @@ protected function trimPermissions(array $config) {
    *   An array of permission definitions.
    */
   protected function getPermissionDefinitions() {
-    if (empty($this->permissionDefinitions)) {
-      // @todo Use injection on user.permissions.
-      // @phpstan-ignore-next-line
-      $this->permissionDefinitions = \Drupal::service('user.permissions')->getPermissions();
-    }
-    return $this->permissionDefinitions;
+    // @todo Use injection on user.permissions. and add caching.
```
tadean, line 198: Something seems to be going on with the caching that was here previously. With it enabled, it was receiving the wrong results for some kinds of permissions...investigation needed.

trackleft:
Tested with quick node clone and it worked
image

modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
joeparsons: > Something seems to be going on with the caching that was here previously. With it enabled, it was receiving the wrong results for some kinds of permissions...investigation needed.

The core patch that adds static caching to the PermissionsHandler service does this in `hook_modules_installed()`:
```
// Reset the user.permissions service to reset statically cached
// permissions.
\Drupal::getContainer()->set('user.permissions', NULL);
```

We may need to do something similar to reset the statically cached permissions in this plugin class, although I'm not sure what the best way to do so would be.

trackleft:
> Are they supposed to have the ability to assign actual admins?

They shouldn't be

joeparsons:
> > Looks like entity bundle permissions that were missing before exist now.
> > Others are still missing (probably expected):
> > 
> > * Role delegation permissions
> > * Full HTML text format permissions
> 
> Aren't these the role delegation permissions that were missing? role

Yes.  Those are the ones.  I'll try rebuilding my local env and try again.

joeparsons:
> Are they supposed to have the ability to assign actual admins?

No.  I was seeing none of them enabled in my testing but I was also experimenting with re-exporting config so I might have messed something up.

tadean:
Not sure if this would be related, but those permissions will only show up on a new drupal install, if they'd missed their opportunity to be set the first time (since their module gets installed before the profile).

joeparsons (approved):
Please disregard my earlier comments/questions about missing permissions.  When I initially cloned the Quickstart repo to test this locally, I foolishly did `git checkout -b issue/1340` instead of `git checkout issue/1340` 🙃.

After checking out the correct branch instead of creating a new one, I was able to verify that all expected permissions existed.  I also did several experiments with re-exporting the roles after installing optional core/contrib modules and found that all permissions in our roles are valid (tied to a module provided by Quickstart) except for 3 that I created a separate issue (#1581) and PR (#1582) for.

I think I'd be OK with merging this as-is so we can expedite creating a stable 2.3.x release based on Drupal 9.3.x ASAP but we'll probably want to create a follow-up issue to investigate whether anything can be done to reduce the performance impact of these changes:

<table>
<tr>
	<td>Task
	<td>main (#1582)
	<td>This PR (latest revision)
<tr>
	<td>Install Quickstart
	<td>2 minutes 11 seconds 72 milliseconds
	<td>2 minutes 56 seconds 153 milliseconds
<tr>
	<td>PHPunit tests
	<td>16 minutes 41 seconds 709 milliseconds
	<td>23 minutes 10 seconds 91 milliseconds
<tr>
	<td>Migration tests
	<td>4 minutes 28 seconds 43 milliseconds
	<td>5 minutes 18 seconds 618 milliseconds
</table>

trackleft:
@tadean mentioned a follow-up task for tracking site-owner selected permissions in a config file we could use to determine the site owner's intended permissions so we do not ever override those.

joeparsons:
Follow-up issues:

- #1595
- #1596
- #1597

PR #1632, opened 2022-06-02
Removed line break in static map, so full width media row  can be migrated.

trackleft:


trackleft:
I'm not sure if this is actually a thing.

PR #1650, opened 2022-06-07
Closes #1623 - Add Split Screen component

### LINKED ISSUE #1623: Add Splitscreen Component
## Motivation
Many University websites utilize a similar design concept of having the screen split in half––half image half content. It also tends to be a feature request on websites. Given the popularity/frequency of the concept, it would be beneficial to make this component available on Quickstart. 

- Aiming for the 2.4 release, but may have to be pushed to the next release depending on how many changes need to be made.
- "Splitscreen" is the name the module/config currently uses, since it "splits" the screen in half. Open to name suggestions from the group.

akslay:
## Description
- Adds the Split Screen paragraph type as an optional component to add to a Quickstart website (off by default).
- Adds more color options (grayscale) to the Text on Background paragraph behavior

## Related issues
Closes #1623

## How to test
https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-1650.probo.build/
- Enable the "Quickstart Paragraphs - Split Screen" module
- Add a Split Screen component to a Page
- Add a Text on Background component to the Page, and select one of the new background color options 

- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [x] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist


smayden:
This is cool! Tested as content editor with no problems.

I wonder if the background pattern should be removed for now since all those triangle patterns are likely to be deprecated.

akslay:
Shoshana is back from the dead! lol

I don't mind removing the triangles, but if we do that it will remove them for Text on Background components as well (since they use the same paragraph behavior). Might be a good follow-up issue for whenever we rebrand?

smayden:
> Shoshana is back from the dead! lol
> 
> I don't mind removing the triangles, but if we do that it will remove them for Text on Background components as well (since they use the same paragraph behavior). Might be a good follow-up issue for whenever we rebrand?

Ah, didn't realize they were linked. Yeah sounds like a follow up issue. I didn't notice that if I chose triangles on the same side as the image, they flash briefly as the image loads. Not a huge deal.

danahertzberg:
I agree, we could just remove the option from this paragraph type, right? I also noticed the background patterns were applied to the bottom of the element instead of the top as it is in the text on background.

I also wonder if we should have the space around content option? It seems to make sense to me to have consistent padding around all sides no matter what. We could always add it in later if it becomes a problem.

And on the grays, I propose we select a few of them to add instead of all of them. (Remove black)
- `Light gray` bg-gray-100 (matches events)
- `Gray` bg-gray-300 (matches btn-light and bg-light)
- `Dark gray` bg-gray-800 (matches bg-dark)

Follow up tickets could be for:
- Adding grayscale colors to text with background and cards
- Updating the bootstrap version in the rendered paragraph preview on the edit page
- Update calendar filter widget to use bg-gray-300 instead of bg-cool-gray as background colors

danahertzberg:
> @danahertzberg I think I would rather keep all of the gray background colors as an option, since they're offered in AZ Bootstrap and give editors more flexibility when it comes to designing their content. For example, `bg-gray-500` might better match the photo they uploaded than `bg-gray-300` does.

I understand the sentiment, but also want to balance brand and user experience. I'd like to get MarCom's review on this before merging anyways, so we can see what they say on this topic as well.


> For the padding issue, I'm not sure if there's really a way for us to remove only that option unless we change how the paragraph behavior currently funnctions (which would also impact Text on Background, and any other paragraph types that use it). 

I must not have the right understanding of how paragraphs with behaviors work. I thought we could pull in specific behaviors, not all of them? Cards, accordions, and text with background have some different and some shared behaviors. @trackleft @tadean @joeparsons Can you confirm?

akslay:
Follow-up issues:

- [ ] Add grayscale background colors

modules/custom/az_paragraphs/az_paragraphs_splitscreen/css/az_splitscreen.css
```diff
@@ -0,0 +1,37 @@
+/* Mobile/Base styles */
```
joeparsons, line 1: Nitpick: Looks like this file is using tabs instead of spaces.  We should change to spaces before merging.

akslay:
Bringing over points of discussion from Chris' PR (that was merged in to this PR) so that we don't loose track of them. These items should be addressed as follow-up tasks and will not be included in this PR.

From @trackleft :
> @tkmccarthy said in reference to the text area
> 
> Can you give it a max width so it never breaks outside of the normal content width?
> 
> [Screen Shot 2022-06-14 at 5 18 15 PM](https://user-images.githubusercontent.com/1023167/173710593-7a607729-8612-4bc7-ad01-17ad99f88c9f.png)
> 
> Which is only applicable in the full-width version.
>
> What if we had an option to align text to content grid if full-width setting is checked.

From @akslay :

> I think I would prefer making it an option to making it the default behavior. I think aligning it to the content grid works well for the screenshot above (since it's only 1-2 sentences), but doesn't work as well for a Splitscreen using a paragraph or more. here's an example on a site I'm building where more content is used, and the image is a graphic rather than a photo. Aligning to the content grid makes the graphic look off-center, since its supporting content aligns with the other page content, but the image now appears not to (feels offbalanced).
> 
> I think keeping it as true full-width is more flexible for a wider variety of content and images, but having a checkbox option for content grid would be okay. How would we implement that? That would be a major change to the current implementation of the Splitscreen markup. An alternative display mode?
> 
> https://user-images.githubusercontent.com/52931751/173868528-c99d78be-16bf-46f6-b3ee-07b8907c2ff4.png

joeparsons (approved):

trackleft (dismissed):

danahertzberg:
I recommend using the large image size (scale to 1140) instead of natural for best mobile responsiveness.

Follow up tasks:
- [ ] Allow for responsive images
- [ ] Option to align body text to content grid

modules/custom/az_paragraphs/az_paragraphs_splitscreen/config/install/core.entity_view_display.paragraph.az_splitscreen.default.yml
```diff
@@ -0,0 +1,96 @@
...
+targetEntityType: paragraph
+bundle: az_splitscreen
+mode: default
+content:
+  field_az_media:
+    type: media_thumbnail
+    label: hidden
+    settings:
+      image_link: ''
+      image_style: az_natural
```
danahertzberg, line 85:
```suggestion
      image_style: az_large
```

modules/custom/az_paragraphs/az_paragraphs_splitscreen/config/install/core.entity_view_display.paragraph.az_splitscreen.default.yml
```diff
@@ -0,0 +1,96 @@
+langcode: en
+status: true
+dependencies:
+  config:
+    - field.field.paragraph.az_splitscreen.field_az_media
+    - field.field.paragraph.az_splitscreen.field_az_text_area
+    - image.style.az_natural
```
danahertzberg, line 7:
```suggestion
    - image.style.az_large
```

tadean (approved):

joeparsons (approved):

danahertzberg (approved):

jwmoore1 (approved):

PR #1666, opened 2022-06-11
Closes #1569 Disabled honeypot module by default.

### LINKED ISSUE #1569: Disable Honeypot by default since it's not used in standard installs
## Problem/Motivation
Honeypot module helps protect public-facing forms from bots. In similar reasoning for why the Antibot module was included but left disabled, Honeypot can also be included but left disabled when it's not being used out of the box.

### Describe the bug
With a standard AZQS install, the only publicly available form is the login form. But the Honeypot module doesn't seem to protect it per [Disable timestamp protection on user login forms](https://www.drupal.org/project/honeypot/issues/1260886).

Without any other public forms to protect, Honeypot module can be left disabled on standard AZQS installs.

## Proposed resolution
Disable Honeypot module on a standard AZQS install.

Optionally consider creating a Quickstart Webforms module (probably in a new issue) that enables Webform modules, submodules, related helpful Webform modules, Honeypot and/or Antibot, and configures reasonable defaults. This new custom module would be in line with the previous UAQS Webform module that existed.

joshuasosa:
## Description
Disabled honeypot module from being installed by default since it's not used by any forms on a standard AZQS install.

AZQS-provided permissions and settings are still applied properly if Honeypot is manually enabled.

## Related issues
Closes #1569.

## How to test
Install AZQS and check to see that Honeypot module is not enabled.

Tested locally on DDEV.

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [x] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_security/az_security.info.yml
+++ b/modules/custom/az_security/az_security.info.yml
@@ -5,4 +5,3 @@ core_version_requirement: ^9
 package: The University of Arizona
 dependencies:
   - az_core
-  - honeypot:honeypot


trackleft (approved):

trackleft (approved):

camikazegreen:
Agree with this, but we should wait until we have the AZ_Webforms module ready to merge at the same time. 
Will reference this in issue #594.

joeparsons:
Might also be a good candidate for recipes.

joshuasosa:
Closing for now with Dana direction to reduce old open PRs. With Cameron's note to wait for an AZ Webform module, #1569 could also potentially be resolved in the same PR that creates the module for #4922.

PR #1667, opened 2022-06-11
Closes #1607 Enabled WebAuth login link on login form.

### LINKED ISSUE #1607: Login form is missing a link to use CAS
## Problem/Motivation
On default UAQS installs, the login form was set to automatically redirect to CAS login.

On default AZQS installs, the login form shows the normal Drupal login with no CAS option, which can be confusing for users trying to log into the site with their NetID.

Per #1170, the merged solution was to provide an option in AZ Quickstart settings to disable the standard /user/login path (and /user login redirect by association). But this setting is not enabled by default, leaving the standard login form available for standard installs. Since CAS is enabled by default, an option to get to the CAS login should be added to the login form.

Also related, users may be used to going to /user or /user/login as was available in UAQS sites. The solution to #1170 doesn't sufficiently address that user expectation when enabled.

Ideally, instead of disabling the form entirely and making it an Access Denied page, the setting should instead either hide the form and leave the WebAuth link on the page (as recommended by the CAS module) or redirect to CAS login as was the experience in UAQS.

### Describe the bug
The Drupal login page is available by default on new AZQS installs. The CAS login option is missing from the login form.

## Proposed resolution
The CAS WebAuth login option should be made available on the login form.
Enable "Place a link to log in via CAS on the standard /user/login form" in CAS settings by default.

### Expected behavior
A link to login with WebAuth should show up on the login form.

Alternatively, if CAS is intended to be the primary method of logins for deployed AZQS sites, follow the CAS module's guidance:
> If CAS is the primary way your users will log in, it is recommended to alter the login page in a custom module to hide the standard form.

Instead of disabling the page entirely as with the current AZQS AZ CAS setting option, change to hiding/removing the form or redirect to CAS login.

### Current login form
image

### Suggested login form
image

joshuasosa:
## Description
Adds the WebAuth login link on the user login form since CAS WebAuth is installed and enabled by default. This change makes it a bit easier for a user to understand how to log in with their NetID when the login form is available, which it is by default.

Optionally, another issue could be created to change the AZ CAS disable_login_form behavior to make /user/login redirect to /cas instead of making it an access denied page. Doing so would better match user expectations and training from the UAQS era and improve the user experience.

## Related issues
Closes #1607.

## How to test
Install AZQS and view the login form at /user or /user/login to see the link "Log in using University of Arizona WebAuth" above the form. The link destination may not work on demo/dev sites since UA's WebAuth limits domain access.

Visual change tested locally on DDEV.

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [x] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_cas/az_cas.module
+++ b/modules/custom/az_cas/az_cas.module
@@ -109,7 +109,7 @@ function az_cas_form_user_login_form_alter(&$form, FormStateInterface $form_stat
  * Modify the hook load order.
  */
 function az_cas_module_implements_alter(&$implementations, $hook) {
-  if ($hook == 'form_alter' && isset($implementations['az_cas'])) {
+  if ($hook === 'form_alter' && isset($implementations['az_cas'])) {
 
     // Move az_cas_form_alter() to the end of the list.
     // \Drupal::moduleHandler()->getImplementations()
@@ -121,4 +121,3 @@ function az_cas_module_implements_alter(&$implementations, $hook) {
     $implementations['az_cas'] = $group;
   }
 }
-
--- a/modules/custom/az_cas/az_cas.module
+++ b/modules/custom/az_cas/az_cas.module
@@ -9,6 +9,7 @@ use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\user\RoleInterface;
 use Drupal\Component\Utility\Html;
+use Drupal\Core\Url;
 
 /**
  * Implements hook_help().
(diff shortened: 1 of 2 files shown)


joshuasosa:
View the change at: https://2ed4716e-d2d8-43c5-94bf-92adeb29f082.probo.build/user/login

joeparsons:
I think this is a good improvement.  One nit-picky complaint is that the placement of the link in the form is kind of awkward (it would be better below the submit button IMHO):
Screen Shot 2022-06-15 at 10 32 04 AM

Looks like the current placement is due to this `hook_form_FORM_ID_alter()` implementation in the CAS module itself:
https://git.drupalcode.org/project/cas/-/blob/2.x/cas.module#L212-241

We could probably add our own `hook_form_FORM_ID_alter()` implementation to `az_cas` that adjusts the `#weight` of the `cas_login_link` element.  That could also be a potential follow-up improvement.

modules/custom/az_cas/az_cas.module
```diff
@@ -69,3 +70,55 @@ function az_cas_validate_roles(array &$form, FormStateInterface $form_state) {
...
+  $form['#validate'][] = '_cas_user_login_validate';
+}
+
+/**
+ * Implements hook_module_implements_alter().
+ *
+ * Modify the hook load order.
+ */
+function az_cas_module_implements_alter(&$implementations, $hook) {
+  if ($hook == 'form_alter' && isset($implementations['az_cas'])) {
```
joeparsons, line 112: Coding standards fix to appease PHPCS
```suggestion
  if ($hook === 'form_alter' && isset($implementations['az_cas'])) {
```

modules/custom/az_cas/az_cas.module
```diff
@@ -69,3 +70,55 @@ function az_cas_validate_roles(array &$form, FormStateInterface $form_state) {
...
+    // \Drupal::moduleHandler()->getImplementations()
+    // iterates through $implementations with a foreach loop which PHP iterates
+    // in the order that the items were added, so to move an item to the end of
+    // the array, we remove it and then add it.
+    $group = $implementations['az_cas'];
+    unset($implementations['az_cas']);
+    $implementations['az_cas'] = $group;
+  }
+}
+
```
joeparsons, line 124: Coding standards fix to appease PHPCS
```suggestion
```

tadean:
Good improvement. I think it's a lot more noticeable below the button also, thanks!

joeparsons (approved):

tadean (approved):

joeparsons:
We decided to merge #1681 (which also includes these changes) instead.

PR #1668, opened 2022-06-14
Allow editing of paragraph display via UI.

trackleft:
This PR is against the issue/1623 branch.

## Description
Added demo content and used hook_field_group_pre_render_alter() to add classes to field group.
Updated readme.

## How to test
In Probo Navigate to /pages/split-screen
Edit and view the edit preview.

Navigate to paragraph types under structure
Select split screen
Edit display
Add field 

## Types of changes
Added demo content
Changed original PR to allow editing the display in the UI.


trackleft:
@tkmccarthy said in reference to the text area

Can you give it a max width so it never breaks outside of the normal content width?

Screen Shot 2022-06-14 at 5 18 15 PM

trackleft:
What if we had an option to align text to content grid if full-width setting is checked.

akslay:
> What if we had an option to align text to content grid if full-width setting is checked.

I think I would prefer making an option to making it the default behavior. I think aligning it to the content grid works well for the screenshot above (since it's only 1-2 sentences), but doesn't work as well for a Splitscreen using a paragraph or more. here's an example on a site I'm building where more content is used, and the image is a graphic rather than a photo. Aligning to the content grid makes the graphic look off-center, since its supporting content aligns with the other page content, but the image now appears not to.

I think keeping it as true full-width is more flexible for a wider variety of content and images, but having a checkbox option for content grid would be okay. How would we implement that? That would be a major change to the current implementation of the Splitscreen markup. An alternative display mode?

screencapture-live-sasg-financialaid-pantheonsite-io-types-of-aid-funding-for-indigenous-wildcats-2022-06-15-08_34_47

tadean (approved):

PR #1679, opened 2022-06-17
Closes #1678: Disable Drupal login form by default.

### LINKED ISSUE #1678: Disable default user login form by default
## Motivation
Follow-up to #1170.  In #1173 we added a new setting to the az_cas module for disabling the default Drupal login form but we did not enable the setting by default because it broke many of our existing automated tests.

## Proposed Resolution
We should update our tests to work with the disable login form setting enabled by default.

joeparsons:
## Description

- Turns on `disable_login_form` config setting provided by Quickstart CAS module by default
- Updates existing tests to prevent failures caused by new default `disable_login_form` setting:
  - Adds new test trait providing a `setup()` method that turns off the `disable_login_form` setting
  - Adds 2 new test base classes (one for Functional tests and one for FunctionalJavascript tests) that utilize the new test trait
  - Replaces base class on existing tests that use `drupalLogin()` with one of the new base classes
- Adds additional steps to ProboCI config to turn off the setting on ProboCI review sites

## Related issues

Closes #1678 

## How to test

- Ensure Drupal login form is still usable on ProboCI site

- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [x] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist

### DIFF
--- a/.probo.yaml
+++ b/.probo.yaml
@@ -55,7 +55,7 @@ steps:
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_editor azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_admin azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_user_admin azuseradmin
-      - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush config:set -y az_cas.settings disable_login_form false
+      - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush config:set -y az_cas.settings disable_login_form 0
       - chown www-data:www-data -R /var/www/html/sites/default/files
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush -y -n pm:enable environment_indicator
       - chmod 777 /var/www/html/sites/default/settings.php
@@ -121,6 +121,6 @@ steps:
       - drush -y cset az_migration.settings migrate_d7_protocol "https"
       - drush -y cset az_migration.settings migrate_d7_filebasepath "dev-${MIGRATION_SOURCE_SITE_NAME}.pantheonsite.io/"
       - drush -y cset az_migration.settings migrate_d7_public_path "sites/default/files"
-      - drush -y cset az_cas.settings disable_login_form false
+      - drush -y cset az_cas.settings disable_login_form 0
       - drush mim --group=az_migration
       - echo "https://${BUILD_ID}--site-migration.probo.build/"
--- a/.probo.yaml
+++ b/.probo.yaml
@@ -55,7 +55,7 @@ steps:
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_editor azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_content_admin azuseradmin
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush user:role:add az_user_admin azuseradmin
-      - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush config:set az_cas.settings disable_login_form false
+      - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush config:set -y az_cas.settings disable_login_form false
       - chown www-data:www-data -R /var/www/html/sites/default/files
       - $SRC_DIR/az-quickstart-scaffolding/vendor/bin/drush -y -n pm:enable environment_indicator
       - chmod 777 /var/www/html/sites/default/settings.php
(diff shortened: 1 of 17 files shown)


tadean (approved):

trackleft:
Looks good, though we should advise people to go back and turn the form off since we have no idea if people are using it and can't make a db update for this.

trackleft (approved):

joshuasosa (changes requested):
If there's no change yet to the function of "Disable login form" in the AZ Quicstart CAS settings screen, then I would not approve this PR yet. Enabling it disables the Drupal login page and gives an Access Denied error. This does not match QS1 parity.

To match parity and provide a better user experience per trained user expectations, the login form should redirect to CAS login.

trackleft:
@joshuasosa I was just about to say that

trackleft:
What if we added a hook_form_alter to set access = false on the form instead of this https://github.com/az-digital/az_quickstart/blob/c7a23ba631a5321c2628646845d37c320622a114/modules/custom/az_cas/src/Routing/AzCasRouteSubscriber.php

And then added markup to that page with a link to log in via webauth?

joeparsons:
The CAS module maintainers decided against porting the redirect login form option that was added to the D7 version of the module to D8 a long time ago (see https://www.drupal.org/project/cas/issues/2723569) so we had to come up with something else for Quickstart 2.

We can remove the `QS1 feature parity` label from this issue/PR if that makes folks feel better about this approach.

joeparsons:
> What if we added a hook_form_alter to set access = false on the form instead of this [`c7a23ba`/modules/custom/az_cas/src/Routing/AzCasRouteSubscriber.php](https://github.com/az-digital/az_quickstart/blob/c7a23ba631a5321c2628646845d37c320622a114/modules/custom/az_cas/src/Routing/AzCasRouteSubscriber.php)
> 
> And then added markup to that page with a link to log in via webauth?

I think you could already do this with a custom 403 page.

trackleft:
> I think you could already do this with a custom 403 page.

Good point

joshuasosa:
> The CAS module maintainers decided against porting the redirect login form option that was added to the D7 version of the module to D8 a long time ago (see https://www.drupal.org/project/cas/issues/2723569) so we had to come up with something else for Quickstart 2.

If you check out the CAS settings for placing a CAS login link, their recommendation is as follows:
> If CAS is the primary way your users will log in, it is recommended to alter the login page in a custom module to hide the standard form.

There are several better options that could work when enabling this AZ CAS setting:

- Add a check in twig template to remove the form and leave the CAS login button.
- Create an AZQS-provided means of redirecting since CAS decided to leave it out.
- Create a custom page that provides a link to the CAS login.

All of these would be better than an unfriendly Access Denied error.

> We can remove the `QS1 feature parity` label from this issue/PR if that makes folks feel better about this approach.

No, that would not make me feel better and would not satisfy the unfriendly approach currently in place for the setting.

trackleft:
I know `/cas` is public and any crawler can find it, but are there security implications to pointing bots at the `/cas` path once the login path they are looking for redirects?

joeparsons:
> There are several better options that could work when enabling this AZ CAS setting:
> 
> * Alter the login form as suggested, possibly through a hook.
> * Add a check in twig template to remove the form and leave the CAS login button.
> * Create an AZQS-provided means of redirecting since CAS decided to leave it out.
> * Create a custom page that provides a link to the CAS login.

We attempted to come up with a way to provide the redirect functionality in Quickstart a long time ago but we ran into issues with redirect loops.

I wouldn't be opposed to coming up with a way to alter the login form or potentially provide a custom page that links to CAS although, I think the latter could already be done by configuring a custom 403 page.

It's worth noting that helping users find `/cas` is not the only reason this functionality is/was needed.  We also want this functionality to be able to make the login form altogether inaccessible to malicious traffic.

joshuasosa:
> I wouldn't be opposed to coming up with a way to alter the login form or potentially provide a custom page that links to CAS although, I think the latter could already be done by configuring a custom 403 page.

A custom page specifically for this situation provided by AZQS would be good. A custom 403 page may not be ideal, since there can be situations where a person is logged in but does not have permission to view a page. Adding verbiage to go to /cas to login would be incorrect in those situations since they're already logged in.

> It's worth noting that helping users find `/cas` is not the only reason this functionality is/was needed. We also want this functionality to be able to make the login form altogether inaccessible to malicious traffic.

There's always a balance to be decided between user experience/convenience and security. I don't know what the webauth login system can handle in terms of traffic or bot attempts, and if it would even matter if one common path to it was turned off. My assumption would be that traffic coming from /user/login channels to webauth doesn't affect security since it's been in place like that for QS1 sites for many years, but I could be wrong.

For some more comparison, nasa.gov and asu.edu turns it off for their main sites, but not for subdomains like science.nasa.gov/user/login, appliedsciences.nasa.gov/user/login, law.asu.edu/user/login or engineering.asu.edu/wp-login.php. Some other universities put in redirects like harvard.edu/wp-login.php and standard.edu/wp-login.php which redirect to their org login pages.

joeparsons:
I've created an alternate PR that changes the functionality of the AZ CAS `disable_login_form` setting here: #1681

joeparsons:
We decided to merge #1681 instead.

PR #1682, opened 2022-06-21
Closes #1557: Removed id hash from CSS selector for Search submit button

### LINKED ISSUE #1557: Search button css sometimes disappears.
## Problem/Motivation
I've encountered the issue a few times after updating where the search submit button styles disappear.
At first we thought this was related to Google custom search, but I've since encountered it on sites the don't use google custom search.

image

### To Reproduce
Unknown but it appears that the button's id changes 
image

The selector for the style is `.search-block-form .input-group-append button#edit-submit`

## Proposed resolution
Change the selector to not depend on an id.

cpoff:
## Description
The background color on the search submit button behaves erratically. It appears to depend on the ID (#edit-submit), which can change. So I removed the ID from the selector (.input-group-append button#edit-submit) in the main style.css

## Related issues
Closes #1557 

## How to test
This should impact the applied style each time the search appears in the nav bar. 
- Add a search term to the top search bar, click Submit. 
- Background should be white on initial load, and turn red on hover & click

- Patch release changes
   - [x] Bug fix
- Minor release changes
   - [ ] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist

### DIFF
--- a/themes/custom/az_barrio/css/style.css
+++ b/themes/custom/az_barrio/css/style.css
@@ -399,7 +399,7 @@ padding: 0.625rem .75rem;
   height: 43px;
 }
 
-.search-block-form .input-group-append .az-search-block button {
+.search-block-form .input-group-append button {
   font-size: 20px;
   color: #8B0015;
   background-color: #ffffff;
--- a/themes/custom/az_barrio/templates/forms/form--search-block-form.html.twig
+++ b/themes/custom/az_barrio/templates/forms/form--search-block-form.html.twig
@@ -10,7 +10,7 @@
  * @see template_preprocess_form()
  */
 #}
-<form{{ attributes.addClass('search-form', 'search-block-form', 'az-search-block').removeClass('form-row') }}>
+<form{{ attributes.addClass('search-form', 'search-block-form').removeClass('form-row') }}>
   <div class="input-group">
     {{ children }}
   </div>


akslay:
Looks fine to me from both the anonymous and authenticated roles on desktop & mobile.

In order to prevent further issues with any class/ID changes that happen down the road, it might be worth while to apply a more unique class like `az-search-block` under the block configuration settings, and then to adjust the CSS specificity accordingly. How do we feel about making that kind of change to improve long-term maintenance?

image

bberndt-uaz:
> Looks fine to me from both the anonymous and authenticated roles on desktop & mobile.
> 
> In order to prevent further issues with any class/ID changes that happen down the road, it might be worth while to apply a more unique class like `az-search-block` under the block configuration settings, and then to adjust the CSS specificity accordingly. How do we feel about making that kind of change to improve long-term maintenance?
> 

If we'd like to use a more specific class, we might go ahead and make that change in the [search block form twig](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/forms/form--search-block-form.html.twig) template. @akslay pointed out that if we make the change in that file instead of in the block configuration settings, people wouldn't be able to delete that class (accidentally or intentionally) which is required for our base styles.

trackleft:
Remember, blocks are reusable and this will affect each instance.

tadean (approved):

joeparsons (approved):

bberndt-uaz (approved):

PR #1720, opened 2022-07-05
Closes #1694: Add granular class selector to search block Submit button

### LINKED ISSUE #1694: Add more granular CSS selector for Search button
> Looks fine to me from both the anonymous and authenticated roles on desktop & mobile.
> 
> In order to prevent further issues with any class/ID changes that happen down the road, it might be worth while to apply a more unique class like `az-search-block` under the block configuration settings, and then to adjust the CSS specificity accordingly. How do we feel about making that kind of change to improve long-term maintenance?
> 

If we'd like to use a more specific class, we might go ahead and make that change in the [search block form twig](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/forms/form--search-block-form.html.twig) template. @akslay pointed out that if we make the change in that file instead of in the block configuration settings, people wouldn't be able to delete that class (accidentally or intentionally) which is required for our base styles.

_Originally posted by @bberndt-uaz in https://github.com/az-digital/az_quickstart/issues/1682#issuecomment-1162203502_

cpoff:
The submit box on the search bar in the main nav was using an obscure CSS selector, so I added `az-search-block` class and applied it to the main CSS sheet. This change should maintain existing functionality and give us a way to style that block specifically without breaking other placements.

## Description
Added class `az-search-block` to style.css and also to the Search block using the Block Configure tool (/admin/structure/block/manage/az_barrio_search?destination=/admin/structure/block)

## How to test
Go to arizona.edu homepage and search. When it takes you to the Search page, search again. The submit button should be white on pageload, and red when you hover.

Off canvas considerations.

## Types of changes
Added new CSS class to existing entry. Applied that class using the Drupal blocks UI.

- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [ ] New feature
- [ ] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist


themes/custom/az_barrio/css/style.css
```diff
@@ -399,16 +399,21 @@ padding: 0.625rem .75rem;
...
   font-size: 20px;
   color: #8B0015;
   background-color: #ffffff;
   border: 1px solid #ced4da;
   padding: 0 8px;
 }
 
-form.search-block-form .input-group .input-group-append button:hover,
-form.search-block-form .input-group .input-group-append button:focus {
+.az-search-block button {
```
joshuasosa, line 416: These properties are a duplicate of lines above, 404-405. I'm not sure why the original lines were removed, but they should be re-added and this new line added to the selector list. Lines 402-406 here should be removed.

The `az-search-block` selector could be added to the button:hover and button:focus selectors to be consistent with `.az-search-block .search-block-form .input-group-append button`.

joshuasosa (commented):
The new class, `az-search-block`, isn't being applied to the offcanvas search form. If the purpose of this issue/PR is to "improve long-term maintenance," the same class or a different new class should be applied in block.block.az_barrio_offcanvas_searchform.yml.

themes/custom/az_barrio/css/style.css
trackleft, line 416: Like this?
```suggestion
.az-search-block form.search-block-form .input-group .input-group-append button:hover,
.az-search-block form.search-block-form .input-group .input-group-append button:focus {
```

themes/custom/az_barrio/css/style.css
```diff
@@ -399,16 +399,21 @@ padding: 0.625rem .75rem;
...
 }
 
-.search-block-form .input-group-append button {
+form.search-block-form .input-group .input-group-append button:hover,
+form.search-block-form .input-group .input-group-append button:focus {
+    background-color: #8B0015;
+    color: #ffffff;
+}
+
+.az-search-block .search-block-form .input-group-append button {
```
trackleft, line 408:
```suggestion
.az-search-block form.search-block-form .input-group .input-group-append button {
```

joshuasosa: Yeah I think this and the other suggest change look good.

trackleft: Thanks!

tadean (approved):

jwmoore1 (approved):

PR #1984, opened 2022-10-07
Closes #1983 Using local patches

### LINKED ISSUE #1983: Possibly move all patches into the quickstart repo to guard against drupal.org deleting patches when they move to gitlab. (and reduce the load on drupal.org)
## Conditions of satisfaction
- [ ] Patches should all apply

Example directory structure.

```
az_quickstart/patches/composer/drupal/cas/patchfile.diff
az_quickstart/patches/composer/drupal/better_exposed_filters/patchfile.patch
```

trackleft:
## Description
When Drupal.org switches to gitlab for issues, and CI, there may, or may not be a way to link to ingest patches tied to a specific commit.

In my opinion, switching to local patches is a good way to help streamline the drupal development process for contrib, and ensure patch stability for our project.

Allows applying patches to composer packages whose install path is not the traditional `vendor` directory because of the use of the [composer-installers](https://github.com/composer/installers) and [composer-installers-extender](https://github.com/oomphinc/composer-installers-extender) composer plugins.
 
When running composer install directly on az-digital/az_quickstart, This PR disables cweagans/composer-patches since this is not a normal use-case and is generally only done by CI tools like dependabot and patches are not needed when just checking version numbers.

## Related issues

Closes #1983 

- Patch release changes
   - [ ] Bug fix
- Minor release changes
   - [ ] New feature
- [x] Other or unknown

### Drupal core
- Patch release changes
   - [ ] Security update
- Minor release changes
- [ ] Other or unknown

### Drupal contrib projects
- Patch release changes
   - [ ] Security update
   - [ ] Add new module
- Minor release changes
   - [ ] Major level update
- [ ] Other or unknown

## Checklist


trackleft:
When running composer install directly on az-digital/az_quickstart, This PR disables cweagans/composer-patches since this is not a normal use-case and is generally only done by CI tools like dependabot and patches are not needed when just checking version numbers.

joeparsons (approved):
This looks good to me.  Thanks!  Tried figuring out a way to do this before but was unsuccessful.  I like this approach.

The only remaining suggestion I have would be to add some documentation about adding/removing patches somewhere.

joeparsons:
One other minor drawback that comes to mind with this is that there's a bit less inline documentation about where patches came from (specific issue/comment number) if we're not referencing d.o. URLs directly.  I wonder if we should also update the keys in the patch lists to include more info like `<issue-number>#<comment-number>` or <mr-number>, etc..  (Just a thought).

trackleft:
I agree, it is a good thought.

trackleft:
Saw this on drupal.org

> 
> The contents of merge request patch file will change as commits are pushed!
> 
> This means that you will get the latest version of the code from an issue. So, it brings a security risk because any new commit to the issue’s branch will be automatically deployed to your site when composer runs.
> 
> It's strongly recommended to lock the patch file versions on production sites. To do so you must download the patch file and use it in composer from a local directory. If you use the URL to the Gitlab MR directly, your codebase will change without warning, as people work on the merge request.
> 
>

trackleft:
Updated for Quickstart 10.4

joeparsons:
Per discussion in 2025-01-29 PR meeting, we're going to close this for now.  We may revisit in future.

PR #2067, opened 2022-11-16
Fixes #2014 Entity embed preprocess consumes elements.

### LINKED ISSUE #2014: Embedded Entity migrate process plugin sometimes consumes subsequent entities
### Describe the bug
Sometimes the entity embed processor, when processing an embedded entity, will delete a following element that appeared after the embed tag. This is most likely due to the arithmetic of the element replacement.

The parsing of the DOM should also be set to not output parsing errors.

### To Reproduce
Migrate using the entity embed processor.

tadean:
Entity embed preprocess emits warnings when parsing invalid tags, and sometimes consumes following elements to the embed.

## Description
Turned off warnings from DomDocument HTML parsing, and wrapped the tags in a temporary root tag. Libxml requires that a root tag exists, or it emits a closing copy of the first tag encountered. See [saveHTML()](https://www.php.net/manual/en/domdocument.savehtml.php)

## Related issues
#2014 

## How to test
Migrate content that contains both an embed and a following tag in the same html block and verify that no warnings are emitted and that the following tags don't disappear.

- **Patch release changes**
   - [x ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

- [x ] My code follows the code style of this project.

### DIFF
--- a/modules/custom/az_migration/src/Plugin/migrate/process/EntityEmbedProcess.php
+++ b/modules/custom/az_migration/src/Plugin/migrate/process/EntityEmbedProcess.php
@@ -216,7 +216,9 @@ class EntityEmbedProcess extends ProcessPluginBase implements ContainerFactoryPl
     $value = mb_convert_encoding($value, 'HTML-ENTITIES', 'UTF-8');
 
     $dom = new \DOMDocument('1.0', 'UTF-8');
-    $dom->loadHTML($value, LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD);
+
+    // @see https://www.php.net/manual/en/domdocument.savehtml.php#121444 Libxml requires root element.
+    $dom->loadHTML('<entity-embed-process>' . $value . '</entity-embed-process>', LIBXML_HTML_NOIMPLIED | LIBXML_HTML_NODEFDTD | LIBXML_NOWARNING | LIBXML_NOERROR);
     $elements = $dom->getElementsByTagName("drupal-entity");
 
     // Configuration of custom content.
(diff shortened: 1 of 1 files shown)


joeparsons (dismissed):

trackleft (dismissed):
We could try using https://git.drupalcode.org/project/migrate_plus/-/blob/6.0.x/src/Plugin/migrate/process/Dom.php and ripping out our own implementation of DOMDocument, however I believe the non-root option will use Masterminds\HTML5.


We would also have to edit many of our migrations to do this.

bberndt-uaz (approved):

joeparsons:
@tadean would like to do some more testing before we merge this.

tadean:
> We could try using https://git.drupalcode.org/project/migrate_plus/-/blob/6.0.x/src/Plugin/migrate/process/Dom.php and ripping out our own implementation of DOMDocument, however I believe the non-root option will use Masterminds\HTML5.
> 
> We would also have to edit many of our migrations to do this.

I took @trackleft 's suggestion about [Dom](https://git.drupalcode.org/project/migrate_plus/-/blob/6.0.x/src/Plugin/migrate/process/Dom.php) and refactored my setting the tags directly to making this filter a child class of Dom with the necessary defaults set.

joeparsons (approved):

tadean:
Tested with HTML special characters to make sure they are not altered by the plugin:
html_test

bberndt-uaz (approved):

trackleft (approved):

PR #2077, opened 2022-12-02
Closes #1993 - Add "modal" view display to AZ Person

### LINKED ISSUE #1993: Add "Modal" view display to AZ Person
## Motivation
A common feature request is for the AZ Person directory to display Person details in a modal instead of loading the node page.

- [ ] Create a "Modal Grid" view display in the AZ Person Directory view (fields instead of display mode)

akslay:
## Description
Adds a "Modal" view display to the AZ Person view.

- Clone of the "Grid Page" view display (but as a block)
- Displays name/job title/bio/picture in modal on click
- Provides link to view full profile within modal

Review site: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-2077.probo.build/
- [Directory Modal](https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-2077.probo.build/directory-modal)

Issue we might want to look in to: the on click behavior of the modal launching prevents editors from using contextual links to edit individual Person nodes.

## Related issues
Closes #1993 

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


akslay:
Questions to discuss:

- What fields should be included in the modal?
- Should bio field be trimmed or not trimmed?

HenryzGreenberg:
@akslay 

I will run your questions by the UX team and report back. 
As far as the edit button not appearing consistently. Would it make sense to disable the modal view when a user is logged into QS and is editing the page? Food for thought.

camikazegreen:
Having an inconsistent experience between logged in and anonymous users can be very confusing for site editors. They would definitely expect it to work when they first enable that view, and would universally be logged in for that experience.

joshuasosa:
This looks like an interesting grid option! Great idea! We actually use modals like this in our [LPL digital signage](https://www.lpl.arizona.edu/digitalsignage) area.

I agree, user experience should be consistent between anonymous and authenticated, as much as possible.

At the moment, seeing the edit options come up and not be able to click on them is confusing and bad UX for content editors. I would like to see that addressed before merging.

In the modals, I'd suggest using the same heading level for Person Categories as is used on profile pages. I think trimming the summary could be good, or adding a "read more" option that trims it to one paragraph and expands when clicked to show any subsequent paragraphs.

I'd also like to see the same info on the grid replicated into the modal: email and phone number.

It looks like the modal is using az_square image style? I'd prefer sticking to either the az_card_image used on the grid or az_small used on the profile page instead of having yet another image style in the mix.

HenryzGreenberg:
After consulting with the team, we agree a trimmed display would be a better solution. The idea of a "quick view" modal with a link to the full profile "detail" page would be moot if we are already delivering all the content in the quick view.  Also for future reference we'd be open to providing ideas for solutions before development begins. The Senior UI/UX Designer lives to solve UI/UX problems. :)

joshuasosa:
Another note: it looks like email needs the truncated style in the cards, which would help it match up with the original grid display.

akslay:
Design options from Tom: https://xd.adobe.com/view/873f212f-0184-4397-af6c-2f2e78bd8a68-d26c/grid

**Decisions:**

- Include phone(s) & email links in the modal. Separate all phone numbers and the email by a vertical bar
- Switch "close" to `btn-outline-blue`
- Swap "close" and "view more" button ordering
- Move "close" icon outside of modal header, and delete modal header
- Remove top border from modal footer
- Include a "Card - Trimmed" display mode for content. Same as "Card" but with phone numbers and email removed on the Person display
- Switch from "bio" to "Summary" field
- On mobile, the "close" icon should sit on top of the image ([use CSS "filter" property](https://css-tricks.com/reverse-text-color-mix-blend-mode/) to change the icon color based on the image it's sitting on top of). Don't need a background color behind it in this case for accessibility, since there's three different ways to dismiss/close the modal.
- Use "Card" image style on [mobile](https://xd.adobe.com/view/873f212f-0184-4397-af6c-2f2e78bd8a68-d26c/screen/f5038c72-06e8-49a2-b5af-a43755316341/) for the Modal
- Use the "Default" image style on [desktop](https://xd.adobe.com/view/873f212f-0184-4397-af6c-2f2e78bd8a68-d26c/screen/430937b4-f64a-481f-a5d2-726fddbb6be5/) for the modal
- 
Screen Shot 2023-01-11 at 2 14 20 PM



**Future Improvements:**

Find a way to make this type of functionality obtainable from within the "Manage Display" settings for entities. Some options include:
- Create a custom field group (modal)
- Add an option on the "Link" field group type 
- Add a link formatter for when you output links, to where they are a link to trigger a modal, accordion, etc.
(@trackleft is creating an issue for this)

HenryzGreenberg:
I don't know if we can use the "Card" image style on mobile for the Modal. As the cropping from mobile to desktop appears to be different. While the mock-up shows an ideal cropping of the image I don't think a headshot will crop quiet as nicely. Do we have any options here?

joshuasosa:
> I don't know if we can use the "Card" image style on mobile for the Modal. As the cropping from mobile to desktop appears to be different. While the mock-up shows an ideal cropping of the image I don't think a headshot will crop quiet as nicely. Do we have any options here from a development standpoint. As in a way to set a focal point of the image regardless of image cropping?

I was never a fan of card style images being used on directories. I think portrait photos should be using portrait-oriented images, not landscape-oriented like the card style.

But besides that, as mentioned above, I think the modal should probably use the same image style as is in the full profile.

akslay:
> As in a way to set a focal point of the image regardless of image cropping?

We actually already have a backlog [issue for adding in the Focal Point module](https://github.com/az-digital/az_quickstart/issues/693) to Quickstart. So I can work on that item separately from this PR.

I also agree that the card image crop isn't the best (not a huge fan either), but I'm also not overly concerned about using the Card crop on mobile, since it's no different than what we currently offer on mobile for the "Grid" layout. It's essentially the same as a card once it gets to mobile since the content stacks.

akslay:
Next steps: Set up a follow-up meeting to come to a consensus on what we want with:
- Dana Hertzberg
- Ashley Slay
- Troy Dean
- Josh Sosa
- Tom McCarthy

trackleft (approved):

akslay:
Decisions from today's discussion (2022/12/18):

- Contact info will be present on the modal display, so that people don't have to close out of the modal in order to get back to an email address, for example.
- Will move forward with the trimmed bio as the default behavior of the modal, because having a full bio would entail whitespace under the image once the content gets longer, which would make the design less than ideal for that length of content. However, the view display does remain modifiable by site admins, so they could remove the trimming, and also remove the "view full profile" button if desired.
- We're going to add a new crop/image style for the modal image. Slightly taller than card image style, but also not the entire/default image (likely 5:4 aspect ratio with cropping 568px (wide) by 710px (tall).

akslay:
Since the Person content type does not use the "summary" field, I kept the bio field as trimmed text. If we want to use summary, we will have to add the summary field to AZ Person. Is this something we want to do or not do?

akslay:
- Will keep the trimmed bio instead of summary
- Will remove the "no contact" view display and add as a future improvement for the other view displays as well (not just modal)

camikazegreen:
The Reduce Duplicates functionality doesn't seem to be working as expected even though that box is checked in the view.

For an example, see this page with both the standard grid and the modal grid with identical settings. Wilma is displayed twice in the modal, since she is tagged as both Staff and Undergraduate even though the views are both filtered to only display Undergraduates (16).
https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-2077.probo.build/directory-modal

trackleft:
Another thing to note, this PR's functionality slightly conflicts with another PR https://github.com/az-digital/az_quickstart/pull/2168

modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -298,6 +305,954 @@ display:
         - user.permissions
       tags:
         - 'config:field.storage.node.field_az_person_category'
+  az_person_grid_modal:
+    id: az_person_grid_modal
```
danahertzberg, line 309: Recommend `grid_modal` to match other view using just `grid`

camikazegreen:
> Another thing to note, this PR's functionality slightly conflicts with another PR #2168
Curious what the "View Full Profile" button would do in that case.

modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -298,6 +305,954 @@ display:
...
+          group_rows: true
+          delta_limit: 0
+          delta_offset: 0
+          delta_reversed: false
+          delta_first_last: false
+          multi_type: separator
+          separator: ', '
+          field_api_classes: false
+        nothing:
+          id: nothing
```
bberndt-uaz, line 1068: Maybe it would be good to change `nothing` on these two lines to something like "person_card" and `nothing_1` below to "person_modal"?

akslay: Had a discussion about it this morning. Sounds like we'll just keep the IDs as "nothing" since those are automatically generated by the view and we won't have access to change them in the UI (so we're not sure if there will be implications in forcing the change in the YML).

akslay:
> Another thing to note, this PR's functionality slightly conflicts with another PR https://github.com/az-digital/az_quickstart/pull/2168

I'll add an IF to the twig that uses the other link, instead of the node URL, if it exists.

akslay:
> I'll add an IF to the twig that uses the other link, instead of the node URL, if it exists.

Actually, if I do this, that means we need to make Metatag a dependency for Person, which we don't necessarily want to do since not all sites have the Metatag module installed.

trackleft:
> Another thing to note, this PR's functionality slightly conflicts with another PR #2168

Resolved ☝️

trackleft (approved):

camikazegreen (approved):

PR #2173, opened 2023-01-23
Closes #2169 Close "x" icon broken in media library edit module's modal/dialog

### LINKED ISSUE #2169: Close icon broken in media library edit module's modal
## Problem/Motivation
While testing a recent update (dependabot PR) for the Media Library Edit module, @bberndt-uaz noticed an issue with the close icon in the modal dialog used by the Media Library Edit module included in Quickstart.

### Describe the bug
From https://github.com/az-digital/az_quickstart/pull/2128#issuecomment-1377629526:

> This update introduces an issue with the close button on the pop-up for editing media: when hovering the cursor over the button, the appearance is incorrect.
> 
> Before this PR:
> 
> 
> After this PR:
> 
>
>It looks like the .ui-icon background image is now being loaded from /core/assets/vendor/jquery.ui/themes/base/images instead of /core/themes/claro/images.

This bug seems to be associated with this change that was made to the module ahead of the 3.0.0 release:
https://www.drupal.org/project/media_library_edit/issues/3292509

## Proposed resolution
Contribute a patch to drupal.org for fixing this issue since it likely affects anyone using the module with Paragraphs and the Claro theme.

trackleft:
## Description
See #2169
Adding patch from https://www.drupal.org/project/drupal/issues/3335935

## Related issues
Closes #2169

## How to test

1. Create a new flexible page
2. Add a card deck
3. Add another paragraph type with media, such as Text on Media
4. Add an image to the second paragraph
5. Click the pencil "edit" icon to view the incorrect appearance for the close icon
6. Reload the page
7. Edit the second paragraph, edit the media item, and observe the correct close icon
8. Click to edit the card deck, then return to edit the media item in the second paragraph and observe the correct icon

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joshuasosa:
I don't think this is a bug with core. I can't reproduce this issue in vanilla Drupal 10. I think it's an issue with the cards paragraph type.

I can't reproduce this issue when creating a flexible page and just adding text on media or photo gallery. But I _can_ reproduce it for both of those once the cards paragraph is added, whether cards are added first or later.

It looks like in Drupal, the loading order is usually correct, with dialog.css loading later than jquery ui's theme.css. Dialog.css therefore already overrides jquery ui for that button in vanilla.
image

But for some reason, when adding in the cards paragraph type and then trying to add media to it or other media sections elsewhere in the form, the loading order gets swapped.
image

The Claro issue may be unnecessary, and the patch solution a band-aid for another underlying issue with the cards paragraph type.

tadean:
> But for some reason, when adding in the cards paragraph type and then trying to add media to it or other media sections elsewhere in the form, the loading order gets swapped.

Good catch @joshuasosa ! I wonder if this might be due to the `media_library_form_element` form element present on the card form. it declares dependencies on several core libraries that might possibly be changing the load order:

```
  dependencies:
    - core/drupal.dialog.ajax
    - core/sortable
    - core/drupal
    - core/jquery
    - core/once
    - media_library/style
```

Or maybe it's an ajax issue with the card form itself.

trackleft:
Does this mean that dialog.css should have a dependency on theme.css in claro then?

trackleft:
After troubleshooting, I found that if I add the following to claro.info.yml

```
libraries-extend:
  media_library_form_element/media_library_form_element:
    - claro/claro.drupal.dialog
    - claro/claro.drupal.dialog
 ```
 
 Extensions load after the library it is extending.
 Dependencies load before the library depending on it. 
 
 
It does appear to be an ajax issue, since the library dependencies are appended after all of the files already in <head>

These things did not work in claro.info.yml (as far as I could tell)

```
libraries-extend:
  core/drupal.dialog:
    - claro/claro.drupal.dialog
    - claro/claro.drupal.dialog
 ```
 
```
libraries-extend:
  core/drupal.dialog.ajax:
    - claro/claro.drupal.dialog
    - claro/claro.drupal.dialog
 ```
 Since the below code already exists in claro.info.yml
 ```
   core/drupal.dialog:
    - claro/claro.drupal.dialog
    - claro/claro.jquery.ui
 ```
 And, `core/drupal.dialog` is a registered dependency of `core/drupal.dialog.ajax`, one would think that the extensions for `core/drupal.dialog` would be loaded after  it, but that does not happen.
 
 If I add to media_library_form_element.libraries.yml
 This also does not work 
 ```
 media_library_form_element:
  version: 2.0
  js:
    assets/js/media_library.form-element.js: {}
  dependencies:
    - core/drupal.dialog.ajax
    - core/sortable
    - core/drupal
    - core/jquery
    - core/once
    - media_library/style
    - core/drupal.dialog
 ```
 
 So at the moment, my best working title for an issue on drupal.org is...
 
 > library extensions of library dependencies not loaded in the correct order on ajax call.

joshuasosa:
I don't think it's just a Claro issue. Testing with Seven, the close icon looks off in there as well for just card paragraph types, and other types once a card paragraph is added.

### Seven theme normal close icon on text on media paragraph
image

### Seven theme broken close icon on card paragraph
image

I still think it's premature to have an issue in core queue when it hasn't been demonstrated to be replicable yet in vanilla. It could be a configuration issue within the AZQS project, an issue within another contributed module, or as this testing with Seven goes, an issue within another core module rather than the theme.

joshuasosa:
Adding animated examples with Seven admin theme.

### Seven media library text on media (normal)
media_close_text_media

### Seven media library card (not normal)
media_close_card

You can see there are also other style differences between the two. The blue button in the normal media library dialog box becomes a white button. There's some minor differences with the font.

It's just another example where jquery ui's theme.css loads normally for other media areas (or rather in this case, doesn't actually even factor in in the media library), and then doesn't load normally for card paragraph type (incorrectly loads later and takes over button style from Seven's buttons.css and font family from Seven's elements.css).

Things are similarly broken with card paragraph media library in contrib themes like Gin and Adminimal.

trackleft:
I think the issue is actually this patch: https://github.com/az-digital/az_quickstart/blob/main/composer.json#L154

joshuasosa:
> I think the issue is actually this patch: https://github.com/az-digital/az_quickstart/blob/main/composer.json#L154

I think that's it. I tested locally: I get the issue before changing anything, and the issue goes away after removing [this patch](https://www.drupal.org/project/drupal/issues/1461322#comment-13669426).

PR #2176, opened 2023-01-25
Closes #2175 Default Pathauto pattern does not reflect parent custom paths

### LINKED ISSUE #2175: Default Pathauto pattern does not reflect parent custom paths
## Problem/Motivation
This issue was filed after further testing of our breadcrumbs in Quickstart revealed a problem with the way automatic URL aliases are generated for pages.

### Describe the bug
The "Default" Pathauto pattern in Quickstart is set to `[node:menu-link:parents:join-path]/[node:menu-link]`. All nodes for Flexible Pages and for any site-specific content types will have automatic URLs generated according to that pattern. (The other Quickstart content types--Events, News, and Person--have their own defined Pathauto patterns.)

The current pattern uses menu link titles to construct the automatic URL for each page. This pattern causes problems when site editors specify custom URLs for pages and then add child pages underneath those pages in the site navigation menu. In the URL for the child page, the part representing the parent page is not the actual path of the parent page but is instead a string derived from the menu link title of the parent page. Because of this discrepancy, the breadcrumb for the parent page is not linked correctly since the path to the parent page as indicated by the URL does not resolve to an actual page: it isn't the custom URL that the user specified for that parent page.

### To Reproduce

1. Create a page with a custom alias and add it to the main navigation menu:
image

2. Create a second page with an automatic alias and add it to the main navigation menu as a child of the custom alias page:
image

3. Observe two issues:
 - The URL for the child page uses a string of the parent page's menu link instead of the parent page's actual relative path. This could be potentially confusing for users if the URL keeps changing as they progress deeper into a tree of pages that include custom aliases. For example, instead of going from /custom-alias-one to a child page at /custom-alias-one/automatic-alias-two, the visitor would be taken from /custom-alias-one to /menu-link-title-one/automatic-alias-two.
 - Since there is no page located at the URL based on the parent page's menu link, the breadcrumb for that page is grayed out and is not a link.

## Proposed resolution
Rather than using `[node:menu-link:parents:join-path]/[node:menu-link]` as the default Pathauto pattern in Quickstart, we could use the pattern `[node:menu-link:parent:url:relative]/[node:title]`, as suggested in [this comment on an issue for the Token module](https://www.drupal.org/project/token/issues/2769299#comment-11496877). This pattern would correctly preserve the custom paths of parent nodes.

With this new pattern, the URL for child page in the example above is automatically generated with the correct path and breadcrumb for the parent page:

image

### Expected behavior

As a visitor to a website, I expect URLs to be consistent and to reflect the menu hierarchy as I follow links through the site. Also, I expect breadcrumbs for parent pages to be functional and to take me to those pages, whether or not those pages have custom paths.

bberndt-uaz:
## Description

Closes #2175 by changing the "Default" pathauto pattern from:

- [node:menu-link:parents:join-path]/[node:menu-link]

To:

- [node:menu-link:parent:url:relative]/[node:title]

This may be a disruptive change which requires significant discussion, but I thought I would go ahead and create this simple PR for testing purposes. I'm open to any feedback!

## Related issues

#2175 

## How to test

To test the issue that this PR is meant to address:

1. Create a page with a custom alias and add it to the main navigation menu.
2. Create a second page with an automatic alias and add it to the main navigation menu as a child of the custom alias page.
3. On the child page, check that the breadcrumb for the parent page works and that the URL includes the actual path of the parent page. That is, you can visit the parent page if you delete the child page's title from the URL.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [X] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/config/install/pathauto.pattern.default.yml
+++ b/config/install/pathauto.pattern.default.yml
@@ -6,7 +6,7 @@ dependencies:
 id: default
 label: Default
 type: 'canonical_entities:node'
-pattern: '[node:menu-link:parents:join-path]/[node:menu-link]'
+pattern: '[node:menu-link:parent:url:relative]/[node:title]'
 selection_criteria: {  }
 selection_logic: and
 weight: 0
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -129,7 +129,7 @@ function az_core_toolbar() {
  */
 function az_core_pathauto_pattern_alter(PathautoPatternInterface $pattern, array $context) {
   // Manually specify default AZQS pattern.
-  $azqs_pattern = '[node:menu-link:parents:join-path]/[node:menu-link]';
+  $azqs_pattern = '[node:menu-link:parent:url:relative]/[node:title]';
 
   // Check if context is node and pattern is the AZQS default.
   if ($context['module'] === 'node' && $pattern->getPattern() === $azqs_pattern) {


bberndt-uaz:
For now, this PR is on hold while we seek a better understanding of the impact of this change. See the [associated issue](https://github.com/az-digital/az_quickstart/issues/2175) for further discussion.

joeparsons:
Talked about this in our 2025-02-07 Friday meeting.  We're still scared to do this but would like to look into whether how we could apply this change to _only_ new Quickstart site installs.

PR #2316, opened 2023-03-30
Fixes #2313 Update Citation Style Configuration

### LINKED ISSUE #2313: Update Citation Style Configuration
In the 2.6 update of `seboettg/citeproc-php`, the package has begun using the newly available packages `citation-style-language/styles` and `citation-style-language/locales`. Previously these had to be installed by hand as the repositories were not available on packagist, so were not really available to the quickstart build process.

The availability of these packages may mean some changes to our citation style configuration would be beneficial.

## Conditions of satisfaction
Now that these packages are available, we should consider if any revisions are needed to our process for including styles and locales in our citation styles. At minimum, we can likely stop using `academicpuma/locales` - this older package was in use as a stopgap because the citation locale repository was not yet available on packagist.

Our custom style entities currently contain the actual xml of the style - we may wish to consider instead referencing a style name in the `citation-style-language/styles` package.

tadean:
## Description
The `seboettg/citeproc-php` package has updated in its `2.6.0` release to begin using the `citation-style-language/locales` and `citation-style-language/styles` packages for its CSL rules. These packages were not previously available on packagist and it used to be a manual step to install them.

Previously, Quickstart was using `academicpuma/locales` (an older copy of the same locale files that was available on packagist) for its locale information, to prevent needing to install them manually.

Now that these styles are available on packagist, we may want to consider refactoring how we include the styles. This PR updates the Publication content type to use `citation-style-language/locales` and modifies our configuration style entities to prefer referencing CSL styles in the `citation-style-language/styles` package by name rather than including the xml directly.

Modifications are made to the style configuration form to allow either referencing a CSL style by name, or by custom XML. Most of our managed styles now refer to the `citation-style-language/styles` by reference. Bluebook Law Review currently still uses custom XML as the `citation-style-language/styles` version of bluebook law review does not support bibliographies The custom version of quickstart's Bluebook Law Review managed style has been modified to support bibliographies.

_Spike: Feedback is desired on the best way to include references to the stylesheets, given there are a very large number of options in `citation-style-language/styles`; dropdown built via cached introspection of the files, specifying name manually, etc._

## Related issues
#2313 

## How to test

- enable `az_publication`
- create some publications
- verify citations are still rendered correctly
- visit `/admin/config/az-quickstart/settings/az-publication/az_citation_style`
- create a new style or edit one
- For the new style, enter a stylesheet name or custom CSL (see [Citation Style Language Repository](https://github.com/citation-style-language/styles) for examples. Enter stylesheet names without the `.csl` extension, e.g. `modern-language-association`
- change the default site style at `/admin/config/az-quickstart/settings/az-publication` to your style and verify they work correctly

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joeparsons (dismissed):
Code changes look good to me.  I'm glad to see that classes/methods for discovering available stylesheets are provided by the CiteProc library so you didn't have to implement any of that. 👍🏻 

Adding autocomplete to the stylesheet selection form element would definitely be a nice future improvement but this version seems totally usable to me as-is and I would be in favor of merging this PR.

joeparsons:
## Suggested change from 2023-04-05

Add a boolean (checkbox) config entity setting and form field that indicates whether the citation style config entity contains CSL or not instead of having 2 separate text fields for the style name and custom style markup.

tadean:
**Todo:**
- [ ] Refactor to use boolean config switch on style loading
- [ ] Add boolean checkbox config setting for style behavior
- [ ] AJAXify form to transform style element based on style type

modules/custom/az_publication/src/Entity/AZQuickstartCitationStyleInterface.php
```diff
@@ -27,4 +27,30 @@ public function getStyle();
...
+   * @return string
+   *   The CSL stylesheet of this citation style.
+   */
+  public function getStyleSheet();
+
+  /**
+   * Returns the custom csl stylesheet.
+   *
+   * @return string
+   *   The custom CSL stylesheet of this citation style.
```
tadean, line 42: Update docblock comments to reflect new API.

modules/custom/az_publication/config/install/az_publication.az_citation_style.az_ama.yml
```diff
@@ -3,4 +3,5 @@ status: true
...
+style: american-medical-association
+custom: 0
```
trackleft, line 7: Traditionally, we've used the boolean type
See https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/config/schema/az_barrio.schema.yml#L160-L162
See https://github.com/az-digital/az_quickstart/blob/eb2f6418c61f6c3dcf8d9bfb6b3ef250e7697a2c/themes/custom/az_barrio/config/install/az_barrio.settings.yml#L2

bberndt-uaz (approved):
Tested in Probo and everything looks good for this PR!

However, I did notice a potential issue with the way pages are displayed by the MLA style:

image

image

tadean:
@bberndt-uaz it seems that MLA uses the `minimal-two` CSL page numbering style. It looks like the intended output of this field should be `140-50`, removing the redundant one, so it seems **some** transformation of this number is expected, but it doesn't seem like the zero should be dropped. I'll do some testing and see if this is an issue in the CSL processor we're using.

modules/custom/az_publication/config/schema/az_citation_style.schema.yml
```diff
@@ -11,5 +11,8 @@ az_publication.az_citation_style.*:
     style:
       type: string
       label: 'Style'
+    custom:
+      type: integer
```
joeparsons, line 15: I like @trackleft's suggestion to make this an actual boolean property instead of integer.
```suggestion
      type: boolean
```

tadean: Thanks @joeparsons ! I was on the fence about which way to go on this. For some reason `Drupal\Core\Render\Element\Checkbox` uses integer `0` an `1` as its normal values and not boolean (not sure why).

I do think an actual boolean in the config makes more sense - I can update this.

joeparsons (approved):

bberndt-uaz (approved):

trackleft (approved):

joeparsons:
@tadean Is going to create a follow-up issue for addressing the page number issue.

Issue #2237, opened 2023-02-17
Convert AZ Publication from Experimental to Stable


trackleft:
## Proposed Resolution
The following tasks/questions should be done/answered before marking the AZ Publication module Stable.

joeparsons:
New bugs identified that should be addressed before we promote the module:

- [ ] #2362
- [ ] #2417

PR #2358, opened 2023-04-21
Fixes #2247 Publication BibTeX Import

### LINKED ISSUE #2247: Publication BibTex Import
## Motivation
I'd like to be able to import publications in the [BibTex](https://www.bibtex.com/format/) format. This will make it easier to enter large numbers of publications since many publishers allow BibTex export.

## Proposed Resolution
Provide an importer to import BibTex citations as publications in a quickstart site.

A migration utility that takes an uploaded BibTex document as input and converts it into a migration array suitable for use with `az_publication` 

Manual entry of publications, other importer types in the future.

tadean:
## Description
This PR adds an importer that creates Publication content from [BibTeX](https://www.bibtex.com/g/bibtex-format/) documents. To support this, a migration data parser `az_bibtex` (via providing a wrapper around `renanbr/bibtex-parser`) is added handles the parsiing of BibTex documents into migration source rows.

As many services that track publications allow for BibTeX export, this serves as a path for importing Publication content from other sites.

## Related issues
#2247 
#2237 (part of, does not close)

## How to test
- enable `az_publication` and `az_publication_bibtex` and `az_publication_doi`
- visit `/admin/content/az-publication/import/bibtex`
- upload a BibTex document and press `import`
- visit `/admin/content/az-publication/import/doi`
- enter a valid DOI and press `import`
- verify that for each import appropriate links are displayed via status message to the content via the title of the publication imported
- create a publication view
- inspect imported output

- **Patch release changes**
   - [ ] Bug fix
   - [x] Adding experimental module
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


tadean:
**Todo:**

- [ ] Migration of Date/Year/Month/etc fields
- [ ] Import currently very slow for large datasets - to be fixed. (Likely data parser iterator or `MigrateBatchExecutable` issue)
- [ ] Add paired DOI import form that accepts a DOI link and leverages the BibTex import
- [ ] Add additional output to show what was imported

tadean:
In prior discussion, we identified that the scope of TeX support for these BibTeX imports is to provide support for special characters only (rather than complete TeX rendering)

tadean:
- Updated to add a specialized DateProcessor for BibTeX import
- Added DOI import form (using BibTeX) based on DOI Content Negotiation

trackleft:
FYI, I recieved this error after enabling az_publication_bibtex (it might be a problem with my site, but wanted to post this here, in case it isn't)

image

trackleft:
Another issue I had while using this PR on a local copy of a real site, was that the citation styles config where not being imported, not sure why

image

trackleft:
The DOI importer, however worked like a charm.

tadean:
> FYI, I recieved this error after enabling az_publication_bibtex (it might be a problem with my site, but wanted to post this here, in case it isn't)
> 
> ```
>  [error]  Could not retrieve source count from az_publication_bibtex_import: file parser plugin: could not retrieve data from profiles/custom/az_quickstart/modules/custom/az_publication/az_publication_bibtex/data/test.bib 

This is a test data file which isn't intended to be committed - I will remove that path from the migration.

tadean:
> For the DOI importer piece I used
> 
> ```
> 10.1111/j.1748-720X.2012.00678.x

Thanks for testing this @trackleft ! `\mathsemicolon` is TeX special character that will be handled by the remaining task ` Implement minimal TeX support necessary for rendering special characters in UTF8`. Right now there are a lot of TeX characters not yet handled. The source data looks a little unusual though - I will take a look at this.

tadean:
> Another issue I had while using this PR on a local copy of a real site, was that the citation styles config where not being imported, not sure why

I'm curious if this is due to the changes in #2313 where `custom` was added as a key in the schema file for citation styles (which your site may not have) in its config entities. This was for the "This style uses custom Citation Style Language" setting on the citation styles. Could you test if the presence of that key in `modules/custom/az_publication/config/schema/az_citation_style.schema.yml` makes a difference, or if you save a citation style (which would set the `custom` key) prior to running the distro update? Maybe we need a DB update that initializes that key for styles that are missing it.

tadean:
> For the DOI importer piece I used
> 
> ```
> 10.1111/j.1748-720X.2012.00678.x
> ```

> Searched for `mathsemicolon` and found this old issue https://gitlab.com/crossref/issues/-/issues/1731

You're right about this @trackleft - it does look like some weird escaping is going on here on the source side.

modules/custom/az_publication/az_publication_bibtex/src/Plugin/migrate_plus/data_parser/AZBibtex.php
joeparsons, line 1: Would it make any sense to create service definitions for any of the classes that are instantiated in this plugin (RenanBr\BibTexParser\Listener, RenanBr\BibTexParser\Parser, RenanBr\BibTexParser\Processor\NamesProcessor, Drupal\az_publication_bibtex\Processor\AZDateProcessor) and use the dependency injection container to inject them into this plugin class?

tadean:
- Updated to add an `Quickstart Publication Import` module to organize menu links and common functionality.
- Relocated import logging to `az_publication_import` common module.
- Added partial LaTeX processing of special characters to BibTeX parser.
- Removed URL to non-existent test data.

tadean:
Previous task `More complete error handling for failed parsing` to be handled in a future unrelated PR for improvements to Batch UI Migrations. We have several other migrations using the Batch Executable API and all of these could benefit from more nuanced error handling - any Batch API migration errors currently just create AJAX errors.

tadean:
- Moved `Import Publication` action link to content admin page.

modules/custom/az_publication/az_publication_import/az_publication_import.links.action.yml
joeparsons, line 1: Our News Feeds and Trellis Events importer modules both add task links instead of action links (tabs instead of buttons).  See:
https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_news/az_news_feeds/az_news_feeds.links.task.yml#L1-L5
https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_event/az_event_trellis/az_event_trellis.links.task.yml#L7-L12

Could we change this to be a task link (tab) instead of an action link (button)?

tadean: Definitely - updated to make this change.

tadean:
> FYI, I recieved this error after enabling az_publication_bibtex (it might be a problem with my site, but wanted to post this here, in case it isn't)
> 
> ```
>  [error]  Could not retrieve source count from az_publication_bibtex_import: file parser plugin: could not retrieve data from profiles/custom/az_quickstart/modules/custom/az_publication/az_publication_bibtex/data/test.bib 
> ```

@trackleft in my testing it appears that this was due to the missing test data file, which has been removed from the migration.

tadean:
> Another issue I had while using this PR on a local copy of a real site, was that the citation styles config where not being imported, not sure why
> 
> image
> I am currently using [d723ef4](https://github.com/az-digital/az_quickstart/commit/d723ef46468569db59a0db460eb7567d18daf53d)
> 

The schema for these styles was updated in [8d33a3e](https://github.com/az-digital/az_quickstart/commit/8d33a3e53a5d1effed1fe4e1fa9d10dfb86e1b45). If the actual site was running a revision previously to this (such as `2.6.3` or `2.6.4`) it would not have seen the schema change yet so it may consider the `custom` parameter invalid.

In my testing (with `2.6.4`), it seems that if I added the PR to the site, but did not clear the cache after running `drush cd-update`, the style changes would not be imported successfully. (I'm guessing because it's expecting the schema from prior to the changes to `modules/custom/az_publication/config/schema/az_citation_style.schema.yml` file, maybe?). If I began from 2.6.4, added the PR, but did clear the cache before running `drush cd-update` then the styles were imported successfully.

I'm guessing that perhaps a layer of complexity on this might be that if the import fails the first time, **it might still get written to the config snapshot table as a distro update the site has 'seen', meaning it thinks it's applied the change**, but you're left without the changes to the `custom` parameter on the citation styles. @trackleft does this match what you see in your testing?

joeparsons (approved):

trackleft (approved):
Can't wait to show this to Law

PR #2373, opened 2023-04-26
Closes #2371: Transform IDs in both source and destination paths for migrated redirects.

### LINKED ISSUE #2371: Redirect migration redirects the wrong node if using the `/node/*` path instead of the aliased path
## Problem/Motivation
For migrated redirects, the incorrect node ends up getting redirected if that redirect is using the `/node/*` path instead of the aliased path. 

For example, a redirect source of `/node/46` on the source site ends up redirecting the incorrect node on the migrated site, due to changes in node ID's during the migration process.

## Proposed resolution
Is there a way in the migration table to take the source node ID from the redirect, and then update that redirect on the migrated site to be the proper (new) node ID? 

### Expected behavior

Redirects using `/node/*` as the source path should still function correctly on the migrated site, instead of redirecting the incorrect page that now has the same node ID.

joeparsons:
## Description
Modifies existing `az_redirects` migration:

- Processes redirect `source` value with `az_migrated_path_lookup` migrate plugin
- Previously just the redirect `redirect` (destination) value was being processed with `az_migrated_path_lookup`

This will allow node or term IDs in both the redirect source path and destination path values to be transformed to the corrected migrated IDs.

## Related issues
Closes #2371

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_migration/migrations/az_redirects.yml
+++ b/modules/custom/az_migration/migrations/az_redirects.yml
@@ -29,6 +29,12 @@ process:
       - az_news_tags
       - az_person_categories_secondary
       - az_person_categories
+    node_migration:
+      - az_node_carousel
+      - az_node_event
+      - az_node_flexible_page
+      - az_node_news
+      - az_node_person
     source: source
   redirect_source/path: "@source_path_processed"
   redirect_source/query:
--- a/modules/custom/az_migration/migrations/az_redirects.yml
+++ b/modules/custom/az_migration/migrations/az_redirects.yml
@@ -21,11 +21,20 @@ process:
     -
       plugin: default_value
       default_value: 1
-  redirect_source/path: source
+  source_path_processed:
+    plugin: az_migrated_path_lookup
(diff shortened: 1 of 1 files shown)


joeparsons:
FWIW, I verified that the regex used in our `az_migrated_path_lookup` process plugins only matches values that start with either `node/xxx` or `term/xxx` so it should be safe to process both the source and destination fields of the redirects being imported.

trackleft (approved):

Issue #1645, opened 2022-06-03
Add additional "3 column" displays for views that have a grid view display for use on pages with one sidebar.


trackleft:
## Motivation
Many people have said tha this display looks bad:
<img width="775" alt="image" src="https://user-images.githubusercontent.com/1023167/171957675-e4c5ae07-7cef-4aa1-8f69-c1f4699b29e6.png">

### Is your feature request related to a problem? Please describe.

4 columns looks bad on sites with sidebars.
This includes az_news, az_events, az_person and az_flexible_page views.

### Describe the solution you'd like
Add an additional view display for each of the four views that have them with slightly altered breakpoint selections.

PR #2381, opened 2023-04-27
Closes #1396: Added 3-col view display for AZ Person

### LINKED ISSUE #1396: Change Grid Page display of AZ Person view to show fewer cards per row
## Problem/Motivation
The cards on the default Quickstart directory page are too narrow, causing names and email addresses to break across two lines.

Related issue: #1328 

### Describe the bug
Currently, on desktop screens, the Grid Page display of the Quickstart AZ Person view shows four cards per row. This results in some last names being pushed to a new line and all email addresses ending in @email.arizona.edu being broken across two lines. This behavior also occurs on tablet screens, where this display shows three cards per row.

### To Reproduce

1. Go to the directory page with a grid display on any Quickstart 2 site or enable the demo content in a test Quickstart site.
2. Notice how many names and email addresses are broken across two lines.

## Proposed resolution
Change the styling for this display to show three cards per row for desktop screens and two cards per row for tablets. This change would allow most names to appear on one line. The email address for people with shorter NetIDs would also appear on one line, even if the address ends in @email.arizona.edu.

### Expected behavior

I expect the default page display to provide a better fit for the typical content.

##  Screenshots
Current grid display:

image

Proposed grid display:

image

 - OS: macOS
 - Browser: Chrome
 - Version: 2.2

### Smartphone (please complete the following information):
 - N/A

cpoff:
## Description
The default view display is 4 columns on large and x-large, so I added a 3-column option. Also changed the default medium layout to 2-column.

## Related issues

Closes #1396
Related #1645

## How to test
Run on local or Probo, create new page, add view, select AZ Person > Grid - 3 Col display, confirm the layout is three columns instead of four. Install demo content if needed.

## Types of changes
Added one additional view display that was cloned from Grid Page, using Duplicate as Block. Then adjusted the Format > Bootstrap Grid settings to specify the column prefs. Grid Page at 4-cols in still the default.

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -30,60 +30,12 @@ base_table: node_field_data
 base_field: nid
 display:
   default:
-    display_plugin: default
     id: default
     display_title: Master
+    display_plugin: default
     position: 0
     display_options:
-      access:
```
joshuasosa, line 38: There's lots of changes to Master and other displays. I'm not sure if the changes matter since the displays visually look the same to me.

modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -377,13 +375,139 @@ display:
...
+        - url
+        - url.query_args
+        - 'user.node_grants:view'
+        - user.permissions
+      tags:
+        - 'config:field.storage.node.field_az_person_category'
+  grid_3col:
+    id: grid_3col
+    display_title: 'Grid - 3 Col'
+    display_plugin: block
```
joshuasosa, line 444: Why was this added as a block and not a Page like Grid Page and Row Page?

I think this display option should be consistent with the other existing options and be its own Page instead of a Block.

cpoff: If it's a Page, it requires a path and builds the entire page. This is not meant to replace the default page, but rather to provide a three-col option.

joshuasosa: Yes, you can make it a page just like Grid Page and Row Page without replacing the 'default' page.

Grid Page exists at: /people
Row Page exists at: /people-list

You can keep this new grid_3col display consistent with those two existing display options by using their precedent and making it a page as well.

Grid - 3 Col could exist at: /people-3col

Then, this new display should also be added to demo content's menu.

cpoff: Yes, thanks. We discussed in the PR review this a.m. and agreed that Block was suitable. Working now on getting the Demo version in the menu and the weird Master changes cleaned up.

joshuasosa: Could you please explain how block is suitable when the precedent set by the other displays is to use page?

cpoff: Block is more flexible and allows editors to set their URL via the page settings rather than setting it in the view. This is meant to be another option, not to replace existing functionality. So the block is the addition.

joshuasosa: There is no suggestion here to replace existing functionality. Adding a page is an _addition_, not a _replacement_. I'm not sure why you keep thinking a page would replace anything.

Yes, block can be more flexible, but is not the standard already set by the other pages. If block is the new format to use, the /people-list should be converted to a block as well.

cpoff:
The additional code changes still appear in the 'clean' export versions. So I QA'd all the views against those on the [Demo site](https://demo.quickstart.arizona.edu/). No issues appeared.

joeparsons (approved):
## Summary of recent meeting discussions/decisions for this PR

- We decided that going forward we want to be more judicious about adding new page displays to views and, although we have many existing page displays on various views provided by Quickstart, there was strong consensus that block displays that are designed to be added to flexible pages via views paragraphs are what we want to recommend for new views/displays added to Quickstart.  We also created some issues to document this change and to figure out what (if anything we should do about our existing page displays (#2404, #2405)
- We spent a good amount of time during our 2023-05-17 AZ Digital workshop experimenting with different ways of adding the new display and re-exporting the view to ensure that the resulting changes (diff) to the exported view were as minimal as possible but it we could not come up with a way of adding the new display without the resulting export having a lot of re-ordered configuration changes, etc
- We did some extensive QA comparing the views configuration and rendered views on the ProboCI site for this PR against the demo.quickstart.arizona.edu site to ensure that no unexpected changes to the views were happening.

modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -305,15 +305,12 @@ display:
         - user.permissions
       tags:
         - 'config:field.storage.node.field_az_person_category'
-  grid:
-    display_plugin: page
-    id: grid
-    display_title: 'Grid Page'
-    position: 1
+  block_1:
+    id: block_1
```
bberndt-uaz, line 309: I recommend changing the machine name for this view display. Otherwise all looks good!

bberndt-uaz (approved):

modules/custom/az_person/config/install/views.view.az_person.yml
```diff
@@ -30,60 +30,12 @@ base_table: node_field_data
...
-          id: 0
-          total_pages: null
-      style:
-        type: views_bootstrap_grid
-        options:
-          grouping: {  }
-          row_class: ''
-          default_row_class: false
-          uses_fields: false
-          col_xs: col-12
```
akslay, line 77: We should probably only be changing the view by adding the 3-col display, and not modifying the default display at all. Although I agree we need to fix the breakpoints at smaller screen sizes, that should probably be added as a separate PR so that it's more easily tracked as an intended change down the road. I would recommend reverting any changes to the default/master display and having those be follow-up issues so we can keep this PR as a view display addition only.

bberndt-uaz (approved):

akslay (approved):

PR #2390, opened 2023-05-02
Closes #2361 - Add multilingual support

### LINKED ISSUE #2361: Add multilingual support
## Motivation
Requests for multilingual support are on the rise. Given that UArizona is a Hispanic Serving Institution, we should be providing a way for site owners to serve their content in multiple languages.

- Leverage core's Language module
- Provide default configuration
- Provide style customizations to the language toggle
- Provide documentation on how site builders can translate their content
- Content owners are responsible for their own translations (QS will not provide content translations, and is only providing the tools to make content multilingual)

## Requirements to be Determined:
- How much configuration should be provided by default? Or, what configuration should _not_ be provided?
  - Content types?
  - Menu links?
  - Blocks?
  - Paragraphs
- Should a new user role be provided with permissions to translate content?
- Should an existing user role have permissions to translate content?

### Potential Blockers
- LinkIt does not currently support multilingual behavior. If you are using a card deck (or other component that has a link field) on your translated page, you receive the error: “This link does not exist or you do not have permission to link to <path>”. 
   - Links work fine in the text editor, though. 
   - Including the links in the text editor instead of as the card link is a possible workaround. You just lose the ability to make the whole card clickable (unless you add .stretched-link class via the source of the card).
   - Curious to see if this is fixed with the 2.6 release (because this error has happened for documents too), since the site is only on 2.5.1 right now.

### Functionality
- It will duplicate the initial English content for you (for the Spanish translation), but after creating your translation, it will no longer automatically add the new content that you added in the English version to the Spanish version. (Makes sense, but something to make note of.)
- Translated pages will not show up in the LinkIt results, so you will need to manually link to the /es/* path of another Spanish page translation.
   - Easy workaround here could be to link to the English page, then go into the source and manually add /es to the path.
- Adding /es to the URL will switch the entire site to “Spanish mode”. Only content which has been given explicit translations will be swapped over, all other content will remain in the default language (English).

### Editor Experience
- When you click “edit” you will be editing the translation you are currently viewing (English or Spanish). If you want to modify the other translation, you need to use the “Translate” tab and edit it from there.
- Translations appear as 2 separate nodes on the Content list. Nothing indicating which is “Spanish” and which is “English”.
   - For editor experience and clarity, we might want to add another field to the Content view which indicates which language the content is in.
- The language toggle block will display on all pages by default, regardless of whether or not that page actually has a translation. Editors will need to restrict the block to specific pages for clarity, if the entire site isn’t being translated.

 

- Should a new user role be provided?
- Should an existing user role have permissions to translate content?

akslay:
## Description
Adds multilingual functionality to QS. 

Closes #2361
Review site: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-2390.probo.build/

Documentation has been added here, in addition to the module's README: https://quickstart.arizona.edu/create-content/translating-content?review=GQqaA9fjvtV--KB5sj6zL3wBPc8h6JhfXNPuVo5rk7U

### User role changes:

- `content admin` can install languages on the site
- `content editor` can make content translations 

### General Additions

- Defines which entities can be translated
- Provides styling for language toggle (uses `.btn-dark` styling to visually differentiate it from normal buttons [red/blue/white] on the site)
- Overrides the language block template
- Adds classes directly to link elements within the language block to avoid maintaining CSS overrides

## How to test
Enabling the Masquerade module will make testing faster.

As an `azadmin`:
1. Enable the `az_multilingual` module.

As an `azcontentadmin`:
1. Install a language on the site.
2. Place the Language Switcher Block.

As an `azcontenteditor`:
1. Create a content translation (as `az_content_admin`).
2. Toggle between languages to see the difference.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_multilingual/az_multilingual.module
```diff
@@ -0,0 +1,28 @@
+<?php
+
+/**
+ * @file
+ * Contains az_multilingual.module.
+ */
+
+/**
+ * Implements hook_preprocess_HOOK() for block templates.
```
joeparsons, line 9: Nitpick: 
```suggestion
 * Implements hook_preprocess_HOOK() for links--language-block templates.
```

modules/custom/az_multilingual/README.md
```diff
@@ -0,0 +1,25 @@
+# Quickstart International
+The Quickstart International module enables site editors to provide language translations for their site content. This module only provides the functionality necessary to serve content in multiple languages, and does not provide content translations themselves. It is up to the site owner(s) to provide translations for their site's content.
```
joeparsons, line 2:
```suggestion
# Quickstart Multilingual
The Quickstart Multilingual module enables site editors to provide language translations for their site content. This module only provides the functionality necessary to serve content in multiple languages, and does not provide content translations themselves. It is up to the site owner(s) to provide translations for their site's content. 
```

modules/custom/az_multilingual/quickstart/language.content_settings.path_alias.path_alias.yml
akslay, line 1: Uncertain whether or not we actually need to include this `.yml` file, or if it's just automatically included when you enable the module.

modules/custom/az_multilingual/quickstart/language.content_settings.redirect.redirect.yml
akslay, line 1: Uncertain whether or not we actually need to include this `.yml` file, or if it's just automatically included when you enable the module.

config/install/user.role.az_content_admin.yml
```diff
@@ -86,8 +87,8 @@ permissions:
   - 'edit terms in az_person_categories_secondary'
   - 'edit terms in az_publication_categories'
   - 'edit terms in tags'
-  - 'edit webform assets'
```
akslay, line 89: These shouldn't have been removed. Need to add back in.

config/install/user.role.az_content_admin.yml
```diff
@@ -86,8 +87,8 @@ permissions:
...
-  - 'edit webform submissions any node'
```
akslay, line 90: These shouldn't have been removed. Need to add back in.

modules/custom/az_multilingual/az_multilingual.module
```diff
@@ -0,0 +1,28 @@
...
+  $links = $variables['links'];
+  $lang = \Drupal::languageManager()->getCurrentLanguage()->getId();
+
+  if (!empty($links)) {
+    foreach ($links as $key => $link) {
+
+      // Include .active class for Bootstrap styling (Drupal uses .is-active)
+      if (!empty($lang) && !empty($links[$lang])) {
+        $variables['links'][$lang]['link']['#options']['attributes']['class'][] = 'active';
+      }
```
tadean, line 24: This section can be moved outside the `foreach` loop I think as it only depends on `$lang`

akslay:
- Content editor role should be able to make content translations
- Need to document the workflow blocker for modifying menu item

trackleft:
Overall, I think this looks good and will be a great addition to quickstart.

akslay:
Per the last meeting, we have chosen to [enable the Splitscreen module by default](https://github.com/az-digital/az_quickstart/pull/2396), which also eliminates the dependency issue upon install of `az_multilingual`.

config/install/user.role.az_content_admin.yml
```diff
@@ -78,6 +79,8 @@ permissions:
   - 'delete webform submissions own node'
   - 'edit any webform'
   - 'edit any webform submission'
+  - 'translate any entity'
+  - 'translate editable entities'
```
joeparsons, line 83: Question: Do we need (or want) these permissions on both the content editor and content admin roles?

trackleft:
Have you seen this issue: https://www.drupal.org/project/linkit/issues/2886455

modules/custom/az_multilingual/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+1. Navigate to a piece of content that you want to translate.
+2. Click on the "Translate" tab .
+3. Click on "Edit" for the language(s) you wish to add translations for.
+   - Add your content translations and save your changes.
+4. Use the language toggle block to switch between your content translation(s).
+
+### Important Notes
+
+- LinkIt does not currently support multilingual functionality. Recommend using LinkIt to link to the English version (regular link), then manually modify the URL in the link so it's prepended with the language code. Example: `/example-page` --> `/es/example-page`.
```
trackleft, line 32: We can add the drupal issue here: https://www.drupal.org/project/linkit/issues/2886455

config/install/user.role.az_content_admin.yml
trackleft, line 83:
```suggestion
```

joeparsons (approved):

modules/custom/az_multilingual/config/install/language.content_settings.paragraph.az_contact.yml
```diff
@@ -0,0 +1,17 @@
...
+third_party_settings:
+  content_translation:
+    enabled: false
+    bundle_settings:
+      untranslatable_fields_hide: '0'
+id: paragraph.az_contact
+target_entity_type_id: paragraph
+target_bundle: az_contact
+default_langcode: site_default
+language_alterable: false
```
trackleft, line 17:
```suggestion
language_alterable: false

```

modules/custom/az_multilingual/config/install/language.content_settings.node.az_flexible_page.yml
```diff
@@ -0,0 +1,17 @@
...
+third_party_settings:
+  content_translation:
+    enabled: true
+    bundle_settings:
+      untranslatable_fields_hide: '0'
+id: node.az_flexible_page
+target_entity_type_id: node
+target_bundle: az_flexible_page
+default_langcode: site_default
+language_alterable: true
```
trackleft, line 17:
```suggestion
language_alterable: true

```

akslay:
It looks like the translations are not working correctly on our custom field types (cards and accordions). There is no differentiation between the 2 versions, and the most recent edit is instead used for all translations. I would consider this a blocker for this PR.

It also looks like the YML files are not capturing all of the configuration set for a given content type. For example, on the Page content type, I only had the following options enabled for translations:
Screenshot 2023-05-11 at 11 42 25 AM


However, they are enabled upon a fresh install of the Quickstart Multilingual module:
Screenshot 2023-05-11 at 11 42 36 AM

Looking at the YML files for each content type, these individual checkboxes/configuration options aren't even included in the YML file, so I'm not sure how to ensure the configuration gets captures. I don't see any other config files within Config Sync/Export that would have captured that configuration either.

tadean:
> It looks like the translations are not working correctly on our custom field types (cards and accordions). There is no differentiation between the 2 versions, and the most recent edit is instead used for all translations. I would consider this a blocker for this PR.

In some testing locally it looks like cards do work with translations when configured by hand - it looks like the issue may be similar to what you noticed about the page content type, where some information about selected fields isn't seemingly represented in the `yml` files maybe...

akslay:
Need to take a look at the full archive rather than single export on the config:
https://github.com/az-digital/az_quickstart/wiki/How-to-export-configuration-for-a-quickstart-module-during-pull-request-creation#finding-hidden-configuration

trackleft:
image

trackleft (approved):

joeparsons (approved):

tadean (approved):

bberndt-uaz (approved):
Did some quick testing on a Pantheon multidev and did not encounter any issues.

PR #2469, opened 2023-06-08
Fixes #2362 Publication Views exposed filters overlap

### LINKED ISSUE #2362: Publication views exposed filters overlap
## Problem/Motivation
On pages that embed the az_publication view and also include a sidebar, the year field in the exposed form is not accessible while using the desktop breakpoint.

image

tadean:
Publication exposed form filters currently attempt to display them as a flex row all in one line in the Bootstrap large viewport size, with the year field allowed to flex shrink as it doesn't need to be as large as the other text fields. This approach works on the normal large viewport width, but fails when a sidebar is present as it causes the type filter to overlap once the year shrinks too far.

## Description
This PR takes an alternate approach and disables the `lg-nowrap` behavior. Instead, the size of the rendered text elements are reduced to have form elements that are not unnecessarily wide to begin with. This creates a situation where they are less likely to need to wrap, but can still wrap if necessary (e.g. pages with sidebars on both sides and similar circumstances.)

Note that this does not affect the maximum character limit of the form elements, only their physical width.

## Related issues
#2362 

## How to test

- Enable `az_publication`
- Create a flexible page with the Search display of the publication view.
- Create a publication
- Visit the flexible page.
- Verify that filters display correctly
- Add a sidebar block to appear on the page (ideally the menu block)
- Verify that the year form input is no longer covered by the type input
- Verify that form elements do wrap to the next line if necessary (mobile sizes etc)

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_publication/az_publication.module
+++ b/modules/custom/az_publication/az_publication.module
@@ -303,11 +303,11 @@ function az_publication_preprocess_views_view(&$variables) {
     // Only run for publication search.
     if ($variables['id'] === 'az_publications') {
       // Reduce the width of the text input form elements from the default.
-      $allow_shrink = [
+      $resize_element = [
         'name',
         'title',
       ];
-      foreach ($allow_shrink as $element) {
+      foreach ($resize_element as $element) {
         if (!empty($variables['exposed'][$element])) {
           $variables['exposed'][$element]['#size'] = 20;
         }
--- a/modules/custom/az_publication/az_publication.module
+++ b/modules/custom/az_publication/az_publication.module
@@ -302,9 +302,19 @@ function az_publication_preprocess_views_view(&$variables) {
   if (!empty($variables['id'])) {
     // Only run for publication search.
     if ($variables['id'] === 'az_publications') {
-      // Allow the publication year field to shrink if present.
+      // Reduce the width of the text input form elements from the default.
+      $allow_shrink = [
+        'name',
+        'title',
+      ];
+      foreach ($allow_shrink as $element) {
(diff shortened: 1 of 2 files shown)


tadean:
One option we may also want to consider is the removal of the `Publication Type` filter from the view. In practice this filter is not very useful because the View also contains a contextual filter for the type already. In Quickstart design patterns it may be a common use case for editors to create view pages using contextual filters that show specific kinds of publications (see the screenshot in #2362 for an example of this, I think?). Contextual arguments and exposed filters do not combine well with each other because they both enforce their own query conditions.

Another limitation is that the Type filter innately displays all of the options, which means many options exist which may have no publications at all. If we choose to keep that particular filter as an exposed filter, we may want to incorporate something that causes only options with actual results to appear in the dropdown.

tadean:
One issue with the exposed Type filter on this particular view is that it displays all options for the type field. Many sites will not use every type of publication, leaving many of the type dropdown options with no results. We may want to eventually incorporate something that trims the dropdown list so that only options that actually have results are listed among the available types in the exposed form.

PR #2653, opened 2023-08-11
Closes #2572 Update Trellis Event Importer to use Trellis Event Search API

### LINKED ISSUE #2572: Update Trellis Event Importer to use Trellis Event Search API
## Motivation
I want to be able to search/filter for the event I made in Trellis so I can select it to show up on my website.

### Requirements
- Update Quickstart event search to use Enterprise Attributes from Trellis
- Master calendar vs Quickstart calendar feature gap review
- Provide detail page or modal for Master calendar?

Add filters on top of current ID paste box to be able to search and filter events coming in from the feed.

## Additional context
Jira ticket: https://jira.arizona.edu/browse/WEBTEAM-15700

Filters to be added to new content page.
image

tadean:
## Description
This PR updates the Trellis Event importer to use the new events search API provided by Trellis.

It provides this by means of a views bulk action form `AZEventTrellisViewsField extends BulkForm` that uses `views_remote_data` and an event subscriber to transparently fetch API results using `AZEventTrellisDataSubscriber` and add responses to the view results.

Updates to `TrellisHelper` have been made to allow for an API to search and fetch API responses.

A new view filter, `az_event_trellis_views_date_filter`, is provided that supports native Trellis date searching and translation of custom ranges into expected trellis values. (_Select or Other_ functionality)

Migration source is now provided by a new Source plugin `az_trellis_events_api`, which natively speaks to the Trellis API. This was necessary as the v2 Trellis API requires POST requests with a JSON body, and the native migrate url tools are primarily based around GET.

- [x] Initial skeleton of import structure with bulk form
- [x] Migration data fetcher
- [x] Implement all filters
- [x] Helper API additions
- [x] Migration updates

## Related issues
Closes #2572 

## How to test

- enable `az_event_trellis`
- enable `az_enterprise_attributes_import` (optionally)
- visit `/admin/trellis-event-importer`
- select filter options, click `search`
- verify that results are correct
- choose events to import by clicking the checkbox to the left of the event row
- click `save` at the bottom of the form
- verify that the event is imported

**Note: the current QA endpoint is not hooked up to the live data. Events that exist on production data are not necessarily expected to be in the results**

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [x] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joeparsons:
I like the approach you're taking with this proof of concept.  Seems pretty reasonable to me so far.  👍🏻

tadean:
Refactored exposed form alters to a dedicated filter plugin that uses configuration to map unique enterprise attribute values to trellis event search API parameters.

tadean:
Refactored to move API logic into `TrellisHelper`, added API caching features for migrate performance. Refactored subclass http fetcher into a dedicated migrate source plugin that understands how to make `POST` requests, since `v2` of the trellis events API requires this.

danahertzberg:
Check in on "auto-add by category" feature request

tadean:
Added Trellis Date Views filter to formalize preselected range options + Custom ranges.

tadean:
Further testing of the sporadic issue I'd seen where attributes seemed to be incorrect for some events has been isolated to an issue in the API responses, with example IDs here:
```
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"ids":"a3U8A000000N27CUAS"}' \
  https://api.qa.trellis.arizona.edu/ws/rest/getevents/v2/eventinfo/
 
curl --header "Content-Type: application/json" \
  --request POST \
  --data '{"ids":"a3U8A000000N27CUAS,a3U8A000000X62uUAC"}' \
  https://api.qa.trellis.arizona.edu/ws/rest/getevents/v2/eventinfo/ 
```
In comparing the response data, it appears that attributes in the responses about the events are being conflated with the attributes of other events retrieved in the same request, e.g. an event is returned with attributes it does not actually appear to have, if another event in the same request has them.

tadean:
Todo:

- [ ] Fix `az_event_trellis_migration_plugins_alter` to add ids for existing nodes only if the import migration is not being run
- [ ] Investigate removing css declaration from easepick instantiation
- [ ] See if claro fieldset markup around date fields can be removed

modules/custom/az_event/az_event_trellis/config/install/az_event_trellis.settings.yml
```diff
@@ -1 +1 @@
-api_hostname: 'prod-trellis-gateway.trellis.arizona.edu'
+api_hostname: 'api.qa.trellis.arizona.edu'
```
trackleft, line 1: Is this what we want?

tadean: TBD: to be reverted when the event search API is available on the production gateway.

tadean:
> Further testing of the sporadic issue I'd seen where attributes seemed to be incorrect for some events has been isolated to an issue in the API responses, with example IDs here:
> 
> ```
> curl --header "Content-Type: application/json" \
>   --request POST \
>   --data '{"ids":"a3U8A000000N27CUAS"}' \
>   https://api.qa.trellis.arizona.edu/ws/rest/getevents/v2/eventinfo/
>  
> curl --header "Content-Type: application/json" \
>   --request POST \
>   --data '{"ids":"a3U8A000000N27CUAS,a3U8A000000X62uUAC"}' \
>   https://api.qa.trellis.arizona.edu/ws/rest/getevents/v2/eventinfo/ 
> ```
> 
> In comparing the response data, it appears that attributes in the responses about the events are being conflated with the attributes of other events retrieved in the same request, e.g. an event is returned with attributes it does not actually appear to have, if another event in the same request has them.

This issue has been resolved by the Trellis team in the latest updates to the event search API :+1:

tadean:
`easepick`'s requirement of the css parameter in its configuration stem from using the shadow DOM; i.e css aggregated via the Drupal asset resolution process and included in the page appear to be out of scope for the elements in question.

I have refactored the build/javascript process for the form element to provide an aggregated, cached CSS file via `drupalSettings` comprised purely of the elements the easepick elements needs rather than aggregating them with the rest of the page.

tadean:
Merge new changes from main, update for 10.1 API:
- Views query condition in `az_event_trellis_views_attribute_filter` needed explicit `IN` operator
- Override `az_event_trellis_views_field`'s version of `BulkForm` validation function to not throw a form error if no action is selected.

tadean:
Updated to add syncing of Trellis events to `cron()` per discussion with the Trellis team.
Todo:
- [ ] Determine if this should be a `QueueWorker` instead of running directly in cron
- [ ] Determine configuration options

modules/custom/az_event/az_event_trellis/az_event_trellis.links.task.yml
```diff
@@ -6,7 +6,7 @@ az_event_trellis.az_event_trellis_settings_task:
 
 az_event.import_form_task:
   title: 'Trellis event importer'
-  route_name: az_event_trellis.az_event_trellis_import_form
+  route_name: view.az_event_trellis_import.page_1
```
joeparsons, line 9: Should we also remove the old form class?

modules/custom/az_event/az_event_trellis/az_event_trellis.module
```diff
@@ -55,27 +59,41 @@ function az_event_trellis_form_node_az_event_edit_form_alter(&$form, FormStateIn
...
+    $configuration['source']['trellis_ids'] = $event_api_ids;
+    $migration = \Drupal::service('plugin.manager.migration')->createInstance('az_trellis_events', $configuration);
+    $migration->getIdMap()->prepareUpdate();
+    // @todo potentially make this a QueueWorker process instead.
+    $executable = new MigrateExecutable($migration, new MigrateMessage());
+    try {
+      $executable->import();
     }
+    catch (MigrateException) {
+      \Drupal::logger('az_course')->notice('An error occurred while updating trellis events.');
```
joeparsons:
```suggestion
      \Drupal::logger('az_event_trellis')->notice('An error occurred while updating trellis events.');
```

modules/custom/az_event/az_event_trellis/az_event_trellis.module
```diff
@@ -55,27 +59,41 @@ function az_event_trellis_form_node_az_event_edit_form_alter(&$form, FormStateIn
...
+    }
+  }
+  catch (PluginException) {
+    \Drupal::logger('az_course')->notice('Could not instantiate az_trellis_events import.');
```
joeparsons:
```suggestion
    \Drupal::logger('az_event_trellis')->notice('Could not instantiate az_trellis_events import.');
```

modules/custom/az_event/az_event_trellis/az_event_trellis.info.yml
```diff
@@ -9,9 +9,11 @@ php: 8.0
 dependencies:
   - az_core
   - az_event
+  - drupal:views
   - migrate:migrate
   - migrate_plus:migrate_plus
   - migrate_tools:migrate_tools
   - migrate_queue_importer:migrate_queue_importer
```
joeparsons, line 16: Maybe we don't need the migrate_queue_importer dependency anymore since this module has it's own `hook_cron()` implementation now (?).  If not, we could also probably change the migration to a migration plugin instead of config entity.
```suggestion
```

tadean: I hadn't actually noticed until I was testing this out that there were migrate queue importer settings for this module - I think it might be better to use that, rather than hook cron implementation I'd previously committed. I'd like to discuss which would be preferable during PR review.

tadean: We discussed this in PR review and leaned towards using `migrate_queue_importer` to handle the `cron` job.

modules/custom/az_event/az_event_trellis/az_event_trellis.links.task.yml
tadean: I have removed the original form class.

tadean:
I have restructured the original default value handling from `hook_migration_plugins_alter` to the `az_trellis_events_api` source plugin class.

modules/custom/az_event/az_event_trellis/js/az_event_trellis_date.es6.js
```diff
@@ -0,0 +1,29 @@
...
+            zIndex: 10,
+            RangePlugin: {
+              elementEnd: end,
+            },
+            plugins: ['RangePlugin'],
+          });
+        });
+    },
+  };
+})(jQuery, Drupal, drupalSettings, once);
```
joeparsons, line 29: Nitpick:  does this really need jQuery?

```suggestion
((Drupal, drupalSettings, once) => {
  Drupal.behaviors.trellisDatePicker = {
    attach(context) {
      const elements = once('aztrellisdate', '.az-trellis-daterange', context)
        elements.forEach(function (element) {
          const begin = element;
          const id = element.dataset.az-trellis-daterange-end;
          const end = document.getElementById(id);
          // todo figure out how to make CSS declaration unnecessary.
          // eslint-disable-next-line no-unused-vars, no-undef, new-cap
          const picker = new easepick.create({
            element: begin,
            css: drupalSettings.trellisDatePicker.css,
            zIndex: 10,
            RangePlugin: {
              elementEnd: end,
            },
            plugins: ['RangePlugin'],
          });
        });
    },
  };
})(Drupal, drupalSettings, once);
```

tadean: Agreed. Thanks! I made a few eslint fixes and also removed the jquery library dependency declaration from the libraries.yml file.

joeparsons (dismissed):
I think this looks ready as of the latest changes.  Thanks @tadean!

danahertzberg:
Questions upon review:
1. I accidentally hit the `Import` button instead of the `Search` button at first and got the `No items selected.` error message which never goes away upon further searches and doesn't have an `X` on it for me to dismiss. Going to another content page and back to it makes it go away.
2. Can we add a `Reset` button?
3. Should the dropdowns be single or multi select?
4. Can the times be in 12-hour format instead of 24-hour?
5. Can the columns be click-to-sort (like by updated date, start date, etc.)?
6. We may want some intro text to help end-users.

I don't think these options hinder merging and can be follow up tasks.

tadean:
> Questions upon review:
> 
> 1. I accidentally hit the `Import` button instead of the `Search` button at first and got the `No items selected.` error message which never goes away upon further searches and doesn't have an `X` on it for me to dismiss. Going to another content page and back to it makes it go away.
> 2. Can we add a `Reset` button?
> 3. Should the dropdowns be single or multi select?
> 4. Can the times be in 12-hour format instead of 24-hour?
> 5. Can the columns be click-to-sort (like by updated date, start date, etc.)?
> 6. We may want some intro text to help end-users.
> 
> I don't think these options hinder merging and can be follow up tasks.

Thanks @danahertzberg ! Some answers about these:

1. This is because it is an AJAX-based view. Only the view itself refreshes on change, the rest of the page does not.
2. Yes, I think we can add a reset button.
3. The trellis event search API would need to support multiple search terms for a single parameter on the Trellis end. Currently I'm not sure if it does.
4. Oops, yes, this change can be made.
5. I think the sorting might need to take place on the Trellis event API results end potentially. It could potentially be implemented on the Quickstart end but may be slow in that case.

tadean:
- Updated to add Reset button for exposed filters
- Updated to change Start and Updated times to 12 hour date format

danahertzberg (approved):

tadean:
@danahertzberg correct, presently each occurrence of the event is a separate event in the Trellis API and needs to be imported separately

For support for truly recurring events that could be imported together, the external API provided by Trellis would need to offer information about the event akin to what is provided by the `RRULE` rucurrence rules in the [RFC 5545 iCalendar specification](https://datatracker.ietf.org/doc/html/rfc5545).

danahertzberg:
THANK YOU! Okay this is perfect information to have when we chat with Trellis and others. Thanks again @tadean! 🎉

joeparsons (approved):

trackleft (approved):

PR #2870, opened 2023-10-26
Closes #2868: Fix wrapping in CKEditor5 source edit mode.

### LINKED ISSUE #2868: Really long text or really long links break edit/source containers
## Problem/Motivation
2.8 RC1 Testing

### Describe the bug

When viewing a very long link in Source in CKE5, the link appears to carriage return only at "/", meaning that some links with long alphanumeric strings between /'s will cause source editor container to expand past intended boundary.

Similarly, when viewing a very long alphanumeric string of text in the WYSIWYG editor (non-source), the string can break the container, and extends past edge of page.  Carriage return behavior appears the same as the link carriage return behavior above. We discovered that "/" will carriage return, but "=" will not, however a ";" preceded by an "=" sign anywhere in the text string will break at the ";". 

### To Reproduce

**Really long link** - view Source view of CKE5 on this page: [https://28rc1-joe-azs-quickstart.pantheonsite.io/long-link](https://28rc1-joe-azs-quickstart.pantheonsite.io/long-link)

**Really long text** - visible in both live view and editor view of this page: [https://28rc1-joe-azs-quickstart.pantheonsite.io/long-text](https://28rc1-joe-azs-quickstart.pantheonsite.io/long-text)

## Proposed resolution

Use CKE4 behaviors as a model for resolution

### Expected behavior

CKE4 Seems to handle these scenarios well:

**Really long link** - text should break where needed, not necessarily where slashes exist:

Screenshot 2023-10-26 at 11 56 04 AM

**Really long text** - text may extend beyond editor window, but horizontal scroll allows for view

Screenshot 2023-10-26 at 11 59 55 AM

**NOTE** Really long text will actually break the live view and editor view in CKE4.  Really long text may be such an edge case that we don't need to overwork this problem.

Screenshot 2023-10-26 at 12 04 53 PM

Screenshot 2023-10-26 at 11 58 44 AM

(above screenshots from this page on live Quickstart site: [https://quickstart.arizona.edu/new-test](https://quickstart.arizona.edu/new-test))

 - OS: Mac OS 13.5.1
 - Browser - happening in Safari and FF

joeparsons:
## Description

Adds CSS to enforce overflow wrapping on the CKEditor5 when in source editing mode.

## Related issues

Closes #2868

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/themes/custom/az_barrio/css/ckeditor5.css
+++ b/themes/custom/az_barrio/css/ckeditor5.css
@@ -4,6 +4,14 @@
 }
 
 /**
+ * Fix source edit wrap issue
+ * See https://github.com/az-digital/az_quickstart/issues/2868.
+ */
+.ck-source-editing-area {
+  overflow-wrap: anywhere;
+}
+
+/**
  * Override the width of the table set by Bootstrap content styles.
  * See: https://github.com/ckeditor/ckeditor5/issues/3253.
  */


themes/custom/az_barrio/css/ckeditor5.css
```diff
@@ -3,6 +3,14 @@
   flex-wrap: wrap;
 }
 
+/**
+ * Fix source edit wrap issue
+ * See https://github.com/az-digital/az_quickstart/issues/2868.
+ */
+.ck-source-editing-area {
```
bberndt-uaz, line 10: Long lines are not wrapped in the standard editor window as well: could we select that <div> as well?

```suggestion
.ck.ck-content, .ck-source-editing-area {
```

bberndt-uaz:
It would be good to add the overflow-wrap rule to paragraph preview elements as well, since long lines can result in the Edit button being hidden underneath the sidebar. [Updating this rule should work](https://github.com/az-digital/az_quickstart/blob/0d24659431e8373a3676a150766edc54f89b32c5/modules/custom/az_paragraphs/css/az_paragraphs_preview.css#L23-L27).

image

themes/custom/az_barrio/css/ckeditor5.css
camikazegreen: This would be preferred.

joeparsons:
@trackleft and I talked about this and wonder if it wouldn't be better to _only_ fix the wrapping in the source code view since we don't really want to support/promote people using inappropriately long text that would be rendered on the frontend...

bberndt-uaz:
In Quickstart 2.7, the preview is not wrapped, so I agree with @joeparsons and @trackleft that we can leave it as is:
image

However, the standard editor in Quickstart 2.7 is wrapped, however (matching the idea from the commit suggestion above):
image

joeparsons:
> In Quickstart 2.7, the preview is not wrapped, so I agree with @joeparsons and @trackleft that we can leave it as is: image
> 
> However, the standard editor in Quickstart 2.7 is wrapped, however (matching the idea from the commit suggestion above): image

Thanks @bberndt-uaz.  I'll make the suggested change.

themes/custom/az_barrio/css/ckeditor5.css
```diff
@@ -3,6 +3,14 @@
   flex-wrap: wrap;
 }
 
+/**
+ * Fix source edit wrap issue
```
bberndt-uaz, line 7:
```suggestion
 * Fix editor wrap issue
```

bberndt-uaz (approved):

trackleft (approved):

tadean (approved):

PR #3111, opened 2024-01-19
Closes #3110: Add viewsreference module patch to fix field config schema issues

### LINKED ISSUE #3110: Disappearing field settings and paragraph options for views paragraphs
## Problem/Motivation
David Celaya-Gonzalez first noticed an issue where custom views being embedded and displayed by the View paragraph type would randomly disappear for anonymous visitors. Clearing the caches restores the embedded view for a time but it would eventually disappear again. Later, it was also reported to me that the "Include View Title", "Items per page", and "Offset results" options from the View paragraph had disappeared.

### Describe the bug
For the disappearing view issue, my custom views (not included with QS2) displayed via a views reference field sometimes disappear for anonymous users. No errors are being reported. Clearing the caches restores the view for a time.

For the disappeared view options, the "Limit results", "Offset results", and "Include View Title" options have been unset and cannot be re-enabled. Attempting to set them via the UI produces a success message but the changes are not persisted. Site builders are no longer able to use "Include View Title", "Items per page", and "Offset results" options on the View paragraph. Attempting to reset to default QS2 config reports a success, but the changes are not persisted.

### To Reproduce
I have not been able to reproduce the issue where the View paragraphs disappear. It currently appears to be random. We have not observed this issue for any authenticated user. They seem more likely to disappear with exposed filters enabled. Early on, I was able to explicitly break the referenced view by opening up the same page in two separate private browser windows. If I set the filter to a term that had no content assigned, it would break the view for the current user and the user in the other private window. Removing the empty filter values seemed to eliminate this condition.

For the disappearing options issue, changing any setting of field_az_view_reference (on the View paragraph type) unsets the "Limit results", "Offset results", and "Include View Title" and they cannot be re-enabled either via the UI or via resetting to default config. No errors are reported. Site builders can then no longer use the "Include View Title", "Items per page", or "Offset results" options of the View paragraph.

Screenshot 2024-01-12 at 11 37 33 AM

Screenshot 2024-01-12 at 11 38 28 AM

## Proposed resolution
There was discussion among the group on 1/19/24 that fixing the field settings problem might also correct the disappearing view problem. Since we are able to reliably reproduce the settings issue, we should probably start there.

### Expected behavior
Views, either those provided by QS2 or custom, should not disappear for anonymous users. If a fault occurs, it should be reported and visible in the Drupal log or elsewhere. Additionally, the settings for the View paragraph should persist when changed. We should likewise be able to revert to the default QS2 config for the View paragraph type.

## Additional context
The [config shipped with QS2 for this views reference field](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_paragraphs/az_paragraphs_view/config/install/field.field.paragraph.az_view_reference.field_az_view_reference.yml) seems to suggest that the types of the view options are all strings. However, the [schema shipping with the most recent beta release of the Views Reference Field module](https://git.drupalcode.org/project/viewsreference/-/blob/8.x-2.x/config/schema/viewsreference.schema.yml?ref_type=heads) seems to suggest that some of the options should be non-string types. Should these be aligned? I have seen at least once instance where the Configuration Manager sees these changes. See below:
Screenshot 2024-01-16 at 1 26 45 PM

I have momentarily been able to escape this problem by uninstalling both the Quickstart Paragraphs - View and Views Reference Field modules, deleting the corresponding QS2 view paragraph config items from the DB, and reinstalling the Quickstart Paragraphs - View module (which, in turn, reinstalls the Views Reference Field module and recreates a working config). As long as I don't change any of the settings on the View paragraph, site builders can continue to use the View paragraph and all of its intended options.

joeparsons:
Testing RTBC patch to see if it resolves disappearing view paragraph field settings phenomenon observed on various sites.

https://www.drupal.org/project/viewsreference/issues/3402036

## Related issues

Closes #3110 (hopefully)

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [x] Patch or minor level update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joeparsons:
Confirmed locally that this patch at least fixes the issue with view field settings not being saved correctly and view paragraph options not being available.

Since we still don't know the steps to reproduce the disappearing views paragraph issue, I haven't been able to test that yet.

trackleft (dismissed):
Approving to fix the disappearing field options

bberndt-uaz (approved):
In Probo I was able to modify the extra settings for the View paragraph's view reference field without any issues.

djcelaya (approved):
Looks like this works. I've check in a vanilla QS2 site and was able to change the view options. In a clone of one of my live sites, it also seems to do the trick. I'm able to adjust the view settings and also add in one of my custom views.

tadean (approved):

joeparsons:
Todo:

- [ ] Test whether a distribution update can be cleany applied on an existing site where this issue was already present
- [ ] If not, figure out a DB update for updating broken field settings

joeparsons:
Tested applying config updates after making these changes on a site where I had previously modified the field config settings in the broken state and didn't see any issues with updating config, etc.

I also re-saved and re-exported the viewsreference field settings after applying the patch.

trackleft (approved):

bberndt-uaz (approved):

PR #3221, opened 2024-02-28
Fixes #3220 Trellis Event Importer can import incorrect Event

### LINKED ISSUE #3220: Trellis Event Importer Form can import incorrect Event
## Problem/Motivation
Due to the multi-step build/rebuild process of Drupal forms and the remote API, it is possible in some circumstances for the Trellis Event Importer Form to import an event other than the event the user selected.

### Describe the bug
Given that some events may be cached and other events must be fetched from the remote API, presently the ordering of events in a given API request is not guaranteed if some of the events are in the cache and others must be fetched. This potential change in ordering means that sometimes an event import form can have its checkboxes change in value ordering during the rebuild process of the form.

### To Reproduce

- Search keyword `blue chip` on a brand new site
- Search keyword `ONE`
- Select the March and February workshop for import
- Note that different events are imported

## Proposed resolution
Force events to return in the same order regardless of if they are sourced from the cache or the Trellis API.

tadean:
## Description
The Trellis Event importer contains a caching layer for events to avoid having to fetch the events from the remote API if they have recently been seen. Other events are fetched as needed a single request because the nature of the Trellis event search API is to fetch multiple events in a single request.

Depending on status of the warmed cached this mix of cached and fetched events means that in some cases, the order of events in an API call is not constant. In most cases this does not matter, but if this happens during a views bulk form submit and rebuild of the form options, it can cause views to think a different row in the results were selected than the ones that actually were, which results in the wrong event being imported.

This PR introduces a defined sort order to the API calls so that events appear in the same order regardless of caching or remote API responses.

## Related issues
#3220 

## How to test

- enable az_event_trellis
- enable az_enterprise_attributes_import
- visit `/admin/trellis-event-importer`
- search keyword `blue chip`
- search keyword `ONE`
- select `March 2024 ONE Workshop` and `February 2024 ONE Workshop` for import
- click `Import`
- verify correct events are imported

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
+++ b/modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
@@ -110,6 +110,8 @@ final class TrellisHelper {
         $json = json_decode($json, TRUE);
         if ($json !== NULL) {
           $ids = $json['data']['Event_IDs'] ?? [];
+          // Ensure events are in Id order.
+          sort($ids);
           // @todo determine cache expiration.
           $expire = time() + 1800;
           // Cache search result.
@@ -135,6 +137,8 @@ final class TrellisHelper {
     $events = [];
     $fetch = [];
     $url = $this->getEventEndpoint();
+    // Remove any duplicate ids to mimic remote API.
+    $trellis_ids = array_unique($trellis_ids);
     // Grab events that are in cache.
     foreach ($trellis_ids as $trellis_id) {
       $cached = $this->getTrellisCache($trellis_id);
@@ -165,6 +169,10 @@ final class TrellisHelper {
       catch (GuzzleException $e) {
       }
     }
+    // Make sure events are in Id order regardless of cached/fetched.
+    usort($events, function ($a, $b) {
+      return strcmp($a['Id'], $b['Id']);
+    });
     return $events;
   }
 
--- a/modules/custom/az_event/az_event_trellis/src/Plugin/views/field/AZEventTrellisViewsField.php
+++ b/modules/custom/az_event/az_event_trellis/src/Plugin/views/field/AZEventTrellisViewsField.php
@@ -63,21 +63,16 @@ class AZEventTrellisViewsField extends BulkForm {
 
     foreach ($this->view->result as $row_index => $row) {
       $form[$this->options['id']][$row_index] = [
-        '#tree' => TRUE,
+        '#type' => 'checkbox',
+        // We are not able to determine a main "title" for each row, so we can
+        // only output a generic label.
+        '#title' => $this->t('Update this item'),
+        '#title_display' => 'invisible',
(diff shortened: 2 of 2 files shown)


camikazegreen (approved):
I was able to test with the same situation that uncovered this issue initially, and everything worked correctly.

modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
```diff
@@ -165,6 +169,10 @@ public function getEvents(array $trellis_ids) {
       catch (GuzzleException $e) {
       }
     }
+    // Make sure events are in Id order regardless of cached/fetched.
+    usort($events, function ($a, $b) {
+      return strcmp($a['Id'], $b['Id']);
+    });
```
trackleft, line 175:
```suggestion
    // Make sure events are in Id order regardless of cached/fetched.
    // phpcs:disable Security.BadFunctions.CallbackFunctions.WarnCallbackFunctions
    usort($events, function ($a, $b) {
      return strcmp($a['Id'], $b['Id']);
   });
   // phpcs:enable Security.BadFunctions.CallbackFunctions.WarnCallbackFunctions
```
See https://app.probo.ci/#/dashboard/build/84716530-c7ac-48ac-94f1-982609a22b4e/task/637c1d48e0400002

bberndt-uaz (approved):

modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
bberndt-uaz: Or maybe just // phpcs:ignore on line 173?

modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
```diff
@@ -165,6 +169,12 @@ public function getEvents(array $trellis_ids) {
       catch (GuzzleException $e) {
       }
     }
+    // Make sure events are in Id order regardless of cached/fetched.
+    // phpcs:disable Security.BadFunctions.CallbackFunctions.WarnCallbackFunctions
+    usort($events, function ($a, $b) {
+      return strcmp($a['Id'], $b['Id']);
+   });
```
trackleft, line 176:
```suggestion
    });
```

joeparsons (approved):

bberndt-uaz (approved):

PR #3250, opened 2024-03-08
Closes #3141: Customize Add CAS User(s) menu local action and optionally remove standard Add user action.

### LINKED ISSUE #3141: Make "+ Add CAS User" button primary at /admin/people
## Problem/Motivation

Our most common support ticket is User Admins having trouble adding new users. 

- Better documentation can help this, however it will be hard to avoid having users mistakenly use the "Add User" button rather than the "+ Add CAS User" button when adding users, as "+ Add User" is the wording that most clearly matches the intent of the User Admin.  The word CAS may actually subtly steer people away since it is not obvious the meaning of a CAS User.

### Describe the bug

Add User and Add CAS User are too similar visually, and the wording compels people to take the wrong action when adding site editors.

Screenshot 2024-01-31 at 1 50 06 PM

## Proposed resolution

- Make most common action (+ Add CAS User) visually dominant.  
- Potentially alter wording to help users understand meaning of buttons
- Update documentation at quickstart.arizona.edu to support proper action

joeparsons:
## Description

- Changes the text of the "Add CAS User(s)" button on the /admin/people page to be "Add user(s) with NetID"
- Updates the weight of the "Add user(s) with NetID" so it appears first
- Adds a new config setting to the AZ CAS module for optionally disabling (removing) the standard "Add user" button

## Related issues
Closes #3141 

## How to test

On [ProboCI site](https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-3250.probo.build/):
- Login with azadmin or azuseradmin account
- Navigate to /admin/people
- Verify that the "Add CAS User(s)" button has been renamed "Add user(s) with NetID"
- Verify that the "Add user(s) with NetID" button appears first (left of the "Add user" button)
- Navigate to /admin/config/az-quickstart/settings/az-cas
- Turn on "Disable 'Add user' button" setting, save
- Go back to /admin/people
- Verify that "Add user" button has been removed

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_cas/config/install/az_cas.settings.yml
+++ b/modules/custom/az_cas/config/install/az_cas.settings.yml
@@ -1,2 +1,3 @@
 disable_login_form: true
+disable_admin_add_user_button: false
 disable_password_recovery_link: true
--- a/modules/custom/az_cas/config/install/az_cas.settings.yml
+++ b/modules/custom/az_cas/config/install/az_cas.settings.yml
@@ -1,3 +1,3 @@
 disable_login_form: true
-disable_admin_add_user_button: false
+disable_admin_add_user_button: true
 disable_password_recovery_link: true
--- a/modules/custom/az_cas/az_cas.module
+++ b/modules/custom/az_cas/az_cas.module
@@ -115,3 +115,21 @@ function az_cas_module_implements_alter(&$implementations, $hook) {
     $implementations['az_cas'] = $group;
   }
 }
+
+/**
+ * Implements hook_menu_local_actions_alter().
+ *
+ * Customizes "Add CAS User(s)" menu local action and optionally disables the
+ * standard "Add user" menu local action on the admin user listing page.
+ */
+function az_cas_menu_local_actions_alter(&$local_actions) {
+  $az_cas_config = \Drupal::config('az_cas.settings');
+
+  $local_actions['cas.bulk_add_cas_users']['title'] = t('Add user(s) with NetID');
+  $local_actions['cas.bulk_add_cas_users']['weight'] = -1;
+
+  $disable_admin_add_user_button = $az_cas_config->get('disable_admin_add_user_button');
+  if ($disable_admin_add_user_button) {
+    unset($local_actions['user_admin_create']);
+  }
+}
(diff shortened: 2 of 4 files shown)


joshuasosa (dismissed):
Should "Add User" just be disabled by default on standard installs? I think more often than not a NetID user is the only kind of user that should be made. And it looks like "Disable login form" is set to true, which means a regular Drupal user account couldn't be used anyways.

joeparsons:
> Should "Add User" just be disabled by default on standard installs? I think more often than not a NetID user is the only kind of user that should be made. And it looks like "Disable login form" is set to true, which means a regular Drupal user account couldn't be used anyways.

Thanks for the suggestions @joshuasosa.  I agree that it would probably make sense to disable the standard "Add User" button by default.  I was also on the fence about whether to tie this to our existing "Disable login form" setting or add a new setting.  I'll bring these points up in our PR meeting(s) this week.

modules/custom/az_cas/config/install/az_cas.settings.yml
```diff
@@ -1,2 +1,3 @@
 disable_login_form: true
+disable_admin_add_user_button: false
```
joeparsons, line 2: Questions to consider:
- Is this additional config setting needed or would it be better to tie this to our existing `disable_login_form` setting?
- If we want to keep the extra setting, should it be turned on by default (disable the standard "Add User" button by default)?
```suggestion
disable_admin_add_user_button: true
```

sanjanans (approved):

trackleft (approved):

danahertzberg (approved):

PR #3279, opened 2024-03-19
Closes #2997 Add Demo Content as a Starting Point for Better Exposed Filter Finder Functionality

trackleft:
## Description

This pull request is here to allow independent QA and acceptance of demo content and to demonstrate how better exposed filters currently works without customize-ations from Quickstart.

This pull request addresses issue #2997 by creating demo content for the finder functionality within the Arizona Quickstart project. Key features added include:
- A new "Finders" top-level menu item.
- 1 new "Finder" view with 5 new view displays (one for each content type enabled by default and another combined type view)
- 6 new menu items, a parent and five children
- 5 new views paragraphs (These reference a view that is un-selectable via the UI `az_finder`)
- 5 new flexible pages (These reference the view paragraphs)
- 5 new blocks in the `modules/custum/az_demo/config/optional` folder (one for each exposed form block)
- Each Finder view is configured to showcase how each content type is different, with possible exposed filters that will appear along-side taxonomy and search.
- Updated some of the JSON to comply with Quickstart Coding standards.

## Related issues
- Closes #2997
- Blocks #2998 
- Blocks #3281 

### General notes:
- Familiarize yourself with how each view display is set up within the az_finder view.
- This demo content populates the views reference paragraphs with the `az_finder` view, which is not selectable via the UI.

### News Finder
1. Navigate to the top-level menu item "Finders" and select the "News Finder" option, or directly go to `/finder/news`.
2. Interact with the Better Exposed Filter block to test its functionality:
   - Use the search field to filter content by titles.
   - Use the checkboxes for nested taxonomy to filter content by taxonomy terms.
   - Use the byline search filter
3. Verify that the content on the page accurately reflects the filters applied.
4. While logged in as an admin, hovering over the exposed filter block will allow you to edit the view, use this contextual link to view the exposed filter settings in both the Advanced and Filter tabs
image
image
image

### Event Finder
1. Navigate to the top-level menu item "Finders" and select the "Event Finder" option, or directly go to `/finder/events`.
2. Interact with the Better Exposed Filter block to test its functionality:
   - Use the search field to filter content by titles.
   - Use the date picker.
   - Use the checkboxes for nested taxonomy to filter content by taxonomy terms.
3. Verify that the content on the page accurately reflects the filters applied.

### People Finder
1. Navigate to the top-level menu item "Finders" and select the "People Finder" option, or directly go to `/finder/people`.
2. Interact with the Better Exposed Filter block to test its functionality:
   - Use the search field to filter content by name.
[...]

PR #2998, opened 2023-12-06
Closes #2997: Add better exposed filter plugin that transforms a taxonomy vocabulary into a nested list of filters for a view.

trackleft:
### Release Notes Snippet (Change Notice)
The AZ Finder functionality is strictly opt-in (Also need to write documentation on how to use/enable.)

## Description
This PR Adds
- In optional config:
  - Block config for the exposed filter block set in az_barrio
  - A view with finder functionality
- In demo content
  - Menu entry
  - Paragraph with the view set
  - A Page node

## Related issues
- Blocks https://github.com/az-digital/az_quickstart/pull/3306
- Closes #2997 
- Closes #2275 
- Blocked by #3289 
- Related to #3298 
- Related to #3294 

## How to test

1. Go to `/pages/page-list-finder`
2. Modify page categories taxonomy list structure and clear cache
3. View style changes to filter
4. Test functionality of the search box and filters
   - _Note: Search box only searches the titles_
5. Test functionality of clear filters button (should appear automatically)
6. Test functionality of filters via keyboard navigation
   - Also should be able to jump to filter/search before the content via keyboard

## Remaining to do items

- [x] li > ul > li bold if it does not have a nested ul
- [x] Update the clear filters button to count search field
- [x] Update search functionality so "return" keyboard button doesn't clear the filters
- [x] Remove overridden margin-bottom: 1rem for .az-bef-vertical
- [x] Get DRC feedback - especially on non-clickable terms
- ~Hierarchical select module for not allowing parent terms to be selectable - Chris~
  - ~https://www.drupal.org/project/hierarchical_select~
  - ~https://www.drupal.org/project/shs~
  - ~This would be useful in this case, but not useful in future cases~
  - _Don't add this module/functionality to change the back-end term selection interface_
- [x] "Remove" and "Add" icon name is being read in the screenreader which is confusing
  - [x] "Expanded" on the icon would make more sense than "expand_less"
  - [x] Have order of levels always be read after the title - Chris
- [x] Make collapsible headings as H3/H4 around button
- [x] Return nested functionality so all parents are collapsible and not clickable
- [ ] Ensure keyboard navigation still works with enter key - Need help
   - [x] Send to Jeff - Is the spacebar-only for checkboxes expected? Not the enter key? Opposite for expand buttons? - Dana
      - [x] Need to have both keys (return and spacebar) available to click on expand/collapse
   - [x] Send to Jeff - Is filter at the bottom of results really okay?
[...]

PR #3281, opened 2024-03-19
Closes #3280 Removing Horizontal Negative Margins from Exposed Forms

### LINKED ISSUE #3280: Improve Form Responsiveness and Alignment by Removing Horizontal Negative Margins from Exposed Forms
## Motivation
The current `.form-row` class adds negative margins which can cause layout issues, particularly when the form is within a container that has its own padding or margins. This can lead to elements overflowing their container or not aligning correctly on the page.

The negative margins on the `.form-row` class can cause content to be misaligned or to overflow its container, especially on smaller screens or when placed in a grid layout. This results in a less than optimal user experience.

To resolve this, I propose adding a template to az_barrio theme with the same code as the template in bootstrap barrio, but removing the `.form-row` class, which applies negative margins, and replacing it with `.d-flex` and `.flex-wrap` classes. This will not only remove unwanted negative margins but also maintain the flexibility of the layout on different screen sizes.

An alternative solution would be to override the negative margins with custom CSS. However, this approach can lead to maintenance issues and lacks the simplicity and responsiveness provided by Bootstrap's native classes.

- **Anonymous user:** Will see improved alignment and spacing on forms, especially on mobile devices.
- **Authenticated user:** Will benefit from better form layouts without overflowing elements.
- **Content editor:** No impact on editorial capabilities, but improved UI for forms.
- **Content administrator:** Easier to manage the appearance of forms within the UI.
- **Administrator:** Simpler theme maintenance and less need for custom overrides.

## Additional context

**Before Suggested Change: (in bootstrap barrio theme) file: [templates/views/views-exposed-form.html.twig](https://git.drupalcode.org/project/bootstrap_barrio/-/blob/5.x/templates/views/views-exposed-form.html.twig?ref_type=heads)**

trackleft:
Remove unnecessary horizontal negative margin from exposed forms.

## Description
This change improves the visual layout of exposed forms by removing the horizontal negative margins previously applied by the `form-row` class. The replacement with `d-flex flex-wrap` enhances form responsiveness across various device widths without affecting the functionality.

## Related issues
- Closes #3280
- Blocks #2998
- Blocked by #3279 (blocked because it is difficult to test without demo content.)

- **Minor release changes**
   
## Checklist

### DIFF
--- a/themes/custom/az_barrio/templates/views/views-exposed-form.html.twig
+++ b/themes/custom/az_barrio/templates/views/views-exposed-form.html.twig
@@ -1,0 +1,21 @@
+{#
+/**
+ * @file
+ * Theme override for a views exposed form.
+ *
+ * Available variables:
+ * - form: A render element representing the form.
+ *
+ * @see template_preprocess_views_exposed_form()
+ */
+#}
+{% if q is not empty %}
+  {#
+    This ensures that, if clean URLs are off, the 'q' is added first,
+    as a hidden form element, so that it shows up first in the POST URL.
+  #}
+{{ q }}
+{% endif %}
+<div class="d-flex flex-wrap">
+  {{ form }}
+</div>


danahertzberg (dismissed):

themes/custom/az_barrio/templates/views/views-exposed-form.html.twig
```diff
@@ -0,0 +1,21 @@
...
+ */
+#}
+{% if q is not empty %}
+  {#
+    This ensures that, if clean URLs are off, the 'q' is added first,
+    as a hidden form element, so that it shows up first in the POST URL.
+  #}
+{{ q }}
+{% endif %}
+<div class="d-flex flex-wrap">
```
trackleft, line 19:
```suggestion
<div class="form-row mx-0">
```

tadean (approved):

joeparsons:
Bootstrap Barrio version of the template: https://git.drupalcode.org/project/bootstrap_barrio/-/blob/5.1.10/templates/views/views-exposed-form.html.twig?ref_type=tags

tadean (approved):

danahertzberg (approved):

joeparsons (approved):

tadean:
Seems to cause some impacts on the calendar filter rendering that might be unintended.

joeparsons:
After taking a closer look at the how this affects various views, etc, we decided to hold off on merging this for 2.11.0-alpha1.  More testing is needed.

joeparsons:
We need to determine if this is still needed as of 2.11.0 and later.

joshuasosa:
Is this issue and PR still needed? The `views-exposed-form.html.twig` template was changed with the upgrade to bootstrap_barrio 5.5 and further adjusted already in az_barrio.

- bootstrap_barrio for Bootstrap 5: https://git.drupalcode.org/project/bootstrap_barrio/-/blob/5.5.x/templates/views/views-exposed-form.html.twig
- az_barrio for Quickstart 3: https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/views/views-exposed-form.html.twig

If anything, I think this issue and PR would be relevant for 2.x only.

trackleft:
Let's close it. Thanks

Closed Won't fix

PR #3481, opened 2024-06-12
Remove `az_google_tag`'s dependency on `ctools` and remove `ctools` from `az_quickstart`.

trackleft:
### Release notes:

Remove dependency on ctools, Removing Chaos Tools
```
~~**IMPORTANT**: Chaos Tools (`ctools`) dependency  removed from Quickstart 2.~~

Quickstart no longer has a runtime dependency on the ctools module, and the module has been slated for removal. This might cause problems if other modules implicitly use ctools functionality without depending on it.

Note that for backwards compatibility, the composer definition still depends on the ctools project, but it's no longer installed on new projects and existing projects may uninstall it if they are not using it.

The composer definition is there so that existing sites do not automatically have the code removed underneath them and continue to work. This will be removed in the future in a new major version. It is recommended to either add an explicit dependency on ctools if you plan to keep using it or uninstall it.

To fully remove ctools from the project, add a replace section to the root composer.json:

 "replace": {
        "drupal/ctools": "*"
    },
Note that this will also skip it in case another module depends on it and the definition will need to be removed then.

Before uninstalling ctools, verify that no configuration dependencies remain on views or other config entities. Views should have that dependency removed automatically when resaving them.
```

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joeparsons:
```
# composer why drupal/ctools
az-digital/az_quickstart dev-dependabot/composer/drupal/ctools-4.1.0-remove-ctools requires drupal/ctools (*) 
drupal/pathauto          1.12.0                                                    requires drupal/ctools (*) 
```

joeparsons:
FWIW, I think we could just make this change directly to the mian branch (change the target branch for this PR) as an alternative to the dependabot PR instead of targeting the dependabot PR branch.

joeparsons (approved):

trackleft:
```
www-data@8ded398cc170:/app$ composer why drupal/ctools
az-digital/az_quickstart  2.10.4  requires  drupal/ctools (3.14.0)  
drupal/pathauto           1.12.0  requires  drupal/ctools (*)       
```

tadean (approved):

trackleft:
Add to https://github.com/az-digital/az_quickstart/blob/main/az_quickstart.install#L80-L83

joeparsons (approved):

tadean (approved):

PR #3589, opened 2024-08-01
Close #3588 Allow newly enabled module config to be overridden.

### LINKED ISSUE #3588: Settings in az_mail/config/quickstart no longer apply.
## Problem/Motivation
The settings in the `az_mail` module are not being applied when the module is enabled.

### Describe the bug

` [notice] Disallowing override of customized configuration smtp.settings.`

In modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php

The `canOverride` method incorrectly marks configuration that does not have a snapshot yet as `FALSE`.

https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php#L274-L288

### To Reproduce

1. Enable `az_mail` module
2. Go to `/admin/config/system/smtp`
3. See settings are not applied.

## Proposed resolution

Return `TRUE` in `canOverride` in https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php

Something like.
```
    // Guard against missing items.
    $snap = (!empty($snap)) ? $snap : [];
    $active = (!empty($active)) ? $active : [];

    // Not relevant for user roles. Permissions created dynamically.
    if (strpos($name, 'user.role.') !== 0) {
      // Prune cache_metadata if present, to not consider it for diffs.
      $active = $this->trimNestedKey($active, 'cache_metadata');
      $snap = $this->trimNestedKey($snap, 'cache_metadata');
      // If $snap is empty, then we can't compare it to $active.
      // In this case, we can assume that the configuration is not customized
      // because it is not present in the snapshot.
      // This means that the module (config provider) is not yet installed.
      if (empty($snap)) {
        return TRUE;
      }
      // Diff active config and snapshot of module to check for customization.
      $diff = $differ->diff($active, $snap);
      // Overrides only allowed if no changes in diff.
      if (!$this->diffIsEmpty($diff) && !empty($active)) {
        return FALSE;
      }
```

trackleft:
## Description
Not sure if this is correct, but it does allow me to override the smtp.settings via az_mail.

## Related issues
Close #3588 

## How to test
1. Enable az_mail module
2. Go to /admin/config/system/smtp
3. See settings are applied.

Additionally, check the build log for any anomalies.

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
+++ b/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
@@ -262,7 +262,6 @@ class QuickstartConfigProvider extends ConfigProviderBase {
     $lister = \Drupal::service('config_update.config_list');
     /** @var \Drupal\config_update\ConfigDiffer $differ */
     $differ = \Drupal::service('config_update.config_diff');
-
     // Read active config value for name.
     $active = $this->getActiveStorages()->read($name);
     // Find out which module owns the configuration and load the snapshot value.
@@ -274,11 +273,19 @@ class QuickstartConfigProvider extends ConfigProviderBase {
     // Guard against missing items.
     $snap = (!empty($snap)) ? $snap : [];
     $active = (!empty($active)) ? $active : [];
+
     // Not relevant for user roles. Permissions created dynamically.
     if (strpos($name, 'user.role.') !== 0) {
       // Prune cache_metadata if present, to not consider it for diffs.
       $active = $this->trimNestedKey($active, 'cache_metadata');
       $snap = $this->trimNestedKey($snap, 'cache_metadata');
+      // If $snap is empty, then we can't compare it to $active.
+      // In this case, we can assume that the configuration is not customized
+      // because it is not present in the snapshot.
+      // This means that the module (config provider) is not yet installed.
+      if (empty($snap)) {
+        return TRUE;
+      }
       // Diff active config and snapshot of module to check for customization.
       $diff = $differ->diff($active, $snap);
       // Overrides only allowed if no changes in diff.
--- a/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
+++ b/modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
@@ -269,6 +269,13 @@ class QuickstartConfigProvider extends ConfigProviderBase {
     if (!empty($owner[1])) {
       $snapshot_storage = $this->getConfigSnapshotStorage(ConfigSyncSnapshotterInterface::CONFIG_SNAPSHOT_SET, $owner[0], $owner[1]);
       $snap = $snapshot_storage->read($name);
+      // StorageInterface::read() returns FALSE if the config does not exist.
+      // In this case, we can assume that the configuration is not customized
+      // because it is not present in the snapshot, likely because the module
+      // is newly installed.
+      if ($snap === FALSE) {
+        return TRUE;
+      }
     }
     // Guard against missing items.
     $snap = (!empty($snap)) ? $snap : [];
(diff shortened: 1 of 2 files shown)


modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php
```diff
@@ -274,11 +273,19 @@ protected function canOverride($name) {
     // Guard against missing items.
     $snap = (!empty($snap)) ? $snap : [];
     $active = (!empty($active)) ? $active : [];
+
```
trackleft, line 276:
```suggestion
```

joeparsons:
Tested locally with lando/drush.

Before:
```
# lando drush en -y az_mail
The following module(s) will be enabled: az_mail, smtp

 // Do you want to continue?: yes.                                                                                      

 [notice] Disallowing override of customized configuration smtp.settings.
 [success] Successfully enabled: az_mail, smtp
```

After:
```
# lando drush en -y az_mail
The following module(s) will be enabled: az_mail, smtp

 // Do you want to continue?: yes.                                                                                      

 [success] Successfully enabled: az_mail, smtp
```
👍🏻

joeparsons (approved):
Seemed to work in my local testing.

Was hoping it might also resolve these problems but I think they're caused by something else:
```
 [notice] Could not validate dependencies of override antibot.settings.
 [notice] Could not validate dependencies of override seckit.settings.
```

joeparsons:
Confirmed this issue also affects 2.10.x.

tadean:
It won't let me make a commit suggestion because it's not one of the lines involved in the PR, but maybe something like this below line 271 instead?
```
      // StorageInterface::read() returns FALSE if the config does not exist.
      // In this case, we can assume that the configuration is not customized
      // because it is not present in the snapshot, likely because the module
      // is newly installed.
      if ($snap === FALSE) {
        return TRUE;
      }
 ```
 StorageInterface seems to use `FALSE` to distinguish between non-existent config and empty config.

File: modules/custom/az_core/src/Plugin/ConfigProvider/QuickstartConfigProvider.php [hunk 2]
```diff
@@ -262,14 +262,20 @@ protected function canOverride($name) {
...
     if (!empty($owner[1])) {
       $snapshot_storage = $this->getConfigSnapshotStorage(ConfigSyncSnapshotterInterface::CONFIG_SNAPSHOT_SET, $owner[0], $owner[1]);
       $snap = $snapshot_storage->read($name);
+      // StorageInterface::read() returns FALSE if the config does not exist.
+      // In this case, we can assume that the configuration is not customized
+      // because it is not present in the snapshot, likely because the module
+      // is newly installed.
+      if ($snap === FALSE) {
+        return TRUE;
+      }
```
trackleft, line 278: I've heard of an `isNew` property on config objects before in this ticket: https://www.drupal.org/project/config_sync/issues/2996039

I wonder if that could be used here.

joeparsons: https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Config%21StorableConfigBase.php/function/StorableConfigBase%3A%3AisNew/11.x

tadean (approved):

joeparsons (approved):

PR #3595, opened 2024-08-01
Close #3588 Add PHPUnit Test for ConfigProvider on 2.9.x

### LINKED ISSUE #3588: Settings in az_mail/config/quickstart no longer apply.

trackleft:
## Description

Adding PHPUnit test without fix in order to test whether #3588 exists on 2.9.x

## Related issues

Relates to #3589 

## How to test
Use PHPUnit or look in the Probo logs.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/tests/src/Functional/ConfigOverrideTest.php
+++ b/modules/custom/az_core/tests/src/Functional/ConfigOverrideTest.php
@@ -51,4 +51,32 @@ class ConfigOverrideTest extends BrowserTestBase {
     $this->assertEquals('shibboleth.arizona.edu', $hostname);
   }
 
+  /**
+   * Tests a config override was applied successfully after module install.
+   */
+  public function testConfigOverrideAfterInstall() {
+    // Install the az_mail module.
+    $this->container->get('module_installer')->install(['az_mail']);
+    $config = $this->config('smtp.settings');
+    $this->assertEquals(TRUE, $config->get('smtp_on'));
+    $this->assertEquals('email-smtp.us-west-2.amazonaws.com', $config->get('smtp_host'));
+    $this->assertEquals('', $config->get('smtp_hostbackup'));
+    $this->assertEquals('2587', $config->get('smtp_port'));
+    $this->assertEquals('tls', $config->get('smtp_protocol'));
+    $this->assertEquals(TRUE, $config->get('smtp_autotls'));
+    $this->assertEquals(30, $config->get('smtp_timeout'));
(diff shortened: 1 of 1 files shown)


trackleft:
We'll backport #3589 instead

PR #3786, opened 2024-10-17
Closes #2929: Strip links from summary field(s).

### LINKED ISSUE #2929: Web page addresses and email addresses turning into links automatically in our summary field, breaks our displays.
image

image

danahertzberg:
## Description
This fix changes the plain text format so it doesn't automatically convert links. All of the news displays wrap the entirety of the preview content (summary included) in a link. Text within the summary field should ultimately be true plain text so we don't get unexpected results with links.

### Release notes

If this change requires release notes: provide a summary of changes, how to use this change, and any related links. This content will be pasted in the [release notes](https://github.com/az-digital/az_quickstart/releases). Use markdown format to ensure proper pasting of information.

Make sure to add the `release notes` label to this PR.

```
### Change to summary fields
News, Events, and Pages with a populated summary field values will notice a change if the content within the field has a link or email. These will no longer be automatically linked and will rendered as plain text.
```

## Related issues
Closes https://github.com/az-digital/az_quickstart/issues/2929

## How to test
1. Add a News item with a link or email in the summary.
2. View item on `/news`
3. Ensure links and emails do not break overall link to news item.

_Repeat for Page (Summary field) and Events (Summary field)_

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

- [x] My change requires release notes.


danahertzberg:
Thanks @natebyerley! Was the hyperlink functionality used much? If so, the link to the page was broken but was that just not noticed?

natebyerley:
No, I can't say that many people use this Summary field, let alone the hyperlink feature. Just seemed like a nice perk.  But this is really a 'Summary' field, not a Call to Action field, so I'm not sure there's a rationale for linking.  And if we're going to link to urls, emails, why not phone numbers?  I think plain text is acceptable and probably appropriate.

tadean:
Is the rationale for this change the news previews only? If so, could this change potentially be handled in a way that is specific to that field? I'm a little worried that this seems like a large change to a Drupal core text format.

joeparsons:
> Is the rationale for this change the news previews only? If so, could this change potentially be handled in a way that is specific to that display? I'm a little worried that this might be a large change to a Drupal core text format.

I agree that it'd be better to handle this another way rather than modifying/overriding the [core-provided text format](https://git.drupalcode.org/project/drupal/-/blob/10.3.6/core/modules/filter/config/install/filter.format.plain_text.yml).

We could consider adding a new text format that and using on the news preview display.

danahertzberg:
I'm happy to add a new text format to use and not override the core format. It will need to be used on News, Events, and Pages for their summary fields.

tadean:
We could potentially consider using the same text format as normal, and just in the template for these summary fields, using the twig `striptags` filter (if we don't want to create a new unrelated text format)

danahertzberg:
I think I actually prefer the custom text format. We want some HTML allowed (like line breaks), just not the link functionality in specific. I would also think a custom text format may be easier to track down in the future if we need to.

danahertzberg:
> For the custom text format route, the change might need an associated database update to switch the text format for existing content.

@tadean Would `striptags` not require a db update and still change the output? If the text format was changed, would the output only change if the content was edited?

tadean:
@danahertzberg of the three approaches discussed in the thread, I think there are the following considerations:

- The field formatter approach suggested by @joshuasosa where a field formatter is used that strips out the tags
  - this would become active after distribution updates are applied (which would change the field formatter for the view mode from the `Default` to the new formatter that removes tags
  - This is probably the most Drupal-standards appropriate way of changing the output without touching the text format
  - No database update would be needed to update existing content
  - Applying the distribution updates is needed
- Stripping tags via `striptags` in either theme preprocessing or a template
  - This change would take place after just a cache clear
  - A disadvantage is there is no way to trivially turn this functionality off if the user does not want it
  - No database update needed
  - Applying distribution updates is not needed (but, as said there is no simple way to turn this approach off)
- Creating a new text format
  - Needs distribution updates to be applied (which changes the allowed text format on the summary fields to be the new text format)
  - Content must be updated, either in bulk via a database update, or by saving the content again manually. If they save it manually, they will need to manually select the text format for the summary, because the text formats allowed will have changed from what is already in the field. Content that isn't updated will still exhibit the issue.
  - content editors need to be given permission on the new format (`use text format az_plain_text`)
  - Related migrations might need to be updated to [no longer specifically refer to the old core text format](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_migration/migrations/az_node_news.yml#L161).
  
  Here is the editor experience with the new text format when resaving existing content without a database update having happened, for example:
  
Screenshot from 2024-10-21 09-36-00

modules/custom/az_core/config/install/filter.format.az_plain_text.yml
```diff
@@ -0,0 +1,13 @@
+langcode: en
+status: true
+dependencies: {  }
+name: 'AZ Plain Text'
+format: az_plain_text
+weight: 0
+filters:
```
tadean, line 7: Is it intended that this format removes the `filter_html_escape` filter normally present in `plain_text`?

It seems like the only filter we'd want to remove would be `filter_url`

trackleft:
This needs a database update.

tadean:
> This needs a database update.

FWIW, I think @joshuasosa 's suggestion has merit that this could be accomplished via a field formatter. A database update would not be necessary in that case.

danahertzberg:
Thanks for the breakdown @tadean. I'd like to talk about it more on Wednesday if that's alright. I'm happy without a DB update for convenience. Just want to ensure that any changes are properly accounted for. And if it's with release notes, that's good too. I just wouldn't want to override the functionality and wipe away anyone's customizations unless that's preferred.

modules/custom/az_core/config/install/filter.format.az_plain_text.yml
danahertzberg: I tried to use the minimum filters necessary not based on what was used with the plain text format. What does `filter_html_escape` do?

tadean: `filter_html_escape` is the filter that encodes raw encountered HTML into HTML entities. For example, if that filter is enabled, if the user enters `<script>alert('run some javascript');</script>` into the text area, the `filter_html_escape` filter will cause that to show as plain text. Without that filter enabled, the HTML will be treated as HTML by the browser and the javascript would run.

Most text formats enable either `filter_html` (if they allow some tags), or `filter_html_escape` (if they allow no tags). Text formats do not enable either of the two filters allow raw HTML.

In the case of the Drupal core `plain_text` format, `filter_html_escape` filter is run before `filter_autop` and `filter_url` to avoid eating the html produced by those filters.

modules/custom/az_core/config/install/filter.format.az_plain_text.yml
```diff
@@ -0,0 +1,13 @@
...
+  filter_autop:
+    id: filter_autop
+    provider: filter
+    status: true
+    weight: 0
+    settings: {  }
```
tadean, line 13:
```suggestion
    settings: {  }
  filter_html_escape:
    id: filter_html_escape
    provider: filter
    status: true
    weight: -10
    settings: {  }
```

danahertzberg:
Decision is to use a field formatter ([example](https://gist.github.com/tadean/00ed7de103814a6de5dddc85fd876e79)) and Troy will help guide as available.

tadean:
Updated to swap displays that use clickable card classes to use a summary formatter that allows only `<p><br>` (which are the HTML tags that can be present in the plain text format aside from links)

- [ ] Needs investigation to determine if other displays need this added
- [ ] Determine naming of formatter, if desirable

tadean:
**Note: this causes visual changes.** Some of these are anticipated, some are consequences of currently broken markup becoming valid.

- Previous automatic links are no longer styled as links because they are not links
- Bold text no longer bold on text in cards where nested anchor links were causing broken markup (Consequence of fixing the broken markup. See, text in before/after: `Matt Osman`)
- Changes to spacing on cards where markup was currently broken by improperly nested links
**Before:**
original


**After:**
after2

modules/custom/az_event/config/install/core.entity_view_display.node.az_event.az_row.yml
```diff
@@ -18,10 +18,10 @@ dependencies:
     - field.field.node.az_event.field_az_trellis_id
     - node.type.az_event
   module:
+    - az_core
     - date_ap_style
     - field_group
     - smart_title
-    - text
```
danahertzberg, line 24: Is this needed in the other files? It seems to work fine in them. This came up in config export from my local dev.

danahertzberg:
Is it possible to modify the help text?
```
No HTML tags allowed.
Lines and paragraphs break automatically.
Web page addresses and email addresses turn into links automatically.
```

modules/custom/az_event/config/install/core.entity_view_display.node.az_event.az_row.yml
tadean: Oops, yes, thanks! - it seems like this is probably necessary in the other files. I did not rexport them, just changed the text formatter.

tadean:
> Is it possible to modify the help text? This is what it is currently.
> 
> ```
> No HTML tags allowed.
> Lines and paragraphs break automatically.
> Web page addresses and email addresses turn into links automatically.
> ```

Possibly, but it may be complex to do so - the removal of the tags is a property of the formatter now, not the text format, and the text format is what generates this message. There are displays where summaries are used where links will not be removed, because they do not use the formatter.

danahertzberg:
> > Is it possible to modify the help text? This is what it is currently.
> > ```
> > No HTML tags allowed.
> > Lines and paragraphs break automatically.
> > Web page addresses and email addresses turn into links automatically.
> > ```
> 
> Possibly, but it may be complex to do so - the removal of the tags is a property of the formatter now, not the text format, and the text format is what generates this message. There are displays where summaries are used where links will not be removed, because they do not use the formatter.

Ah okay. Thank you for the information. I will make this a separate ticket and low priority. I expect we will not get any complaints.

ewlyman:
Requesting feedback:

This works in testing as designed --but there's a "BUT".
There might be cases when a content editor intentionally includes a link in the summary, expecting it to be used from the summary. This becomes a problem if the editor does not include the same link somewhere in the article. Please see screenshots from the probo build, which provide an example of this.

Screenshot 2024-10-25 at 10 54 54 AM
Screenshot 2024-10-25 at 11 01 39 AM

danahertzberg:
> Requesting feedback:
> 
> This works in testing as designed --but there's a "BUT". There might be cases when a content editor intentionally includes a link in the summary, expecting it to be used from the summary. This becomes a problem if the editor does not include the same link somewhere in the article. Please see screenshots from the probo build, which provide an example of this.

This is a good callout, thank you. I think that is expected and we should address with documentation. I can work on updating that on the Quickstart site.

ewlyman:
> > Requesting feedback:
> > This works in testing as designed --but there's a "BUT". There might be cases when a content editor intentionally includes a link in the summary, expecting it to be used from the summary. This becomes a problem if the editor does not include the same link somewhere in the article. Please see screenshots from the probo build, which provide an example of this.
> 
> This is a good callout, thank you. I think that is expected and we should address with documentation. I can work on updating that on the Quickstart site.

In light of @tadean's comment that the summary is infrequently used, I agree my catch may only cover a few edge cases.
If we can update the documentation, I think that would take care of it.

ewlyman (approved):
Works as described and update to documentation fixes issue I noted earlier.

joeparsons (approved):

PR #3933, opened 2024-12-09
Closes #3821: Make the # link stay at the current page

### LINKED ISSUE #3821: Using # in card link field now goes to the homepage instead of the same page
## Problem/Motivation
The `#` link in a link field no longer goes to the same page I am on.

### Describe the bug
When I add a `#` link in a link field, it goes to the site homepage.

### To Reproduce

1. Add a card to a page
2. Add `#` to the link URL field
3. Save
4. Observe the link redirecting the user to the site homepage

## Proposed resolution
Description of the proposed solution, the rationale behind it, and workarounds for people who cannot use the patch.

##  Screenshots

BaoNguyen09:
## Description

Fix: Make the # link stays at the current page

## Related issues

Closes #3821

## How to test

1. Add a card to a page
2. Add # to the link URL field
3. Save
4. Observe the link redirecting the user to the current page

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
```diff
@@ -125,13 +125,20 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
       $link_url = '';
+      $current_path = \Drupal::service('path.current')->getPath();
       if ($item->link_title || $item->link_uri) {
         if (str_starts_with($item->link_uri ?? '', '/' . PublicStream::basePath())) {
           // Link to public file: use fromUri() to get the URL.
           $link_url = Url::fromUri(urldecode('base:' . $item->link_uri));
         }
         else {
-          $link_url = $this->pathValidator->getUrlIfValid($item->link_uri ?? '<none>');
+          // Check if the link is an anchor within the current page
```
bberndt-uaz, line 135: Adding a period to this comment should fix the PHPCS error:
```suggestion
          // Check if the link is an anchor within the current page.
```

bberndt-uaz:
This works! From a [quick test in Probo](https://4f2f5037-2da7-4190-9640-f2b9f16d15ea.probo.build/brian-test-page), entering `#heading` in the Card Link URL field results in a link that looks like this:
```html
<a href="/brian-test-page#heading" class="btn-block">Card link to heading</a>
```

As an alternative, we could consider skipping the addition of the path and we could just use the anchor exactly as entered by the user. In that case, we could set `$link_url` to something like `Url::fromUri(urldecode('internal:' . $item->link_uri))`. Then the resulting href for the link would include only the anchor tag, matching how it currently works when creating an anchor link in a text paragraph:

```html
<a href="#heading" class="btn-block">Card link to heading</a>
```

BaoNguyen09:
> Good improvement that works. Could look into the suggested alternative. I also see that the Link field formatter works fine with these kinds of links. Could investigate how the field formatter works (e.g., the 'Where' field on Event type works properly).

Can you provide me where the code for it lies, I've looked through the az_event module but I didn't find the code for this

BaoNguyen09:
Then I'm fine with this

bberndt-uaz:
I looked into this issue a little further: with both the current solution and the alternative one I proposed earlier, entering only `#` into the Card Link URL field results in an href attribute without the hashtag, so the page reloads instead of simply jumping to the top. For example, with the current code, entering `#` results in `<a href="/some-page">Link</a>` instead of `<a href="/some-page#">Link</a>`.

It turns out that in Drupal, creating a URL object that solely contains `#` is impossible due to this code in UrlGenerator: https://git.drupalcode.org/project/drupal/-/blob/11.x/core/lib/Drupal/Core/Routing/UrlGenerator.php#L276-280.
```php
      if (isset($options['fragment'])) {
        if (($fragment = trim($options['fragment'])) != '') {
          $fragment = '#' . $fragment;
        }
      }
```

When a Url is meant to point to `#anchor`, the fragment option of the Url object will be `anchor`. But when a link is only `#`, the fragment ends up being empty.

Is it important that card links to `#` jump to the top of the page instead of reloading the page? (This behavior would match links to `#` in text areas.) If so, I could make some updates to [az-card.html.twig](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_card/templates/az-card.html.twig) to provide more specific handling of links from [AZCardDefaultFormatter.php](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php). Is it worth it to do this?

trackleft:
The test only fails because the branch is out of date with main.

modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
```diff
@@ -125,13 +125,20 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
       if ($item->link_title || $item->link_uri) {
         if (str_starts_with($item->link_uri ?? '', '/' . PublicStream::basePath())) {
           // Link to public file: use fromUri() to get the URL.
           $link_url = Url::fromUri(urldecode('base:' . $item->link_uri));
         }
         else {
-          $link_url = $this->pathValidator->getUrlIfValid($item->link_uri ?? '<none>');
+          // Check if the link is an anchor within the current page.
+          if (str_starts_with($item->link_uri, "#")) {
+            $link_url = $this->pathValidator->getUrlIfValid($current_path . $item->link_uri);
```
bberndt-uaz, line 137: To prevent an error from occurring when `$item->link_uri` is null, we can add a null coalescing operator to line 136 (just like the previous if statement above). In line 137, we can simplify the variable assignment and use the Drupal [Url::fromUserInput](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Url.php/function/Url%3A%3AfromUserInput/10) function, which was designed for this use case. Then we won't need the new `$current_path` variable.

```suggestion
          if (str_starts_with($item->link_uri ?? '', "#")) {
            $link_url = Url::fromUserInput($item->link_uri);
```

modules/custom/az_card/src/Plugin/Field/FieldFormatter/AZCardDefaultFormatter.php
```diff
@@ -125,13 +125,20 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
       // Link.
       $link_render_array = [];
       $link_url = '';
+      $current_path = \Drupal::service('path.current')->getPath();
```
bberndt-uaz, line 128: With the code change from my other suggestion, we can remove this line since we won't need this variable any longer.

```suggestion
```

bberndt-uaz:
@BaoNguyen09  I have made a couple code suggestions above: if you think they look good, you can commit them and then I think this PR will be ready for final review!

I added a note to https://quickstart.arizona.edu/create-content/editor-toolbar/edit-links about using `#top` to jump to the top of the page. Personally, I don't think we need to add help text in Quickstart about that. However, if we want to do that, I recommend handling that in a separate issue since it would apply to any links, not just card links.

Finally, the only other Drupal issue I could find that was somewhat related was this one (which was fixed in Drupal 8):
 - [`<none>` does not work as expected with fragments](https://www.drupal.org/node/2401395)

I did try a solution using the `<none>` route and ran into the same issue with links that are just the `#` character. I think we can consider `#top` as the recommended way to provide a link to the top of the page.

BaoNguyen09:
Looks good to me!

bberndt-uaz (approved):
Approved! I made a test page here: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-3933.probo.build/brian-test-page.

joeparsons (approved):

trackleft (approved):

tadean (approved):

PR #3966, opened 2025-01-07
Closes #3003: Add menu parent to chosen settings

### LINKED ISSUE #3003: Add menu settings to Chosen to always be enabled
## Motivation
Chosen is set to only apply to visible select menus and the 'edit-field-az-page-category' select menu. It should apply to more select menus by default, including node menu selection.

When creating or editing a node that is not in the menu, the menu selection is collapsed and not visible. Opening the menu area and clicking to "Provide a menu link," the Parent link select doesn't get Chosen applied.

Editing a node that is already in the menu doesn't run into this problem. Chosen gets applied appropriately to the Parent link select.

Add `select#edit-menu-menu-parent` to the Chosen jQuery selector list.

## Additional context
Menu currently doesn't get Chosen applied for new nodes or nodes not yet in the menu.
image

Menu currently gets Chosen applied for nodes already in the menu.
image

joshuasosa:
## Description
Adds a jQuery selector to Chosen: when creating or editing a node that has a main menu option, the dropdown will now use Chosen instead of the regular dropdown menu.

Nodes already in the main menu already get Chosen applied to the dropdown menu. This PR therefore adds parity.

### Before
image

### After
image

## Related issues
Closes #3003 

## How to test

1. Visit `/node/add/az_flexible_page`.
2. Open "Menu settings" accordion in the sidebar.
3. Checkmark "Provide a menu link."
4. Click open the "Parent menu" dropdown and see that it uses Chosen module.

## Width Considerations
The width of the Chosen dropdown menu for Main menu selecting may look too narrow. That's due to default Chosen settings currently setting "Minimum width of widget" to 150px. Instead of the associated CSS using `min-width`, it sets a fixed `width`.

The minimum width might be set when the Chosen menu is in an area that hasn't been visibly displayed yet. It seems to calculate a width properly for a node's Main menu selector when that node is already in the menu.

Options to fix:
A. An issue and patch could be created for Chosen to change `width` to `min-width` in the CSS. 
B. Minimum width could be set to a relative percentage (e.g., 30 with "Use relative width" enabled).
C. Minimum width could be unset and left blank, where `width: auto` will then get applied.

It might be good to address this in a separate issue to improve usability since it is a global change.

### Example of node's menu Chosen width already in menu
image

## Types of changes
Minor enhancement.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [x] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/config/quickstart/chosen.settings.yml
+++ b/modules/custom/az_core/config/quickstart/chosen.settings.yml
@@ -4,7 +4,7 @@ disable_search_threshold: 0
 minimum_width: 150
 max_shown_results: null
 use_relative_width: false
-jquery_selector: 'select:visible,select#edit-field-az-page-category'
+jquery_selector: 'select:visible,select#edit-field-az-page-category,select#edit-menu-menu-parent'
 search_contains: true
 disable_search: false
 allow_single_deselect: false


bberndt-uaz (approved):
Looks like a good change to me!

However, I agree that the dropdown looks too narrow and that would be best addressed in a separate issue. Hopefully, we could get that taken care of sooner rather than later after merging this. From some quick testing locally, I think we could leave that Chosen "Minimum width of widget" option blank (so the container would be set to `width: auto`) and then we could also override the `.chosen-drop` class to have `width: max-content`. That seemed to provide a good solution, although there might be a better one.

trackleft (approved):

tadean (approved):

PR #4050, opened 2025-02-07
Closes #4028: Update person fields and display

danahertzberg:
## Description
This adds new fields to the Person content type and displays them in a tabbed format.

### New fields
- az_awards
- az_teaching_interests
- az_research_interests
- az_licensure_certification
- az_work_experience

### Form changes
- New Experience section (collapsed by default)
- New Interests section (collapsed by default)
- Degrees moved into Experience section
- az_seo updated to match these changes to the form

### Default display changes
- New tabbed display of Experience and Interests sections
   - Experience contains: Degrees, Work Experience, Awards, Licensure & Certification fields
   - Interests contains: Teaching Interests & Research Interests fields
- Degrees moved into Experience tab section

### Release notes

If this change requires release notes: provide a summary of changes, how to use this change, and any related links. This content will be pasted in the [release notes](https://github.com/az-digital/az_quickstart/releases). Use markdown format to ensure proper pasting of information. [Release notes example](https://github.com/az-digital/az_quickstart/releases/tag/2.11.0)

Make sure to add the `release notes` label to this PR.

```
AZ Person has new fields for Awards, Interests, Licensure & Certification, and Work Experience. There are new Experience and Interest sections are found at the bottom of the AZ Person editing interface. The editing interface for the Degrees field has been moved into the new Experience section. These new fields and existing Degrees field are now displayed in a tabbed format when viewing the published page. 
```

#### Documentation update
https://quickstart.arizona.edu/create-content/people-directories/managing-people

## Related issues
Relates to #4028 

## How to test

Option 1
1. Add or edit a person
2. Complete the fields in the new Experience and Interests sections
3. Observe added tab styles on the front-end

Option 2
1. Select a person from profiles.arizona.edu
2. Copy their netID
3. On probo, go to Content > Import profiles
4. Paste the copied netID
5. Import
6. Observe profile content on the website with imported data

### Arizona Quickstart (install profile, custom modules, custom theme)
- **Patch release changes**
   - [ ] Bug fix
   - [ ] Accessibility, performance, or security improvement
   - [ ] Critical institutional link or brand change
   - [ ] Adding experimental module
   - [ ] Update experimental module
- **Minor release changes**
   - [ ] New feature
   - [X] Breaking or visual change to existing behavior
   - [ ] Upgrade experimental module to stable
   - [ ] Enable existing module by default or database update
   - [ ] Non-critical brand change
   - [ ] New internal API or API improvement with backwards compatibility
[...]

PR #4109, opened 2025-03-06
[Branch for issue/4028] Try twig template override for horizontal tabs

bberndt-uaz:
## Description

This PR is a branch of the `issue/4028` branch as tracked in this PR:

 - #4050 

This PR does the following:

1. Adds a twig template which overrides the display of horizontal tabs from the Field Group module and instead displays the tabs in the style of Arizona Bootstrap. See https://digital.arizona.edu/arizona-bootstrap/docs/2.0/components/navs/#javascript-behavior.
2. Removes CSS styling associated with classes from the Field Group module. Some of this could be added back as needed, but I thought it would be best to start fresh from the default Arizona Bootstrap styling.
3. Moves a little styling to the AZ Person default view display (as classes on the "Tabs" element).
4. Removes the JS library, which should hopefully not be needed any longer.

To do:
 - ~~Fix issue when a tab is empty~~

## How to test

Probo review site: https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-4109.probo.build


modules/custom/az_person/templates/horizontal-tabs--node--az-person--group-tabs.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    {% set isFirstTab = true %}
+    {% for key, value in tabElements %}
+      <li class="nav-item" role="presentation">
+        <a class="nav-link{{ isFirstTab ? ' active' }}" id="{{ (key) }}_tab" data-toggle="tab" href="#{{ (value['#id']) }}" role="tab" aria-controls="{{ (value['#id']) }}" aria-selected="{{ isFirstTab }}">{{ (value['#title']) }}</a>
+      </li>
+      {% set isFirstTab = false %}
+    {% endfor %}
+  </ul>
+  <div class="tab-content px-3 pb-3 mt-3">
+    {% set activePane = ' show active' %}
```
danahertzberg, line 34:
```suggestion
    {% set activePane = 'show active' %}
```

modules/custom/az_person/templates/horizontal-tabs--node--az-person--group-tabs.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    {% for key, value in tabElements %}
+      <div class="tab-pane fade px-1 pb-1{{ activePane }}" id="{{ (value['#id']) }}" role="tabpanel" aria-labelledby="{{ (key) }}_tab">
```
danahertzberg, line 36:
```suggestion
      <div class="tab-pane fade px-1 pb-1 {{ activePane }}" id="{{ (value['#id']) }}" role="tabpanel" aria-labelledby="{{ (key) }}_tab">
```
Fixes the interests tab not appearing.

bberndt-uaz: That lack of a space there is actually intentional so if the pane is active the classes will end in `pb-1 show active"` and if the pane is not active the classes will end in `pb-1"`. We use a similar technique in our other twig files (e.g., `<div{{ attributes.addClass(classes) }}>` will simply end up as `<div>` if there are no classes).

bberndt-uaz:
Last tasks before merging:
 - Add az-horizontal-tabs class
 - Fix empty tabs element when all fields are empty

bberndt-uaz:
@danahertzberg I just made a commit which:

- Skips rendering anything if the tab elements are all empty
- Updates the Tabs field group to have only the new `az-horizontal-tabs` class (I moved the margin and border classes to the main div in the template.)
- Change the `.tab-content` div in the template to use the handy `-card` classes! I left the top spacing as a margin instead of padding so that it will collapse with the top margin of the heading in the tab content.

I think this PR is ready to be merged into your branch, but if you would like me to make any more changes, just let me know!

modules/custom/az_person/templates/horizontal-tabs--node--az-person--group-tabs.html.twig
danahertzberg: Ah I see. Thanks!

danahertzberg (approved):

PR #4144, opened 2025-03-21
Fixes #4137 Citation-key Unique Values

### LINKED ISSUE #4137: The `citation-key` is not a sufficiently unique identifier for the `az_publication_bibtex_import` migration.
## Problem/Motivation
When importing two distinct publication DOI's via the form located at `/admin/content/az-publication/import/doi`
the importer updates the same node for multiple DOI's instead of two separate nodes.

### Describe the bug

> Something weird seems to be going on with a couple publications I am trying to import: 10.1016/j.geoderma.2023.116377 (Mechanisms and health implications of toxicity increment from arsenate-containing iron minerals through in vitro gastrointestinal digestion) and 10.1016/j.gca.2023.09.012 (the effect of biogeochemical redox oscillations on arsenic release from legacy mine tailings)
> I import one and after importing the second one the first one is gone taking over its node number. Same the other way around.

To trouble shoot I ran the following CLI commands
```
curl -LH "Accept: application/x-bibtex" https://doi.org/10.1016/j.geoderma.2023.116377
curl -LH "Accept: application/x-bibtex" https://doi.org/10.1016/j.gca.2023.09.012
```

Where I can note the similarities and differences of the returned Bibtex.
```
 @article{Liu_2023, title={Mechanisms and health implications of toxicity increment from arsenate-containing iron minerals through in vitro gastrointestinal digestion}, volume={432}, ISSN={0016-7061}, url={http://dx.doi.org/10.1016/j.geoderma.2023.116377}, DOI={10.1016/j.geoderma.2023.116377}, journal={Geoderma}, publisher={Elsevier BV}, author={Liu, Ruiqi and Kong, Shuqiong and Shao, Yixian and Cai, Dawei and Bai, Bing and Wei, Xiaguo and Root, Robert A. and Gao, Xubo and Li, Chengcheng and Chorover, Jon}, year={2023}, month=apr, pages={116377} }
```

```
 @article{Liu_2023, title={The effect of biogeochemical redox oscillations on arsenic release from legacy mine tailings}, volume={360}, ISSN={0016-7037}, url={http://dx.doi.org/10.1016/j.gca.2023.09.012}, DOI={10.1016/j.gca.2023.09.012}, journal={Geochimica et Cosmochimica Acta}, publisher={Elsevier BV}, author={Liu, Yizhang and Root, Robert A. and Abramson, Nate and Fan, Lijun and Sun, Jing and Liu, Chengshuai and Chorover, Jon}, year={2023}, month=nov, pages={192–206} }
```

These both appear in the `migrate_map_az_publication_bibtex_import` table in the db as shown in the screenshot.

Image

At the moment, Quickstart is using a `citation_key` as the unique id.

https://github.com/az-digital/az_quickstart/blob/7bc0bd998a0129b8327ad8371f6c91288c3b6d7b/modules/custom/az_publication/az_publication_bibtex/migrations/az_publication_bibtex_import.yml#L11-L12

The `citation-key` field is created via https://github.com/renanbr/bibtex-parser/blob/16f68c2e9785fb2b485d1b97887633f4ad3ddba1/src/Parser.php

### To Reproduce

1. Go to '/admin/content/az-publication/import/doi'
2. Import '10.1016/j.geoderma.2023.116377'
3. Import '10.1016/j.gca.2023.09.012'
4. See same node updated with different data.

## Proposed resolution
Unknown since changing the ID could cause weird behavior for existing migrations.

### Expected behavior
Importing a publication with one DOI should not overwrite publications with a different DOI

tadean:
This PR attempts to provide a method for adding additional uniqueness to publication imports by allowing the data parser to add additional values to the citation key.

## Related issues
#4137 

## How to test

- enable `az_publication` and `az_publication_doi`
- visit `/admin/content/az-publication/import`
- import multiple publications which have similar citation keys in their content negotation bibtex output, such as:
  - `10.1016/j.gca.2023.09.012`
  - `10.1016/j.geoderma.2023.116377`
- verify both publications are imported and do not overwrite one another

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_publication/az_publication_bibtex/src/Plugin/migrate_plus/data_parser/AZBibtex.php
+++ b/modules/custom/az_publication/az_publication_bibtex/src/Plugin/migrate_plus/data_parser/AZBibtex.php
@@ -29,6 +29,19 @@ class AZBibtex extends DataParserPluginBase {
   protected $citations = [];
 
   /**
+   * Additional arguments for the citation key.
+   */
+  protected array $suffix;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
+    $this->suffix = $configuration['citation_key_suffix'] ?? [];
+  }
+
+  /**
    * {@inheritdoc}
    */
   protected function openSourceUrl($url): bool {
@@ -51,6 +64,15 @@ class AZBibtex extends DataParserPluginBase {
       $citations = array_filter($citations, function ($citation) {
         return (isset($citation['citation-key']) && isset($citation['title']));
       });
+      // Concat configured suffixes onto the citation key.
+      // @todo make into a processor.
+      foreach ($citations as &$citation) {
+        foreach ($this->suffix as $s) {
+          if (!empty($citation[$s])) {
+            $citation['citation-key'] .= '_' . (string) $citation[$s];
+          }
+        }
+      }
       $this->citations = $citations;
     }
     catch (ExceptionInterface $exception) {
--- a/modules/custom/az_publication/az_publication_doi/src/Form/AZPublicationDOIForm.php
+++ b/modules/custom/az_publication/az_publication_doi/src/Form/AZPublicationDOIForm.php
@@ -95,6 +95,9 @@ class AZPublicationDOIForm extends FormBase {
           'source' => [
             'urls' => $url,
             'data_fetcher_plugin' => 'http',
+            'citation_key_suffix' => [
+              'doi',
+            ],
             'headers' => [
               // Content Negotiation. https://citation.crosscite.org/docs.html
               'Accept' => 'application/x-bibtex; charset=utf-8',


tadean:
@trackleft notes in #4137 that this might alter existing migrations. Perhaps we can offer release notes, or provide a way for the import form to configure this behavior?

trackleft:
Interesting, I think I like the idea of adding the suffix as configuration.
That way we can leave it off existing migrations.

I wonder what an upgrade path would look like.

trackleft (approved):
@camikazegreen Here is an example of what the `migrate_map_az_publication_bibtex_import` table looks like after re-importing the problematic publications with this change in place:
image

The previously imported publication remains mapped to node `709`, and two new nodes (`1084` and `1083`) are created on re-import. The source row is still rebuilt consistently from the form input — the difference is that the source ID used in the migrate map no longer matches the original one, so the re-import is treated as a new item rather than updating the existing node.

Given that:

- The module is experimental
- The number of sites using it is known and limited,
- Re-imports are done manually and one-at-a-time through the form. (`drush migrate:import --update` does nothing with or without this PR)
- Re-imports via form show clear messaging.

image

I can live with duplicate node creation on re-import here. This avoids unintentional overwrites of existing publications and is the simplest behavior to reason about in the current workflow.

We can revisit update behavior later if we want a more sophisticated approach to key stability, but this solves the immediate overwrite problem effectively.

modules/custom/az_publication/az_publication_bibtex/src/Plugin/migrate_plus/data_parser/AZBibtex.php
```diff
@@ -28,6 +28,19 @@ class AZBibtex extends DataParserPluginBase {
...
+  /**
+   * Additional arguments for the citation key.
+   */
+  protected array $suffix;
+
+  /**
+   * {@inheritdoc}
+   */
+  public function __construct(array $configuration, $plugin_id, $plugin_definition) {
+    parent::__construct($configuration, $plugin_id, $plugin_definition);
```
joeparsons:
```suggestion
  public function __construct(array $configuration, $plugin_id, $plugin_definition, DataFetcherPluginManager $plugin_Manager) {
    parent::__construct($configuration, $plugin_id, $plugin_definition, $plugin_Manager);
```

PR #4205, opened 2025-04-07
Fixes #4203, Fixes #4216 Update Add to Calendar Modal Links

### LINKED ISSUE #4203: Update Outlook Links on Add to Calendar Modal
## Motivation
Most of our user base uses Office365, and the Outlook.com button does not work for these users, so the existence of two similar options may be unnecessarily confusing.

## Proposed Resolution
~~Remove Outlook.com link from add to calendar modal in preference to the office365 link.~~

Instead, rename the outlook buttons to be `Outlook Personal` and `Outlook Enterprise` to better reflect the way these links get used.

### LINKED ISSUE #4216: Add to Calendar button does not properly support All-day events
## Problem/Motivation
The Add to Calendar button does not create all-day events for dates marked so by smart date, instead just representing them as an event spreading the entire day - minus one granularity.

## Proposed resolution
Pass all-day information to calendar link creation.

### Expected behavior
The created calendar events should be flagged as all day events.

tadean:
## Description
~~This PR removes the add to calendar link for Outlook.com from the modal. Most of our userbase uses Office365, and these two very similar options might be confusing. It appears that the Outlook.com link will not work correctly with Office365 accounts.~~

This PR updates the add to calendar links for Outlook in the add to calendar modal. Most of our userbase uses Office365, and these two very similar options might be confusing. It appears that the Outlook.com link will not work correctly with Office365 accounts. This PR changes the names of the links from the potentially confusing `Outlook.com` and `Office365` to instead be `Outlook Personal` and `Outlook Enterprise`

## Related issues
#4203 
#4216 

## How to test

- enable demo content
- visit an event page that has a future occurrence
- Verify add to calendar link appears
- Click add to calendar link
- Verify that Outlook.com link is not present
- Verify that Google, Ical, yahoo links appear unchanged.
- Verify that the Outlook.com and Office365 links are now Outlook Personal and Outlook Enterprise respectively.
- Verify both links still work

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


tadean:
> I would request a block on this PR as event pages are generally public-facing. The general public may not use Office365. Office.com is meant for personal Outlook email accounts, and removing the option makes it harder for those users to use this "Add to Calendar" option.

I agree. Perhaps a change in wording or help text would fill the original issue better than removing the option.

tadean:
Updated to update issue description and incorporate a suggestion from @joshuasosa to leave both links in place but to use Personal/Enterprise language respectively:
image

danahertzberg (dismissed):

trackleft (dismissed):

tadean:
During PR review @joeparsons potentially suggests:
```
Outlook.com (Personal)
Office 365 (Enterprise)
```
or
```
Outlook Personal (outlook.com)
Outlook Enterprise (Office365)
```
@danahertzberg notes that part of the confusion is that Office365 users might expect "Outlook.com" to work for them
@joeparsons notes that Office365 users might be specifically looking for Office365

tadean:
Updated per @danahertzberg 's recommendations, thanks! I think the block buttons are a good idea, since the options are rather different in overall length.

image

bberndt-uaz:
A couple suggestions for consideration:
 - Change "iCal" to "Download iCal" or "Download .ics"
 - Move Yahoo! to the end of the list

tadean:
> A couple suggestions for consideration:
> 
> * Change "iCal" to "Download iCal" or "Download .ics"

I'm not sure if this would be accurate because the "download vs. open" behavior may depend somewhat on the system handler for ics files based on the user's environment.

bberndt-uaz:
> I'm not sure if this would be accurate because the "download vs. open" behavior may depend somewhat on the system handler for ics files based on the user's environment.

Right, makes sense! Maybe something like "iCal File" could work then.

tadean:
- Added @bberndt-uaz 's suggestions on order and ical naming
- Added support for all-day events

danahertzberg (approved):

bberndt-uaz (approved):

modules/custom/az_event/templates/az-event-calendar-add-button.html.twig
```diff
@@ -7,7 +7,15 @@
 */
 #}
 {% set attributes = attributes.addClass('clear-both') %}
-{% set links = calendar_links(title, date(start_date), date(end_date), FALSE, description, location) %}
+{# List and order of calendar implementations. #}
+{% set calendar_types = {
+    'google' : 'Google',
+    'ics' : 'iCal File',
+    'webOffice': 'Outlook Enterprise (Office365)',
```
joeparsons, line 14:
```suggestion
    'webOffice': 'Outlook Enterprise (Office 365)',
```

I think I originally suggested `Office365` but I think that should actually include a space.

trackleft (approved):

tadean:
Update with @bberndt-uaz 's recurrence changes and @joeparsons suggestion!

bberndt-uaz (commented):
@tadean This looks great, but I just noticed my comment on line 124 of az_event.module is inaccurate:

```
// For the full content view mode, modify the display of recurring events.
    if (($view_mode === 'full') && ($entity instanceof NodeInterface)) {
```

I should have updated that in my PR. Maybe we could change it here to "modify the display of az_event_date" (or whatever makes sense and will fit under the character limit).

joeparsons (approved):

danahertzberg (approved):

PR #4195, opened 2025-04-03
Closes #4047: when-using-inline-image-text-wrap-bulleted-text-needs-more-left-padding

BaoNguyen09:
Update the `margin-right` of `align-left` class to 1.5em

## Description

This is also the `margin-right` of `drupal-media-style-align-left` class in edit mode

## Related issues

Closes #4047 

## How to test

1. Create a new page with Text module
2. Add an image and align it to the left, then add some bulleted text to the right
3. Save and view the published page

### Arizona Quickstart (install profile, custom modules, custom theme)
- **Patch release changes**
   - [x] Bug fix
   - [ ] Accessibility, performance, or security improvement
   - [ ] Critical institutional link or brand change
   - [ ] Adding experimental module
   - [ ] Update experimental module
- **Minor release changes**
   - [ ] New feature
   - [ ] Breaking or visual change to existing behavior
   - [ ] Upgrade experimental module to stable
   - [ ] Enable existing module by default or database update
   - [ ] Non-critical brand change
   - [ ] New internal API or API improvement with backwards compatibility
   - [ ] Risky or disruptive cleanup to comply with coding standards
   - [ ] High-risk or disruptive change (requires upgrade path, risks regression, etc.)
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
   - [ ] Patch level release (non-security bug-fix release)
   - [ ] Patch removal that's no longer necessary
- **Minor release changes**
   - [ ] Major or minor level update
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Patch or minor level update
   - [ ] Add new module
   - [ ] Patch removal that's no longer necessary
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have read the **CONTRIBUTING** document.
- [ ] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [ ] My change requires release notes.

PR #4219, opened 2025-04-11
[Branch for PR 4195] Styling fixes for left-aligned content

bberndt-uaz:
## Description
This PR is a branch of the `4047-when-using-inline-image-text-wrap-bulleted-text-needs-more-left-padding` branch as tracked in this PR:
 - #4195

This PR adds some styling to two CSS files:

- claro-node-form-overrides.css: Copies some styling from style.css so the node edit form can match the published page. Also includes styling fixes for sibling list elements, including a `padding-inline-start` value to revert from 0px (the override from the Claro theme) to 40px (from the browser user agent stylesheet).
- ckeditor5.css: Includes the same styling fixes for sibling list elements. Also removes the `margin-right` since that is being added with our Claro overrides now.

This solution seems to work but it would be good to do more testing with varied content to ensure it doesn't cause any unwanted side effects.

## How to test

https://37d4d6cf-4042-4492-9ac4-523fd71b504f.probo.build

### DIFF
--- a/modules/custom/az_core/css/claro-node-form-overrides.css
+++ b/modules/custom/az_core/css/claro-node-form-overrides.css
@@ -44,5 +44,5 @@ tr.odd {
 .align-left ~ ol {
   overflow: auto;
   overflow-wrap: anywhere;
-  padding-inline-start: 40px;
+  padding-inline-start: revert;
 }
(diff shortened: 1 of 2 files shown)


BaoNguyen09 (approved):
LGTM!

PR #4238, opened 2025-04-16
Closes #4237: PHPStan level 3 fixes.

### LINKED ISSUE #4237: Increase PHPStan rule level to 3
## Motivation
Starting with Drupal 11.1.x, [Return types are required in all new methods and functions](https://www.drupal.org/node/3461320).

## Proposed Resolution
We should consider increasing our [PHPStan rule level](https://phpstan.org/user-guide/rule-levels) from 2 to 3 to better detect issues with incorrect return types, etc.

joeparsons:
## Description
Raises the PHPStan rule level from 2 to 3 and fixes the resulting errors.

### Type/nullability fixes
- **`AZAuthor.php`**: `getOwner()`/`getOwnerId()` now correctly handle the  nullable/string-vs-int types the `user_id` entity reference field actually  returns, instead of promising non-nullable `UserInterface`/`int`.
- **`AZProfilesAPIFetcher.php`**: `create()` now returns `static` instead of  `self`, matching the actual late-static-bound instance it constructs.
- **`AZSecrets.php`**: `getKeyValue()` returns `''` instead of `NULL` for a  missing secret, satisfying the Key module's non-nullable `string` contract.
- **`EntityEmbedProcess.php`**: `updateEmbedTag()`'s return type corrected to `\DOMElement|null` (was `\DOMDocument`), with a guard for `DOMDocument::createElement()`'s documented `false` return.
- **`AZParagraphsItem.php`**: `prepareRow()` now delegates to `parent::prepareRow()` (returns `bool`) instead of returning the `Row` object directly.
- **`AZPoliteCacheStrategy.php`**, **`AZEntityRoleReferenceItem.php`**: bare `return;` changed to `return NULL;` to match declared nullable return types.
- **`az_quickstart.install`**, **`az_global_footer.install`**: magic property writes on `MenuLinkContent` replaced with `->set()`, or scoped `@phpstan-ignore` where core's own field API doesn't offer a clean alternative for a one-shot update hook.

### Entity/form cleanup
- **`AZPublicationType.php`/`Interface.php`**, **`AZPublicationTypeListBuilder.php`**: removed a redundant `load()` override whose covariant return type didn't match the parent contract; grouping logic inlined into `render()`.
- **`AZAuthorRevisionDeleteForm.php`/`AZAuthorRevisionRevertForm.php`**:
  replaced injected `$authorStorage` object with `$entityTypeManager` service; `$revision` is now correctly typed nullable, with a `NotFoundHttpException` guard for a stale/invalid revision ID (previously an unguarded null dereference).
- **`AZAuthorInterface.php`**: removed method redeclarations already inherited from `RevisionLogInterface`.

### Suppressions (with justification)
- **`AzDropbutton.php`**: `render()` legitimately returns a render array, matching core's own `Dropbutton` field plugin despite the interface documenting `string|MarkupInterface`.
- **`AZAuthorRevisionRevertForm::getDescription()`**: kept as `return '';` with a targeted ignore — `$this->t('')` satisfies PHPStan but fails phpcs's "do not pass empty strings to t()" check.

### Also included
- Corrected several test classes' property docblocks to `{@inheritdoc}` for covariance with `BrowserTestBase`.
- `phpstan.neon`: level 2 → 3, plus `entityMapping` entries for `az_ldap_role` and `az_recurring_import_rule`.

## Related issues
Closes #4237 

## How to test

General testing to make sure changes didn't break anything.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
   - [x] Risky or disruptive cleanup to comply with coding standards
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_publication/src/Form/AZAuthorRevisionRevertForm.php
```diff
@@ -88,13 +88,14 @@ public function getConfirmText() {
    * {@inheritdoc}
    */
   public function getDescription() {
-    return '';
+    return $this->t('');
```
joeparsons, line 91:
```suggestion
    /* @phpstan-ignore-next-line */
    return '';
```

modules/custom/az_event/az_event.module
```diff
@@ -153,13 +153,11 @@ function az_event_node_view(array &$build, EntityInterface $entity, EntityViewDi
       $number_of_occurrences = $event_date_field->count();
       if ($number_of_occurrences <= 1) {
         // Add to Calendar button for an event with a single occurrence.
```
trackleft:
```suggestion
        // Add to Calendar button for an event with a single occurrence.
        $field_item = $entity->get('field_az_event_date')->first();
        $field_render_array = $field_item ? $field_item->view('default') : [];
```

modules/custom/az_paragraphs/src/Plugin/Field/FieldFormatter/AZBackgroundMediaFormatter.php
```diff
@@ -340,20 +340,18 @@ public function viewElements(FieldItemListInterface $items, $langcode): array {
 
     /** @var \Drupal\media\MediaInterface $media */
     foreach ($media_items as $delta => $media) {
-      $element['#media_type'] = $media->bundle();
-
       switch ($media->bundle()) {
         case 'az_remote_video':
           $element[$delta] = $this->remoteVideo($settings, $media);
+          $element[$delta]['#media_type'] = $media->bundle();
```
trackleft:
```suggestion
```

bberndt-uaz (dismissed):

bberndt-uaz (approved):

modules/custom/az_event/az_event.module
```diff
@@ -156,11 +156,11 @@ function az_event_node_view(array &$build, EntityInterface $entity, EntityViewDi
...
         $field_item = $entity->get('field_az_event_date')->first();
         $field_render_array = $field_item ? $field_item->view('default') : [];
-
-        foreach ($entity->field_az_event_date as $item) {
+        /** @var \Drupal\smart_date\Plugin\Field\FieldType\SmartDateItem $item */
+        foreach ($event_date_field as $item) {
           if (($item->value >= $now) || (!empty($item->end_value) && ($item->end_value >= $now))) {
             $start = $item->value;
-            $end = $item->end_value;
+            $end = $item->get('end_value')->getValue();
```
tadean, line 163: Does this return the same information as `$item->end_value;` ? I thought this might be one of the situations where the one returned a scalar and one returned an array...

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZTextWithMediaParagraphBehavior.php
```diff
@@ -209,8 +209,8 @@ public function preprocess(&$variables) {
     $variables['attributes']['class'][] = 'paragraph';
     $variables['attributes']['class'][] = 'position-relative';
     $variables['attributes']['class'][] = HTML::getClass('paragraph--type--' . $paragraph->bundle());
-    if (!empty($variables['content']['field_az_media']['#media_type'])) {
-      $variables['attributes']['class'][] = 'media--type--' . HTML::getClass($variables['content']['field_az_media']['#media_type']);
+    if (!empty($variables['content']['field_az_media'][0]['#media_type'])) {
```
tadean, line 212: Is this array nesting a preexisting bug that needs to be fixed in other branches?

modules/custom/az_publication/src/Entity/AZPublicationType.php
```diff
@@ -143,11 +143,4 @@ public static function getMappableTypeOptions():array {
     ];
   }
 
-  /**
-   * {@inheritdoc}
-   */
-  public static function load($id): ?AZPublicationTypeInterface {
-    return \Drupal::entityTypeManager()->getStorage('az_publication_type')->load($id);
```
tadean, line 150: Is this deleted because the parent implementation does the same thing?

modules/custom/az_secrets/src/Plugin/KeyProvider/AZSecrets.php
```diff
@@ -117,7 +117,7 @@ public function getKeyValue(KeyInterface $key) {
     $key_value = getenv($secret_name);
 
     if (!$key_value) {
-      return NULL;
+      return '';
```
tadean, line 120: This change does seem accurate:
```
  /**
   * Gets the value of the key.
   *
   * @return string
   *   The value of the key.
   */
  public function getKeyValue();
```

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZTextWithMediaParagraphBehavior.php
joeparsons: That seems possible.  I can investigate and break that change out into it's own issue/PR if needed.

modules/custom/az_publication/src/Entity/AZPublicationType.php
joeparsons: Yes, more or less.  I think PHPStan was also complaining about the return type not being covariant with the parent interface.
>AZPublicationType.php/Interface.php, AZPublicationTypeListBuilder.php: removed a redundant load() override whose covariant return type didn't match the parent contract

See`Drupal\Core\Entity\EntityInterface::load()` :
```
  /**
   * Loads an entity.
   *
   * @param mixed $id
   *   The id of the entity to load.
   *
   * @return static|null
   *   The entity object or NULL if there is no entity with the given ID.
   */
  public static function load($id);
```

joeparsons: Actual PHPStan error:
```
 ------ ----------------------------------------------------------------------------------------------------------------------------------------- 
  Line   web/profiles/custom/az_quickstart/modules/custom/az_publication/src/Entity/AZPublicationType.php                                         
 ------ ----------------------------------------------------------------------------------------------------------------------------------------- 
  150    Method Drupal\az_publication\Entity\AZPublicationType::load() should return static(Drupal\az_publication\Entity\AZPublicationType)|null  
         but returns Drupal\az_publication\Entity\AZPublicationType|null.                                                                         
         🪪  return.type                                                                                                                          
 ------ ----------------------------------------------------------------------------------------------------------------------------------------- 
```

modules/custom/az_paragraphs/src/Plugin/paragraphs/Behavior/AZTextWithMediaParagraphBehavior.php
joeparsons: Actually, this change was needed due to the change to [modules/custom/az_paragraphs/src/Plugin/Field/FieldFormatter/AZBackgroundMediaFormatter.php](https://github.com/az-digital/az_quickstart/pull/4238/files#diff-66f4d2015a6065d1695d4fd999684fcd62b28a8b45f0e1ebd659adec5b174b6d) (in this PR branch) that was made to address this PHPStan error:

```
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  Line   web/profiles/custom/az_quickstart/modules/custom/az_paragraphs/src/Plugin/Field/FieldFormatter/AZBackgroundMediaFormatter.php               
 ------ -------------------------------------------------------------------------------------------------------------------------------------------- 
  355    Method Drupal\az_paragraphs\Plugin\Field\FieldFormatter\AZBackgroundMediaFormatter::viewElements() should return array<int, array<int|stri  
         ng, mixed>> but returns array<array|string>.                                                                                                
         🪪  return.type                                                                                                                             
  359    Method Drupal\az_paragraphs\Plugin\Field\FieldFormatter\AZBackgroundMediaFormatter::viewElements() should return array<int, array<int|stri  
         ng, mixed>> but returns array<array|string>.                                                                                                
         🪪  return.type                                                                                                                             
 ------ --------------------------------------------------------------------------------------------------------------------------------------------
```

modules/custom/az_event/az_event.module
joeparsons: AI-assisted analysis:

> Yes — they return identical data, and here's why.
> 
> `FieldItemInterface`/`FieldItemBase` in Drupal core implements the magic getter as:
> 
> ```php
> public function __get($name) {
>   return $this->get($name)->getValue();
> }
> ```
> 
> So `$item->end_value` **is** `$item->get('end_value')->getValue()` under the hood — the property-access shorthand is just sugar for the exact same call. The change at [az_event.module:163](modules/custom/az_event/az_event.module:163) didn't alter what's returned at all; it just spelled out explicitly what the magic getter was already doing internally.
> 
> For `SmartDateItem::end_value`, that property's typed-data plugin is a timestamp (int), so `getValue()` returns a scalar int in both the old and new form — not an array. No type mismatch was introduced.
> 
> One thing worth double-checking: this presumably was another PHPStan-driven change (magic `__get()` calls are often flagged since PHPStan can't infer their return type). Since `getValue()` is explicit and typed, that's likely why the swap was made — for static-analysis clarity, not behavior change.
> 
> Confirmed against the actual core/contrib source in the DDEV container — both return the same type (a scalar int timestamp, or `NULL`).
> 
> **The mechanics:**
> - `SmartDateItem` (`web/modules/contrib/smart_date/src/Plugin/Field/FieldType/SmartDateItem.php:73`) defines `end_value` as `DataDefinition::create('timestamp')`.
> - `Timestamp` (`web/core/.../Plugin/DataType/Timestamp.php`) extends `IntegerData` and stores `protected $value` as a plain int. It has no custom `getValue()` override, so `getValue()` returns that raw int (or `NULL`) — never an array.
> - `Map::get('end_value')` (`web/core/.../Plugin/DataType/Map.php:118`) lazily instantiates that `Timestamp` property object, seeding it from `$this->values['end_value']`, and returns the property object.
> - `FieldItemBase::__get('end_value')` (`web/core/lib/Drupal/Core/Field/FieldItemBase.php:150`) does one of two things: if the property object was already instantiated, it returns `$this->properties['end_value']->getValue()` — literally identical to the new code. If not yet instantiated, it short-circuits and returns the raw `$this->values['end_value']` directly — same underlying scalar, just without building the property object.
> 
> So in both branches of the magic getter, the value returned is the same raw int timestamp (or `NULL`) that `->get('end_value')->getValue()` produces. `SmartDateItem` itself uses the exact `$this->get('end_value')->getValue()` pattern internally (line 187), confirming it's the idiomatic equivalent, not a behavioral change.
> 
> **Conclusion:** the change at [az_event.module:163](modules/custom/az_event/az_event.module:163) is safe — old and new both return a scalar (int timestamp or `NULL`), not an array.

tadean: I tested this locally to verify. The value of `$item->get('end_value')->getValue();` is a string unix timestamp. It looks like it ends up this way since the get is being done on the _item_, not the _item list_, so this looks correct. :+1:

tadean (approved):

Issue #4278, opened 2025-04-25
Events show wrong date in view listings other than calendar.


trackleft:
## Problem/Motivation
I'd like to use the calendar event display in other views but at the moment, the display doesn't work well with repeating events.
https://demo.quickstart.arizona.edu/finders/events

### Describe the bug
![Image](https://github.com/user-attachments/assets/bdee9de1-220e-45de-8d29-5f5cc72fe1b5)

### To Reproduce
Go to https://demo.quickstart.arizona.edu/finders/events

## Proposed resolution
Allow using the display in views other than the main calendar listing.

### Expected behavior

The number in the left box should reflect the date of the view row it is attached to.

tadean:
@BaoNguyen09 this probably requires altering the views that this customization is performed for:
https://github.com/az-digital/az_quickstart/blob/f6fbccec40f007b33297cf322599dea7a3b1e0bd/modules/custom/az_event/az_event.module#L374

BaoNguyen09:
What is the default View that this finders/events is using? Because when I go into the edit, it's just an empty View element

![Image](https://github.com/user-attachments/assets/52498788-dccf-412b-ac00-a363b2fde2dc)

PR #4294, opened 2025-05-02
Fixes #4291 AZ Event View Caching

### LINKED ISSUE #4291: Calendar sporadically shows all occurrences
### Describe the bug
Occasionally, an event will be seen on the calendar where a single row is showing all of the occurrences for the event, rather than the specific one that matches the row date.

### To Reproduce
Sporadically seen when viewing a recurring event on the calendar. Notably, this is often **not seen** while logged in.

### Expected behavior
For  the calendar view, only the occurrence that matches the row should be shown.

##  Screenshots

Image

tadean:
Quickstart has custom functionality that causes rendered events in views to  have their date field trimmed so that only the recurrence for the delta associated with that date shows in the rendered event.

This doesn't take into account that not every view has this functionality, and so that views which render the same event in the same display mode can seed the cache with render arrays for that display mode which do not trim the recurrence information. This can currently be seen in #4278 because the finder views do not trim recurrence information.

This can lead to circumstances where the calendar view inherits rendered events that do not have trimmed date fields, leading to calendars showing all the recurrences for a given event.

This PR introduces a cache context for events which cause them to have separate caches per route.

This problem could also potentially be fixed by making sure EVERY event view trims its dates properly so that no cache issues result, but that may not be able to account for custom views.

## Related issues
#4291 
#4278 (this is a related problem but not part of this PR)

## How to test

- enable `az_demo`
- **begin with a clear cache** `drush cr`
- be logged out, anonymous user **(logged in does not sufficiently test the issue)**
- visit the event finder `/finders/events`
- search for the name of a recurring event, such as `canyon running`
- see the output showing events without recurrence-trimmed dates (this is #4278 and not solved here)
- visit the calendar page, `/calendar`
- Page through events and verify that the event you searched for does not have the events shown with each event having all occurrences

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_event/az_event.module
+++ b/modules/custom/az_event/az_event.module
@@ -121,6 +121,11 @@ function az_event_node_view(array &$build, EntityInterface $entity, EntityViewDi
       }
     }
 
+    // Cache events per route to avoid cache issues from different views.
+    // Each page view has its own route.
+    // @todo do we need to be concerned about multiple view blocks of the same events per page?
+    $build['#cache']['contexts'][] = 'route.name';
+
     // For the full content view mode, modify the display of az_event_date.
     if (($view_mode === 'full') && ($entity instanceof NodeInterface)) {
       $now = \Drupal::time()->getCurrentTime();


tadean:
- [ ] Determine if route-based caching is sufficient, or if we need to solve the more complex case of multiple view blocks on the same page where each view uses the same event display mode but one renders their dates trimmed to delta occurrence and one does not
- [ ] The above would require an as-yet non-existent cache context that understands which view is being executed, potentially?

joeparsons (approved):
Seems like this would be worth merging as-is, even if it's determined additional cache contexts are also needed.

bberndt-uaz (approved):

tadean:
I'd like to check if a cache context that understands what view is being executed is technically possible before merging. This current implementation fixes the issue, but may cause unnecessary cache bloat.

modules/custom/az_event/az_event.module
```diff
@@ -121,6 +121,11 @@ function az_event_node_view(array &$build, EntityInterface $entity, EntityViewDi
       }
     }
 
+    // Cache events per route to avoid cache issues from different views.
+    // Each page view has its own route.
+    // @todo do we need to be concerned about multiple view blocks of the same events per page?
+    $build['#cache']['contexts'][] = 'route.name';
+
     // For the full content view mode, modify the display of az_event_date.
```
trackleft, line 129: We might be able to use view display here.

trackleft (approved):

PR #4341, opened 2025-05-20
Fixes #4338 Prepopulate Alphabetical index letters on AZ Person

### LINKED ISSUE #4338: Prepopulate Alphabetical Index Letters on Person where available
## Motivation
@BaoNguyen09 notes in work on #1869 that it would be useful if the index letter field prepopulated with a default value if possible. If persons do not have an index letter, they will not show up on views that use it.

## Proposed Resolution
Consider assigning index letters programmatically if the person content type has the index letter field available.

If a person is saved without an index letter field value, consider assigning one if it's programmatically possible to determine.

Complex cases (accent letters) may require `iconv` which we might(or not) want to avoid.

Leaving index letter field blank by default.

tadean:
## Description
This PR attempts to prepopulate the alphabetical index character on AZ Persons where:

- The content type has the field (`field_az_alphabetical_index`) added
- A last name is available
- The person is saved with `field_az_alphabetical_index` as empty, e.g. the editor did not add one themselves.
- Some special characters are supported, via [PhpTransliteration::removeDiacritics](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Transliteration%21PhpTransliteration.php/function/PhpTransliteration%3A%3AremoveDiacritics/10)

### Release notes

```
In circumstances where field_az_alphabetical_index has been added to the AZ Person content type, Quickstart will attempt to prepopulate the index letter field where possible.
```

## Related issues
#4338 

## How to test

- Log in
- Enable `az_alphabetical_listing`
- Visit `/admin/structure/types/manage/az_person/fields/reuse`
- Add `field_az_alphabetical_index` to the AZ Person content type
- Edit a person
- Leave the alphabetical index field empty
- Save the person
- Verify that the alphabetical index letter is set correctly, if the person's surname was a regular letter
- Create another person; use a compound last name such as `Van Gogh`
- Leave the index letter field empty
- Save the person
- Verify that the index letter is set to the leading letter of all portions of their surname
- Edit other persons and verify that Quickstart does not attempt to override your selections if you choose different index letters

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_person/az_person.module
+++ b/modules/custom/az_person/az_person.module
@@ -152,7 +152,7 @@ function az_person_theme() {
 /**
  * Implements hook_ENTITY_TYPE_presave().
  *
- * Prepulate the alphabetical index letter field for az_person where available.
+ * Prepopulate the alphabetical index letter field in az_person where available.
  */
 function az_person_node_presave(EntityInterface $entity) {
   if ($entity->bundle() === 'az_person'
--- a/modules/custom/az_person/az_person.module
+++ b/modules/custom/az_person/az_person.module
@@ -158,8 +158,7 @@ function az_person_node_presave(EntityInterface $entity) {
   if ($entity->bundle() === 'az_person'
     && $entity instanceof NodeInterface
     && $entity->hasField('field_az_alphabetical_index')
-    && $entity->hasField('field_az_lname'))
-  {
+    && $entity->hasField('field_az_lname')) {
     $last = $entity->get('field_az_lname')->value;
     $index = $entity->get('field_az_alphabetical_index');
     // Only operate if we have no current index values.
--- a/modules/custom/az_person/az_person.module
+++ b/modules/custom/az_person/az_person.module
@@ -6,8 +6,10 @@
  */
 
 use Drupal\Core\Entity\EntityFormInterface;
+use Drupal\Core\Entity\EntityInterface;
 use Drupal\Core\Entity\EntityTypeInterface;
 use Drupal\Core\Form\FormStateInterface;
+use Drupal\node\NodeInterface;
 
 /**
  * Implements hook_preprocess_node().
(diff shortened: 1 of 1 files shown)


modules/custom/az_person/az_person.module
```diff
@@ -146,3 +148,40 @@ function az_person_theme() {
...
+/**
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * Prepopulate the alphabetical index letter field in az_person where available.
+ */
+function az_person_node_presave(EntityInterface $entity) {
+  if ($entity->bundle() === 'az_person'
+    && $entity instanceof NodeInterface
+    && $entity->hasField('field_az_alphabetical_index')
+    && $entity->hasField('field_az_lname')) {
```
joeparsons, line 161: Opinionated suggestion to improve readability.
```suggestion
    && $entity->hasField('field_az_lname'))
  {
```
Apparently there Drupal's coding standards discourage breaking conditions into new lines currently but there's an open issue about relaxing that: https://www.drupal.org/project/coding_standards/issues/3392321

tadean:
> I've understood the scenario of multiple words in a last name to be treated as a single unit, not meant to have multiple indexes. For example, the demo person "Antonio de Espinosa" should _only_ be indexed under 'D' and not also 'E'. Currently, the PR automatically adds both indexes which may be incorrect.
> 
> ANSI ARMA's Alphabetic Filing Rules mentions as much, but I don't have a [recent edition](https://webstore.ansi.org/standards/arma/ansiarma122005) to reference. See UA Library's older edition of [Rules for alphabetical filing](https://arizona-ua.primo.exlibrisgroup.com/permalink/01UA_INST/1h3gjm1/cdi_globaltitleindex_catalog_34030032).
> 
> > "Names with prefixes, such as De, Des, Du, La, Le, Les, Mac, Mc, San, VAn, Vanden, Vander, Von, etc., are filed as one unit regardless of whether the name is written as one word or two."

Thanks @joshuasosa for taking a look at this! The [IFLA rules](https://www.ifla.org/g/cataloguing/names-of-persons/) specify that this is a country-specific choice and names in one language might or might not treat a particle as being one unit with the name.

Some glossaries file people under both the particle and and the surname, with the one or the other being more of a notation that the actual record exists under the other heading. (e.g. Gogh, van. -> see. van Gogh)

I'd be okay with either approach. Using only the leading letter would certainly make the code more readable I think. I mostly implemented this version first because it was the more complicated case. I think using only the leading letter would certainly be reasonable.

tadean:
While there are names where alphabetizing by the prefix is not used (I think Dutch is one example), they can override this behavior if necessary by just selecting the letter they want. To me that suggests that maybe @joshuasosa 's suggestion of using the leading letter makes sense, for a default value.

tadean:
> it might still be more common to keep them together as a single unit.

Agreed - it seems the most common case to keep them together. Thanks @joshuasosa !

modules/custom/az_person/az_person.module
```diff
@@ -146,3 +148,41 @@ function az_person_theme() {
...
+ * Implements hook_ENTITY_TYPE_presave().
+ *
+ * Prepopulate the alphabetical index letter field in az_person where available.
+ */
+function az_person_node_presave(EntityInterface $entity) {
+  if ($entity->bundle() === 'az_person'
+    && $entity instanceof NodeInterface
+    && $entity->hasField('field_az_alphabetical_index')
+    && $entity->hasField('field_az_lname'))
+  {
```
joeparsons, line 162: Seeing a PHPCS annotation now.  Apparently my suggestion about readability for multiline conditions wasn't a good idea.  Sorry about that.
```suggestion
    && $entity->hasField('field_az_lname')) {
```

tadean:
@joshuasosa I've updated the matching to consider only the leading character of the last name as a potential index letter. Thanks for your suggestions!

joeparsons (approved):

PR #4473, opened 2025-07-17
Closes #4471 Change photo gallery grid modal image style to az_natural

### LINKED ISSUE #4471: Change photo gallery grid modals to use az_natural instead of az_large
## Motivation
The photo gallery grid is a nice way to easily show images which can be clicked to potentially show a larger version. But in some cases, the image may be smaller than az_large which ends up upscaling the image and adds pixelation.

The modal image in photo gallery grids may upscale an image when the image is below the 1140px width that az_large sets. This upscaling is not very good and can add a lot of pixelation if the original image is much smaller.

Somewhat related: #1953

Example of the issue:

Here is the original image we have of an event in 1997:
Image

We have no way to get larger versions of these images, of which there are many.

Here is how the image ends up looking in the photo gallery modal:
Image

The image becomes very pixelated and distorted.

Change photo gallery to use az_natural instead of az_large.

Here is how the image looks using az_natural:
Image

It's preferable to keep the image looking how it originally looks from this image set rather than apply bad upscaling.

Get larger versions of the images from the original photographers. Takes a lot of administrative work to track down who took what photos (e.g., in galleries that have images from multiple photographers) and investigate contact options to request larger or original photos. They may not even have the original photos anymore since it's been decades since the event took place.

Manually upscale images with another program like Photoshop. Can take a lot of administrative work to process many old photos, and the result may still not look good anyways.

Add an option to allow disabling of the modal for photo gallery grids. The gallery slider currently doesn't have any modal available for clicking on the image. However, with photo gallery grids using az_card for thumbnails, pictures may get cut off and it may be better to retain the clickable option to view the full version. Along those lines, it may be good to enable slider image modals or make it an option.

Change the action when clicking on an image to use a different library (e.g. Fancybox) instead of the basic Bootstrap carousel modal.

Example with Fancybox:
Image

Add a customization in a custom theme/module to change to using az_natural. This is what we currently do.

joshuasosa:
## Description
Changes the photo gallery grid modal image style to az_natural instead of az_large.

## Related issues
Closes #4471 

## How to test

1. Install a new site with demo content and visit the Photo Galleries demo page.
2. Click on an image in the photo gallery grid format.
3. See that the image in the modal slider continues to work and looks ok compared to previous.

https://7b5e2985-cedf-4a13-ab91-3a35916c835d--pr-4473.probo.build/pages/photo-galleries

Note: AZQS/Probo dev builds now uses Bootstrap 5 by default which has photo gallery issues fixed in #4404. To test, AZ Barrio should set AZ Bootstrap CSS to stable.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [x] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


bberndt-uaz:
@joshuasosa Looking at the Probo build, the natural size image is not being resized to fit the container, so the caption is also not visible:

image
<p></p>

However, I think the change in this PR should work by itself with the updated Arizona Bootstrap 5 Photo Gallery component. What would you think if I incorporated this code change into my Photo Gallery PR?
 - #4404

Then this PR could target the 2.x branch (for the 2.14 release) and some additional changes could be made to support the Arizona Bootstrap 2 version of the Photo Gallery. (It seemed like adding `mx-auto` with a bunch of `h-100`s might be one option.)

bberndt-uaz:
> The screenshot is using the Bootstrap 5 styling. Please check it with Bootstrap 4. I updated Probo to use AZ Bootstrap stable instead. Doing so shows that it does fit itself to the container. It likely won't look good for BS5 until #4404 is merged.
> 
> I agree that this should be targeted to 2.14 if it's an agreeable change to add there. If not, such that this proposed change should just be in 3.x, then this PR can either be listed as blocked by #4404 or closed and you could include it in that one if you're ok with it.
> 
> #4474 may have a similar issue.

@joshuasosa Thanks! I had tried briefly with the stable AZ Bootstrap version but I thought I was still seeing an issue. In any case, I see it absolutely works as expected: leaving smaller images at their natural size and scaling down large images as needed.

I think we can go ahead and merge this and then backport the change to the 2.x branch as well.

bberndt-uaz (approved):

joeparsons (approved):

Issue #3387, opened 2024-04-30
When adding a webform, I can't edit that section of the page


danahertzberg:
## Problem/Motivation

I added a webform to a block, then added that block to a text area in a page. I want to edit that text area in the page.

### Describe the bug

After adding the block and saving the page, I am unable to edit the page and edit the paragraph type containing the block. 

Seemingly, there may be an issue with the Submit button of the webform and Edit button of the paragraph type?

### To Reproduce

1. Add a webform
2. Embed that webform within a block
3. Embed that block on a page
4. Save the page
5. Edit the page
6. Click Edit on the paragraph type with the webform block in it

## Proposed resolution

Provide a solution that modifies only the edit interface of a page.

### Expected behavior

I should be able to edit page content even if it has an embedded webform on it.

## Examples

- https://dates-sasg-registrar.pantheonsite.io/new-dates-deadlines
- https://dev-azs-capla.pantheonsite.io/js-test-1

bberndt-uaz:
Related issue: [#2328 Support embedding of Webforms](https://github.com/az-digital/az_quickstart/issues/2328)

danahertzberg:
Need to add a display for the preview 
`/admin/structure/block-content/manage/capla_webform/display`

Similar to how previews for views work
`https://github.com/az-digital/az_quickstart/blob/1b175ea19531325f7585d585556d2ce751867cf7/modules/custom/az_paragraphs/az_paragraphs_view/config/install/core.entity_view_display.paragraph.az_view_reference.preview.yml#L5`

danahertzberg:
Brian implemented a fix for Registrar which may be useful here as well if it can be done in the same way.
https://github.com/uaz-web/registrar_dates/blob/main/src/Form/RegistrarDatesForm.php#L72-L75

danahertzberg:
Enabled the following modules
- Webform
- Webform Node
- Webform UI

PR #4509, opened 2025-07-23
Core fix for issue-3387

zsaenz:
## Description
Fixes the issue but discussion needed on best placeholder.
-The issue was that embedded webforms, when rendered, also rendered their submit button. This caused every connected button with 'type="submit"' (includes page save, edit, and collapse buttons) to attempt to submit the webform, which caused javascript and AJAX errors. When AJAX is on in the webform, the edit page is inoperable. With AJAX off, clicking those buttons actually submits the webform.

After discussion, we decided the easiest thing to do is to prevent rendering the webform content in the editor. By accident I also found that the issue is only prevelant in the collapsed view; rendering the webform with the editor open is ok.

## Related issues

https://github.com/az-digital/az_quickstart/issues/3387
## How to test

Follow steps in related issue to test.

- **Patch release changes**
   - [x ] Bug fix
- **Minor release changes**
   - [ ] New feature
   - [x ] Breaking or visual change to existing behavior
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/az_core.module
+++ b/modules/custom/az_core/az_core.module
@@ -529,3 +529,17 @@ function az_core_preprocess_container__text_format_filter_help(&$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_entity_embed_container().
+ * 
+ * Prevents double-submit button ajax errors in edit pages
+ * by swapping embedded webforms with a custom preview when in collapsed mode.
+ */
+function az_core_preprocess_entity_embed_container(array &$variables) {
+  if(isset($variables["element"]["entity"]["webform"])) {
+    if(\Drupal::routeMatch()->getRouteName() === 'entity.node.edit_form') {
+      $variables["children"] = "Webform Content (Preview)";
+    }
+  }
+}


zsaenz:
Questions:
-Is route the best way to get if on edit page?
-What should the preview content look like? Should we 'fake' a webform or submit button?
-Needs testing on different kinds of embedded content to make sure it shows up properly

modules/custom/az_core/az_core.module
```diff
@@ -529,3 +530,17 @@ function az_core_preprocess_container__text_format_filter_help(&$variables) {
...
+/**
+ * Implements hook_preprocess_entity_embed_container().
+ *
+ * Prevents double-submit button ajax errors in edit pages
+ * by swapping embedded webforms with a custom preview when in collapsed mode.
+ */
+function az_core_preprocess_entity_embed_container(array &$variables) {
+  if (isset($variables["element"]["entity"]["webform"])) {
+    if (\Drupal::routeMatch()->getRouteName() === 'entity.node.edit_form') {
+      $variables["children"] = Markup::create("<hr /><p><h3>Webform Preview</h3></p><hr />");
```
joeparsons, line 543: I think technically we shouldn't be using self-closing elements (XHTML) in since we're using a HTML5 doctype.

```suggestion
      $variables["children"] = Markup::create("<hr><p><h3>Webform Preview</h3></p><hr>");
```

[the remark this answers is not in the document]
joeparsons: > Should the webform preview string be the label of the webform?

I like that idea.

modules/custom/az_core/az_core.module
```diff
@@ -529,3 +530,18 @@ function az_core_preprocess_container__text_format_filter_help(&$variables) {
...
+/**
+ * Implements hook_preprocess_entity_embed_container().
+ *
+ * Prevents competing submit button lockup errors in edit pages
+ * by swapping embedded webforms with a custom preview when in collapsed mode.
+ */
+function az_core_preprocess_entity_embed_container(array &$variables) {
+  if (isset($variables['element']['entity']['webform'])) {
+    $embedded_webform_preview = $variables['element']['entity']['title'][0]['#context']['value'];
+    if (\Drupal::routeMatch()->getRouteName() === 'entity.node.edit_form') {
```
trackleft, line 543: we should add the route for editing content blocks as well, I'm sure people will want to embed via block.

joeparsons: > we should add the route for editing content blocks as well, I'm sure people will want to embed via block.
Good catch.

modules/custom/az_core/az_core.module
```diff
@@ -529,3 +530,18 @@ function az_core_preprocess_container__text_format_filter_help(&$variables) {
...
+      $variables["children"] = Markup::create('<hr><p><h3>' . $embedded_webform_preview . ' (Preview)</h3></p><hr>');
```
tadean, line 544:
```suggestion
      // Get the view builder for the entity.
      $entity = $variables["element"]['#entity'];
      $type_id = $entity->getEntityTypeId();
      $builder = \Drupal::entityTypeManager()->getViewBuilder($type_id);
      // Render the entity as a teaser.
      $variables['children'] = $builder->view($entity, 'teaser');
```
One option would be rendering the entity as a teaser instead of the full view for the edit form.

modules/custom/az_core/az_core.module
zsaenz: I tested with embedded webforms on both nodes and blocks, it seems to do both as is (unless there's some other setup I'm missing)

joeparsons:
Seems like the current implementation relies on the Weboform content type (node type) provided by the optional `Webform Node` submodule (or a similar content type) existing since the existing `Content` embed button we have in our editor config is only for embedding node entities.

Perhaps we should also consider adding an additional entity embed configuration (button) that can be used to embed webform entities directly (instead of needing to embed a node containing a webform)?  (Could probably be a follow-up task).

zsaenz:
> Perhaps we should also consider adding an additional entity embed configuration (button) that can be used to embed webform entities directly (instead of needing to embed a node containing a webform)? (Could probably be a follow-up task).

Posting an issue for this. This can also be related to Slate forms.

danahertzberg (approved):
Please update the testing steps on this PR so they can be referenced in the release notes when merged.

trackleft (approved):

joeparsons (approved):

PR #4543, opened 2025-07-28
Fixes #4542 Resolve Profiles API Importer lock for failed imports

### LINKED ISSUE #4542: Profiles API Importer can lock people subsequently created
### Describe the bug
The profiles API importer has a feature that locks some fields of the edit form for people importer from the API. It does this by mechanism of checking for the presence of the NetID in the `migrate_map_az_person_profiles_import` table.

### To Reproduce
Import a person that is not in the profiles API.
Receive message that the person could not be imported.
Create the person manually by hand
See that their node is locked after saving because it appears in the map table with a null destination id.

## Proposed resolution
The functionality for locking a person should be dependent on the fact that the person was imported, **and the node they were imported as IS the node being edited**. The second half of that check currently does not occur.

### Expected behavior
Nodes should only be locked if they were actually imported by the API.

tadean:
## Description
An individual that was attempted to be imported via the profiles API (and fails) can lock a subsequently created node. This is because the functionality only checks that an import happened (potentially with a null destination in the case of a failed import). It should instead check if the mapped node is the one attempting to be edited.

## Related issues
Closes #4542 

## How to test

- drush en `az_http az_person_profiles_import`
- Login to site
- Visit `/admin/config/az-quickstart/settings/az-person-profiles-import` and add your API token
- Visit `/admin/content/profiles/import`
- Import a combination of people, some present in the API and some not
- For the people not present in the API, verify that the migration emits a warning that they were not found
- After the import, create nodes manually for the people not imported (**make sure that you list the netID you attempted to import them with**)
- Attempt to edit both the imported people, and the people that were created with the same names as the people that failed to import
- Verify that the imported people have locked fields, and that the manually-created people do not

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_person/az_person.module
+++ b/modules/custom/az_person/az_person.module
@@ -61,7 +61,8 @@ function az_person_form_node_az_person_edit_form_alter(&$form, FormStateInterfac
           // Migration did not exist, or migrate service not found.
           // We have no data on this person being imported or not.
         }
-        if (!empty($imported)) {
+        // Lock person if migrate map for this person points to them.
+        if (!empty($imported[0]['nid']) && ($imported[0]['nid'] === $node->id())) {
           $person_warning = t('This person has been imported from the Profiles API.');
           \Drupal::messenger()->addWarning($person_warning);
           $disabled_fields = [


joeparsons (approved):
Looks good to me.

Issue #3660, opened 2024-08-22
Make it easier to embed Request for Information (RFI) forms within content.


trackleft:
## Motivation
Many colleges use Slate to help them gather information about their prospective students using embedded RFI Forms.

### Is your feature request related to a problem? Please describe.
At the moment, embedding RFI forms within content is cumbersome and not user-friendly. This makes it difficult for content editors and administrators to incorporate forms seamlessly into their pages.

### Describe the solution you'd like
A user-friendly interface  that allows easy embedding of RFI forms into content and managing them.

### Describe alternatives you've considered
Using third-party embed codes or custom HTML blocks, which are less intuitive for non-technical users.

### Roles and Permissions considerations
A clear and concise description of how each of the following roles would be impacted by this change:
- **Anonymous user**: Should be able to view and submit RFI forms without issues.
- **Authenticated user**: N/A
- **Content editor**: Should have a simple interface to embed and configure RFI forms within content.
- **Content administrator**: Might require additional controls or settings for form management.
- **Administrator**: Should have full control over form embedding, settings, and permissions.

## Additional context
<img width="1345" alt="image" src="https://github.com/user-attachments/assets/22d96323-4814-4520-a054-187b9f05c569">
Added to page by embedding a Flexible Block .
<img width="1017" alt="image" src="https://github.com/user-attachments/assets/a19fb566-2572-495b-a92c-e7067ed276ce">
Full HTML (Requires special permission)
<img width="1611" alt="image" src="https://github.com/user-attachments/assets/18b90a55-883f-49c7-8f34-5b8450880732">

trackleft:
At the moment, the workaround is to create a flexible block. add a full HTML paragraph, Add the RFI embed HTML to the Full HTML paragraph, Embed the flexible block in the page you want the RFI to appear on.

camikazegreen:
Here is a similar example for embedding from Form Assembly (Trellis):
<script src="https://forms-a.trellis.arizona.edu/publish/118" data-qp-target-id="" defer></script>
with additional info here: https://help.formassembly.com/help/javascript-form-publishing

ewlyman:
**tl;dr: Suggest creating new QS paragraph type for script embeds and supporting documentation and/or UI elements. Also, Slate forms interfere with views that use Drupal AJAX if they are on the same page.**

[...]

PR #4860, opened 2025-09-23
Issue #3660 more suggestions.

trackleft:
## Description
In this suggestion, I am commenting out much of the added styles as we want our forms to match university styles.

I also add some JS to remove all inline styles from the forms and any added CSS files from trellis.

I added some JS to remove the privacy link since most sites using QS already have that link.

I added some JS to add the arizona bootstrap spinner for when the form is loading.

I am not really sure what those empty text areas are that are added  now that I strip the styles.

### DIFF
--- a/modules/custom/az_media/az_media_trellis/az_media_trellis.libraries.yml
+++ b/modules/custom/az_media/az_media_trellis/az_media_trellis.libraries.yml
@@ -7,3 +7,8 @@ az-media-trellis:
   dependencies:
     - core/drupal
     - core/drupalSettings
+
+az-media-trellis.styles:
+  css:
+    theme:
+      css/az-media-trellis.css: {}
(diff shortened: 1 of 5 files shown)


trackleft:
Is there any way we can get trellis to remove all of their added stylesheets?

djcelaya:
The empty text areas are the prefill fields. For some reason, these "labels" are implemented as input fields on the Trellis end. I'll investigate as to why these aren't being set anymore but my JS on the original PR should pre-fill these fields appropriately. Their styles were hiding these fields.
Screenshot 2025-09-24 at 9 09 23 AM

djcelaya:
> Is there any way we can get trellis to remove all of their added stylesheets?

I have requested a new test form with styles stripped.

PR #4908, opened 2025-10-07
Closes #4906: Update corner styling of Split Screen paragraphs for Arizona Bootstrap 5

bberndt-uaz:
## Description

Fixes corner styling of Split Screen paragraphs as follows:
 - Content-width paragraphs: the outside corners of the entire paragraph element are rounded
 - Full-width paragraphs: none of the corners are rounded

To implement this, this PR makes these changes to the CSS:
 - Sets all Split Screen paragraph images to `border-radius: 0`
 - Sets the border-radius for content-width Split Screen paragraphs to `.375rem` (the default Arizona Bootstrap border radius) along with `overflow: hidden` to round the corners of the paragraph
 - Minor change to remove `max-width: 100%;` from Split Screen image styling since `.img-fluid` already includes this

### Questions:

1. Currently the mobile display of Split Screen paragraphs is the same as the desktop display, with sharp corners for full-width paragraphs and rounded corners for content-width paragraphs. This matches the mobile display of [Text on Media](https://pr4908-zcylxqxtlzbkdfcwjsvlcato70hzumef.tugboatqa.com/pages/text-media). Is this correct?

## Related issues

 - #4906 

## How to test

Verify that these pages look correct on both mobile and desktop:
 - https://pr4908-zcylxqxtlzbkdfcwjsvlcato70hzumef.tugboatqa.com/pages/split-screen
 - https://pr4908-zcylxqxtlzbkdfcwjsvlcato70hzumef.tugboatqa.com/pages/split-screen-no-sidebar

### Arizona Quickstart (install profile, custom modules, custom theme)
- **Patch release changes**
   - [x] Bug fix
   - [ ] Accessibility, performance, or security improvement
   - [ ] Critical institutional link or brand change
   - [ ] Adding experimental module
   - [ ] Update experimental module
- **Minor release changes**
   - [ ] New feature
   - [ ] Breaking or visual change to existing behavior
   - [ ] Upgrade experimental module to stable
   - [ ] Enable existing module by default or database update
   - [ ] Non-critical brand change
   - [ ] New internal API or API improvement with backwards compatibility
   - [ ] Risky or disruptive cleanup to comply with coding standards
   - [ ] High-risk or disruptive change (requires upgrade path, risks regression, etc.)
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
   - [ ] Patch level release (non-security bug-fix release)
   - [ ] Patch removal that's no longer necessary
- **Minor release changes**
[...]

Issue #4700, opened 2025-08-22
Update blockquote handling for Bootstrap 5 (figure/figcaption attribution)


trackleft:
While upgrading sites from **Arizona Bootstrap 2 (BS4)** to **Arizona Bootstrap 5 (BS5)**, we found that recommended blockquote markup for attributions changed.  

In **BS4**, attribution was often placed *inside* the `<blockquote>` as:  

```html
  <blockquote class="blockquote">
    <p class="mb-0">Quote…</p>
    <footer class="blockquote-footer">Name in <cite>Source</cite></footer>
  </blockquote>
```

See https://getbootstrap.com/docs/4.6/content/typography/#blockquotes
See https://digital.arizona.edu/arizona-bootstrap/docs/2.0/content/typography/#naming-a-source

In **BS5**, the recommended markup is:

```html
<figure>
  <blockquote class="blockquote"><p>Quote…</p></blockquote>
  <figcaption class="blockquote-footer">Name in <cite title="Source Title">Source</cite></figcaption>
</figure>
```
See: https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.0/content/typography/#blockquotes
See: https://getbootstrap.com/docs/5.1/content/typography/#naming-a-source

Quickstart currently uses CKEditor’s built-in Blockquote button, which inserts only a `<blockquote>` and provides no attribution UI, leaving authors without a BS5-compliant way to add attribution. Existing BS4 content in the field may also be using the older pattern, risking inconsistent rendering.

<img width="1608" height="1130" alt="Image" src="https://github.com/user-attachments/assets/d3bb94e8-91e8-4dd1-846e-6a4a7d12705e" />

### Describe the bug
- New content authored through CKEditor cannot easily add **BS5-compliant attribution** (`figure/figcaption`).
- Legacy content using the **BS4 pattern** (`blockquote > footer.blockquote-footer`) may not match BS5 guidance and could look inconsistent.

### Additional context
Possible drupal module solutions: 
- https://www.drupal.org/project/ckeditor_quote
- https://www.drupal.org/project/ckeditor_blockquote_attribution
Discussion on Bootstrap Barrio
https://www.drupal.org/project/bootstrap_barrio/issues/3391100

PR #4909, opened 2025-10-07
Optional: Small change to splitscreen demo content

djcelaya:
In reviewing #4908, I noticed a minor visual inconsistency with the blockquote used in the split screen demo paragraph content.

Before:
Screenshot 2025-10-07 at 4 03 28 PM

After:
Screenshot 2025-10-07 at 4 04 15 PM

In BS5, I guess the blockquote-footer (citation/source) element needs to be outside the blockquote: https://digital.arizona.edu/arizona-bootstrap/v5/docs/5.0/content/typography/#naming-a-source

I think we could get closer to the intending display by wrapping the whole thing in a `<figure>` but when I tried to do that locally, Drupal purged it out. I'm guessing it's not allowed in this text format. Still, at least the text doesn't overlap anymore!

### DIFF
--- a/modules/custom/az_demo/data/az_demo_splitscreen_paragraph.json
+++ b/modules/custom/az_demo/data/az_demo_splitscreen_paragraph.json
@@ -184,7 +184,7 @@
         },
         {
             "id":72,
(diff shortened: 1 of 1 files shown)


bberndt-uaz:
@djcelaya Thanks for your review! I was actually looking at this too. I think this would be best as a separate PR: we might want to add `<figure>` to the allowed HTML tags for CKEditor. We can also add `.border-white` to the `.blockquote-footer` element as well.

See this issue also:
 - #4700

djcelaya:
This is a known issue and is being addressed in #4700.

PR #4790, opened 2025-09-08
Closes #1474: Add Rankings paragraph type

kevdevlu:
## Deck settings

- [x] Rankings per row on desktop: Dropdown: 1, 2, 3, 4
- [x] Ranking content alignment: Dropdown: Left aligned, Center aligned
- [x] Ranking header style: Dropdown: Bold headers, Thin headers
- [x] Clickable rankings: Checkbox
   - [x] If checked, add a shadow (match cards)
     - [x] Hover effect: Checkbox
       - [x] If checked, add hover effect (specified below)
- Additional options
  - [x] Rankings per row on tablet: Dropdown: 1 (default), 2, 3, 4
  - [X] Ranking per row on phone: Dropdown: 1 (default), 2, 3, 4
- [x] Rankings group title color: Dropdown: Arizona Blue (default), Sky, Oasis, Azurite, Midnight, Ash, Dark Silver, Black, White
- [x] Rankings group title: Single line

## Individual ranking settings

- [x] Change accordion to say "Edit ranking" 
   - [x] Change to use "Edit" on each card rather than an accordion (match cards) if page load time is slower
- [x] Ranking type: Dropdown: Standard, Image only
  - If Standard
    - If hover effect _is_ checked
      - [x] Ranking Background with Hover Effect: Dropdown: Chili (default), Arizona Blue, Sky, Cool Gray
        - [x] White text on Chili background, inverse on hover (Chili text on white background)
        - [x] White text on Arizona Blue background, inverse on hover (Arizona Blue text on white background)
        - [x] Midnight text on Sky background, inverse on hover (Sky text on Midnight background)
        - [x] Azurite text on Cool Gray background, inverse on hover (Cool Gray text on Azurite background)
        - [x] Black text on Oasis background, inverse on hover (Oasis text on black background)
      - [x] Ranking Link URL: Single line
    - If hover effect _is not_ checked
      - [x] Ranking background: Dropdown: Chili, Arizona Blue, Sky, Oasis, Azurite, Cool Gray, Warm Gray, White, Transparent
        - [x] These should have the default accessible text color applied
          - If Transparent is selected
            - [x] Transparent should have the source mt-auto class removed so it aligns below the text
            - [x] Ranking Font Color: Dropdown: Black (default), White, Arizona Blue
        - [x] Ranking Link Title: Single line
        - [x] Ranking Link URL: Single line
[...]

PR #4936, opened 2025-10-16
[Branch of PR 4790] Potential solution for cropping images

bberndt-uaz:
This PR is a branch of the `issue/1474` branch as tracked in this PR:

 - https://github.com/az-digital/az_quickstart/pull/4790

This PR makes some updates to handling of image-only ranking cards to ensure that images are cropped to the same size as other cards via CSS. There may be a better solution!

### How to Test
https://pr4936-xcemkwjnfg1tapbsmplmlpcmsocvt7tj.tugboatqa.com/brian-test-page

1. In Chrome DevTools, try duplicating the heading of the first ranking card to see how the cropping of the image is adjusted.
2. Also try deleting the column element for the first ranking card to confirm that the image maintains the correct minimum height.


kevdevlu:
The fix works great on desktop, however, when we open it in mobile (Safari/Chrome), it will not render the image because of the height: 0 !important;

bberndt-uaz:
> The fix works great on desktop, however, when we open it in mobile (Safari/Chrome), it will not render the image because of the height: 0 !important;

@kevdevlu Yes, that is a showstopper for sure. I think we can go ahead and close this pull request.

PR #5187, opened 2026-01-12
Closes #3081: Add total page count to view pagers

### LINKED ISSUE #3081: Add total page count to view pagers
## Motivation
In discussing [improvements to our styling of pagers](https://github.com/az-digital/az_quickstart/issues/1015), @accesswatch mentioned that another improvement we could make to pagers would be displaying the total number of pages to the user (e.g., "Page 2 of 16").

Although the full pager does provide a link to the last page, it does not indicate the number of the last page or how many pages there are in total.

### Describe the potential solution and alternatives

1. Add text before the pager indicating the current and total number of pages (e.g., "Page 2 of 16").
2. Replace "First" and "Last" link text with the page numbers of those pages. Adjust the "Previous" and "Next" links as needed.

No roles or permissions would be impacted by this change.

#### Example AZ News pager
image

zsaenz:
(only tested Full Pager)

It doesn't look like we have access to pages between that are ellipsed, so we can't say grab the length of the pages array. We do have the href of the page length as ?page=10, so we can grab the total pages with the code below, with +1 for arraay testing. Plan to overshoot the slice value for much large pagers after testing, probably at least up to 10 characters (with more 0's at the end). With this, we should be able to get by with a fix updating only the theme Twig file.

`{{ (items.last.href|slice(6,9)+1 }}
`
Notes on proposed solutions:
"First" and "Last" look good and are helpful being a bit bigger for mobile navigation. We might want to keep these buttons for accessibility. We also don't have much room in a typical pager with these buttons.

## Description

Testing options to show total amount of pages. Functionality is there, we just want to style it in the pager for whichever way looks best.

## Related issues

Closes #3081: Add total page count to view pagers

## How to test

This is currently a Twig change.
-Install AZ Demo
-check the /calendar page, which has a large enough pager to test with.
-perhaps also test with a page without a Secondary block on the right, for a full fledged pager.
-Test mini-pagers, if requested.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


themes/custom/az_barrio/templates/navigation/pager.html.twig
```diff
@@ -99,6 +100,15 @@
         {% endif %}
       {% endif %}
       <li class="page-item pager__item--ellipsis{{ hide_ellipsis_class }}">&hellip;</li>
+      {% if ellipses.next %}
+        {# Prevent error by checking for last page again after ellipses #}
+        <li class="page-item">
+          <a href="{{ items.last.href }}" title="{{ (items.last.href|slice(6,10))+1 }}"{{ items.last.attributes|without('href', 'title') }} class="page-link">
+            <span aria-hidden="true">{{ (items.last.href|slice(6,10))+1 }}</span>
```
tadean: It seems like it might be more appropriate to do this in preprocessing than in the template. `(items.last.href|slice(6,10))+1` may not work in all cases or pager types.

tadean:
> Instead of making Twig changes for all pagers, how about using Results Summary on the specific views that would benefit from it?
> In a view display's footer area, add "Result summary" and enter `Page @current_page of @page_count` as the content. <img 

I agree with @joshuasosa comment. Also, if this feature is desired, does it mean we need to reexamine which pagers we are using on our views in the first place? The full pager offers more information than the mini pagers.

joshuasosa:
> I couldn't find a "total page count" in the preprocess. I had to do the same features I did in the single twig update previously, just now as an extra variable that we do in the backend and then pass them to twig.
> 
> Full pager only. Mini pagers unaffected.

I'm not referring to a preprocessor nor a twig change. I'm referring to using the existing tools in the view UI. This suggested change would replace the proposed twig changes and theme code changes with a views configuration change.

Here are the steps to find it using your testing example of /calendar:
1. Visit `/admin/structure/views/view/az_events`
2. On the "Full Calendar" display, find the 'Footer' section and click 'Add'
3. Change the dropdown 'For' to "This page (override)" since we just want to see it on this page
4. Scroll down or search to find the "Results Summary" option and checkmark it
5. Click "Apply (this display)"
6. In the 'Display' box, replace the text with `Page @current_page of @page_count`
7. Click "Apply (this display)"
8. Click Save to save the view
9. Visit the `/calendar` page and see the total page count below the pagination

These steps would be repeated for any view display that needs a total page count.

bberndt-uaz:
> My suggestion above to use views results summary instead of twig/php changes would satisfy the issue's first suggested solution, which is my preference with Troy in agreement. Maybe the issue submitter @bberndt-uaz could add a note about preferred solution?

The solution @joshuasosa proposed sounds good to me, especially since [Drupal already has a default message for this](https://git.drupalcode.org/project/drupal/-/blob/main/core/modules/views/src/Plugin/views/area/Result.php#L25).

The only thing we may need to look at after adding those footers to the view displays is the bottom spacing. The `.pagination` element of the pager gets some bottom margin from Arizona Bootstrap, but I think it would be good to add some bottom margin below the "Page x of y" footer text as well.

modules/custom/az_finder/config/quickstart/views.view.az_events.yml
```diff
@@ -2014,10 +2019,22 @@ display:
...
+        result:
+          id: result
+          table: views
+          field: result
+          relationship: none
+          group_type: group
+          admin_label: ''
+          plugin_id: result
+          empty: false
+          content: '<p><em><div class="pager-summary">Displaying @start - @end of @total</em></p>'
```
zsaenz:
```suggestion
          content: '<p><em><div class="pager-summary">Page @current_page of @page_count</em></p>'
```

zsaenz:
Removed previous work, added a Footer content with 1 rem margin bottom on the new pagination-summary class.

-Which sites already have footers for events or news? Testing needed for existing sites that already have a footer in the targeted views.

-There's no current check to make sure pagination is live. If there's less than 1 page of content, the footer will still show up. We could look at hooks to check for pagination targeting this class specifically (display: none on pagination-summary class), if it's an issue?

-Only added to Events for now. We'll want to add it to the major views that expect pagination, listed below.

Event
- [x] Full Calendar
- [x] Past Grid View with Images
- [x] Past Row View

News

- [x] Paged Row View
- [x] Grid View with Filter

Publications

- [x] Search
- [x] Person
- [x] List

joshuasosa:
The update to add the page count to full calendar footer looks good to me. Thanks!

themes/custom/az_barrio/js/az-hide-single-pager.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * @file
+ * A JavaScript file to hide page count summaries in the footer
+ *  when there is only one page and no pagination.
+ */
+
+(function (Drupal, once) {
```
joeparsons: ESlint fix:
```suggestion
((Drupal, once) {
```

joshuasosa:
Note: There may be a views hook that can be used to check the pagination to disable/unset the newly added results summary when there's only one page.
https://api.drupal.org/api/drupal/core%21modules%21views%21views.api.php/11.x

themes/custom/az_barrio/js/az-hide-single-pager.js
zsaenz: File to be removed. (also weird this didn't show up for me, I ran and checked it multiple times and it passed?)

zsaenz:
Using "view_post_render" hook seems to work, as suggested. This uses the "ViewExecutable" class, which I added to barrio_theme. As a bonus this looks like it'll only unset footers using 'Result Summary' footer types, so folks who may have added footers like "unfiltered text" should be unaffected (check: Grid View with Filter).

Also removes ID from P tag, which should allow us to handle multiple view cases.

themes/custom/az_barrio/az_barrio.theme
```diff
@@ -763,3 +764,17 @@ function az_barrio_preprocess_block__az_navbar_fullscreen_menu_block(array &$var
...
 }
+
+/**
+ * Implements hook_views_post_render_HOOK() for view pagination.
+ *
+ * In a paged view, if pager exists but theres only one page (therefore hidden),
+ * this hides the page count summary in the views footer.
+ */
+function az_barrio_views_post_render(ViewExecutable $variables) {
+  $items_per_page = (int) $variables->pager->options['items_per_page'];
```
tadean: I think it would be good to check if the pager property exists before using it.

trackleft (approved):

danahertzberg (approved):

bberndt-uaz (approved):

trackleft (approved):

PR #5234, opened 2026-01-22
Fixes #5232: With AZ Navbar enabled, hover open does not fire until the page is visited once

### LINKED ISSUE #5232: With AZ Navbar enabled, hover open does not fire until the page is visited once
## Problem/Motivation
Hover opens do not seem to fire when first enabling AZ Navbar.

### To Reproduce

1. Clone AZQS and adjust the theme settings to use AZBS main rather than 5.0.4.
2. Enable AZ Navbar and third level navigation. 
3. Hover over a Dropdown item in main navigation and notice the dropdowns do not expand on hover.
4. Click open the primary dropdown-toggle and navigate to a page.
5. If you haven't visisted the page before on your build, and attempt to hover open its parent dropdown, hover open does not fire.
6. Now, if you refresh the page or navigate back to that same page, hover functionality now works.

## Proposed resolution
1. Uninstall the Big Pipe module.
2. Keep Big Pipe, but provide some mechanism for navbar-az to bypass it.

### Expected behavior

Hover opens should always fire, regardless if I have been to the page or not before.

djcelaya:
## Description
Adds a theme library with a Drupal behavior wrapper that calls arizonaBootstrap.enableAzNavbarHoverDropdowns() and attaches it directly in the AZ navbar menu template so the behavior reattaches whenever the menu markup is streamed.

## Related issues

#5232 

## How to test

1. [Enable AZ Navbar](https://pr5234-ux4sx8nhdjp4oxlzb3mngdyviinfoyio.tugboatqa.com/admin/appearance/settings/az_barrio) if needed at `/admin/appearance/settings/az_barrio`. It would be best to test this in a non-Pantheon environment.
2. Verify open-on-hover functionality works when authenticated.
3. Clear Drupal caches to verify problem doesn't resurface.

Tugboat review environment: https://pr5234-ux4sx8nhdjp4oxlzb3mngdyviinfoyio.tugboatqa.com/

- **Patch release changes**
   - [X] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/themes/custom/az_barrio/templates/navigation/menu--main-navbar-az.html.twig
+++ b/themes/custom/az_barrio/templates/navigation/menu--main-navbar-az.html.twig
@@ -18,6 +18,7 @@
  *   - in_active_trail: TRUE if the link is in the active trail.
  */
 #}
+{{ attach_library('az_barrio/az-navbar-hover') }}
 {% import _self as menus %}
 {#
   We call a macro which calls itself to render the full tree.
--- a/themes/custom/az_barrio/az_barrio.libraries.yml
+++ b/themes/custom/az_barrio/az_barrio.libraries.yml
@@ -25,6 +25,13 @@ az-barrio-off-canvas-nav:
     js/az-barrio-off-canvas-nav.js: {}
   dependencies:
     - az_barrio/global-styling
+az-navbar-hover:
+  js:
+    js/az-navbar-hover.js: {}
+  dependencies:
+    - az_barrio/arizona-bootstrap-js
+    - core/drupal
+    - core/once
 az-proxima-nova:
   css:
     theme:
--- a/themes/custom/az_barrio/js/az-navbar-hover.js
+++ b/themes/custom/az_barrio/js/az-navbar-hover.js
@@ -1,0 +1,20 @@
+((Drupal, once) => {
+  Drupal.behaviors.azNavbarHoverDropdowns = {
+    attach: (context) => {
+      once('azNavbarHoverDropdowns', '.navbar-nav', context).forEach(
+        (navList) => {
+          if (!navList.closest('.navbar-az')) {
+            return;
+          }
+          const azBootstrap = window.arizonaBootstrap;
+          if (
+            azBootstrap &&
+            typeof azBootstrap.enableAzNavbarHoverDropdowns === 'function'
+          ) {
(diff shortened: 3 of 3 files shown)


themes/custom/az_barrio/templates/navigation/menu--main-navbar-az.html.twig
```diff
@@ -18,6 +18,7 @@
  *   - in_active_trail: TRUE if the link is in the active trail.
  */
 #}
+{{ attach_library('az_barrio/az-navbar-hover') }}
```
trackleft: Have you tried just this
```suggestion
{{ attach_library('az_barrio/arizona-bootstrap-js') }}
```

joeparsons (approved):
This is very similar to the fix I independently came up with while looking into this issue locally.  I pushed a branch up with my changes for comparison:
https://github.com/az-digital/az_quickstart/compare/main...issue/5232-joeparsons

We probably ought to look into whether all the Bootstrap JS components we use in Quickstart have similar issues with big_pipe.  If other Bootstrap JS components aren't affected, I think it'd be worth taking a closer look at the custom `NavbarHoverDropdown` JS component in Arizona Bootstrap to see if there is something we could do in that component that makes this unnecessary (i.e. see if we can attach a listener to the `DOMContentLoaded` event directly within the Bootstrap component itself somehow).

I think this fix should be sufficient for the Quickstart 3.2.0 release though 👍🏻.

trackleft (approved):
I also attempted a fix by adding a Lazy Builder based on the approach in this article: https://www.droptica.com/blog/drupal-bigpipe-using-lazy-builders-2023, but I couldn’t get anything working better than the solution implemented in this PR.

I agree with @joeparsons that we should take a closer look at arizona-bootstrap to ensure the code actually behaves the way we expect (and the way we describe it).

tadean (approved):

tadean:
This behavior-based approach seems reasonable.

PR #5309, opened 2026-02-13
Closes #5156: Fix `Update Preview` and `Add Another Item` Reorder Issue for AZ Ranking

### LINKED ISSUE #5156: `az_ranking` re-order preview issue
## How to reproduce 
1. Create a page, Add Rankings
2.  Fill in the first ranking, just use default settings
3. "Add another item" 
4. For the second ranking, do "Image Only" 
5. Upload / pick an image
6. Reorder, by putting the second "Image Only" ranking above the first one.
7. Click on "Update Preview" anywhere or "Add another item". 
    Image

    **Result: The ordering of the preview and the corresponding fields get broken.**
    Image

## Where the Issue Is
In [AZRankingWidget.php](https://github.com/az-digital/az_quickstart/blob/main/modules/custom/az_ranking/src/Plugin/Field/FieldWidget/AZRankingWidget.php), the `formElement()` function handles creation of preview items from the user's input. 

This seems to be triggered when 1 of 3 buttons are clicked: 

Image

- Update Preview
- Add Another Item
- Remove

If I have 3 rankings showing like this: 
```
    Position 1
    [Preview of Ranking 1]
    [Fields of Ranking 1]
    
    Position 2
    [Preview of Ranking 2]
    [Fields of Ranking 2]
    
    Position 3
    [Preview of Ranking 3]
    [Fields of Ranking 3]
```

And I reorder it using the drag and drop so that it looks like this:
```
    Position 1
    [Preview of Ranking 2]
    [Fields of Ranking 2]
  
    Position 2
    [Preview of Ranking 1]
    [Fields of Ranking 1]
    
    Position 3
    [Preview of Ranking 3]
    [Fields of Ranking 3]
```

And then if i click on "Update Preview" or "Add Another Item", it will look like this:
```
    Position 1
    [Preview of Ranking 1]
    [Fields of Ranking 2]
  
    Position 2
    [Preview of Ranking 2]
    [Fields of Ranking 1]
    
    Position 3
    [Preview of Ranking 3]
    [Fields of Ranking 3]
```
Preview of Ranking 1 will show on top of fields of ranking 2. It almost seems like the preview is stuck to its original position even though its corresponding fields have moved: Preview of Ranking 1 is still in position 1, preview of ranking 2 is still in position 2 despite the fact that I've switched the order/position of those items.

If I click on "Remove" (Remove an item) however, somehow that makes everything right again.  The correct previews stick to the corresponding fields. There is something wrong with the way Ajax is working for "Update Preview" and "Add Another Item" that causes the form rebuild to put the previews with the wrong fields, but whatever is affecting those two buttons is not affecting the "remove" button.

kevdevlu:
## Description

Closes #5156 - ranking widget previews getting out of sync with their form fields after drag-and-drop reordering and clicking ajax buttons: "Update Preview", "Add Another Item", or "Remove". 

### The Core Problem

In `formElement()`, we set both `details` (form elements like textfields and selects) and `preview_wrapper` (that renders the preview for the ranking item). The KEY ISSUE is: after `formElement()` returns, Drupal's Form API treats these two children differently:

- **`details`** contains standard form elements (`#type` => `textfield`, `select`, etc.). The Form API automatically populates their `#value` from user input, keyed by the original form delta. **_This means after drag-and-drop reorder, the fields always show the correct data for their row._**

- **`preview_wrapper`** is a plain render array (`#theme` => `az_ranking`). The Form API does not touch render arrays — whatever values are set in `formElement()` remain as-is. The preview was originally built from `$items[$delta]`, **_which could be in a different order than user input during AJAX rebuilds_**.

### The Fix 
If the fields are ALWAYS in the right order after ajax requests, why not build previews using the same Drupal form processing pipeline as the fields?! So instead of fidgeting / playing around with widgetState and trying to hack the deltas / original deltas using some sort of mapping: **WE JUST BUILD THE PREVIEW AFTER the fields have been built.** 

- Use `#after_build` callback: 
    ```php
    $element['#after_build'][] = [$this, 'afterBuildRebuildPreview'];
    ```
    This builds the preview from the same Form API-populated `#value` sources as the form fields. We use this to replace the current function that builds from `$item` (`buildRankingPreview()`).
- The media field required special treatment...since the media library widget's `#value` structure differs from simple form elements.  We just read from `$form_state->getUserInput()` to ensure it matches the same delta order as text fields. 
- Table-drag handles visual row reordering on the client by moving entire table rows (preview + fields together), so no server-side delta remapping is needed (yay!)

## How to test

Make sure `az_paragraphs_rankings` is enabled

- [ ] Create a ranking paragraph with 4+ ranking items (mix of "standard" and "image_only" types)
- [ ] Verify previews render correctly on initial page load
- [ ] Drag-and-drop to reorder items once or multiple times, then click "Update Preview" or "Add Another Item" — verify previews match their corresponding fields
- [ ] Specifically test that image_only rankings keep their images on the correct card after reorder
- [ ] Test collapse/expand of "Edit Ranking" details panels after reorder
- [ ] Test the "Remove" button (on new items and existing items) after reorder
- [ ] Test removing an existing item, and then creating a new item 
- [ ] Save the form and verify the saved order is correct

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


kevdevlu:
A better version of this is implemented in #5834 , taking into consideration Troy's recommendation to use a custom form api element for Rankings instead of using an #after-build callback (this PR).

PR #5364, opened 2026-02-27
Closes #4343 Adds accordion expand/collapse all option

### LINKED ISSUE #4343: Add "Expand all" option for accordions
_TL;DR: When accordions are closed, users can't search the full page contents by using `CRTL+F`. But, when all the accordions are open, the page becomes overwhelmingly long. Let content managers add a button to expand/collapse all panels._

## Motivation
This feature was initially suggested by library colleagues managing pages where one or more accordions are used to organize the content, sometimes with a large amount of information in each accordion panel. 
Content examples: 
- [Federal public access policies page](https://lib.arizona.edu/research/share/federal-policies)
- NSF section on [Example data management plans page](https://data.library.arizona.edu/data-management/data-management-plans/examples).

When lots of content is split up into accordions, it can be easier to find if the user recognizes the categories in accordion headers. However, if a user is less familiar with the topic, or if they want to search for a relevant keyword in the body content, they may not know which accordion section to open.

In this case, users should be able to use `CTRL+F` to search the page contents for a keyword they are familiar with. The problem is: The contents of closed accordion panels can't be searched with `CTRL+F`.

This could be applied to any page using an accordion. It would make sense for the content manager to select whether the page should have an ‘expand all’ button. When an expand all button is present, it can use existing bootstrap classes for styling.

- Add an option to the accordion paragraph type “[ ] Include Expand/Collapse all button” or similar
- When the option is checked, display an expand/collapse button above the accordion when the page is viewed.
- Display something like:

Image

- Don't use an accordion at all: The page was too long; that's why an accordion was used to split up the content in the first place
- Expand all the sections by default; this would have the same effect as not using an accordion; it would be a long page and the way the user could interact with it would be to close each panel that wasn't relevant(?)

- Anonymous user: Can view and interact with the expand/collapse button when they are reading the page content
- Content editor: Can select the option to include an expand/collapse button when they're adding an accordion paragraph to a page

joshuasosa:
## Description
Adds an option in accordion paragraphs to add an "Expand all/Collapse all" button above the related accordion paragraph. The button will expand or collapse all accordion elements in the accordion set when clicked, respective of the button text.

Note: Assisted by Copilot for formatter and javascript functions.

## Related issues

Closes #4343

## How to test

1. Edit the Accordions demo page.
2. Edit the first accordion paragraph and checkmark the new 'Expand/Collapse' option.
3. Save.
4. Click the new 'Expand all' button above the accordion set to see all accordion elements in the set expand. The button text should change to 'Collapse all'.
5. Click the 'Collapse all' button to see all accordion elements in the set collapse. The button text should return to 'Expand all'.

Example using demo content:
https://pr5364-vehwk09c2tmxyxtbgdbjhbkm7jzb3pzh.tugboatqa.com/pages/accordions
image

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


danahertzberg (dismissed):

modules/custom/az_paragraphs/az_paragraphs.libraries.yml
```diff
@@ -68,3 +68,11 @@ az_paragraphs.az_rankings:
...
     - core/drupal
     - core/once
+az_paragraphs.az_accordion:
+  js:
+    js/az_paragraphs_accordion.js: {}
+  dependencies:
+    - az_barrio/arizona-bootstrap-css
+    - az_barrio/arizona-bootstrap-js
+    - az_barrio/button-no-conflict
+    - core/drupal
```
bberndt-uaz: This change, which could be applied along with my next suggestion, adds the Drupal core/once library so we can prevent most of the JS code from running multiple times. In fact, I think the accordionBound variable might be working to prevent multiple event listeners from being attached, but we could potentially remove that variable in favor of the preferred Drupal way of doing things.

```suggestion
    - core/drupal
    - core/once
```

bberndt-uaz (commented):
Here are a couple suggestions for updating the JavaScript.

modules/custom/az_paragraphs/js/az_paragraphs_accordion.js
djcelaya: Unfortunately, I am seeing that this new functionality does not update ARIA labels. See the `aria-expanded` properties in the `.accordion-header` button. I think these need to be updated to "true" or "false" corresponding to whether "Expand all" or "Collapse all" was clicked, respectively.

danahertzberg (dismissed):

djcelaya (commented):
This is a nice addition and I think the more compact toggle works well right-aligned where the visual accordion "controls" are already positioned.

In my local testing, I did notice that there is a chance for a single element accordion to become out of sync with its expand/collapse state. I checked **Show "Expand/Collapse All" button** for both accordions included in the demo content and was looking at the second one which has just one element. If it is expanded on page load, clicking the up chevron does not update the **Collapse all** text to **Expand all**.

Image

Likewise, if it is collapsed by default and I click the down chevron, **Expand all** is not updated to **Collapse all** (since all 1 elements are now expanded).

Image

joshuasosa:
> Originally I was testing on Safari. I just pulled up Firefox on my Mac and am seeing the same behavior.

I have reduced motion on my machine which made me miss this case. Re-enabling motion let me reproduce the issue. I updated it with a different approach. Please test again.

djcelaya (dismissed):
I tested the config and DB updates locally against the demo content and they worked great!

kevdevlu (dismissed):

kevdevlu (approved):
Looking great! https://issue4343-azs-anthropology.pantheonsite.io/undergraduate-scholarships-awards

djcelaya (approved):
I tested again locally with the demo content and everything is looking good. The new transitions are very smooth and a great improvement.

camikazegreen (approved):
Reviewed frontend experience and it looks/works great!

bberndt-uaz (approved):

PR #5444, opened 2026-03-26
Closes #5371: Add FAQ Schema + Checkbox to Accordions

### LINKED ISSUE #5371: FAQ Content Type
## Motivation
With the rise of generative search and bot reliance on schema markup we need a more structured way to easily add FAQ content to pages. 

Often FAQs are added using accordions and individual Q/As may be repeated throughout the site, across multiple pages; this makes content management a nightmare. Another shortcoming of using that method to include FAQs on pages is the lack of schema markup support and again placing the burden of making content updates in multiple places on the content editor. 

The content editor should have the ability to add Question/Answers at the site-level which can be used a la carte inside a FAQ content type available on a basic page. Correctly formatted FAQ schema markup should also be added to a page when FAQs are added, this would not be something the content editor can modify, as it should reference the on-page FAQ content exactly. 

- Anonymous user: Could view the new content on page.

- Authenticated user: Could view the new content on page.

- Content editor: Ability to create/modify/remove FAQs

- Content administrator: Ability to create/modify/remove FAQs

- Administrator: Ability to create/modify/remove FAQs and enable/disable the content type.

kevdevlu:
## Description

Issue #5371. This PR adds a "This is a FAQ" checkbox to accordions and generates the JSON LD block when that checkbox is checked using this 2 step process: 

1. **AZAccordionDefaultFormatter**: Each accordion formatter attaches its questions as a `html_head` entry with a `faq_questions_ENTITY_ID` **key** and JSON data as the **value**. The formatter strips all HTML tags except [those supported by Google](https://developers.google.com/search/docs/appearance/structured-data/faqpage#:~:text=%3Ch1%3E%20through%20%3Ch6%3E%2C%20%3Cbr%3E%2C%20%3Col%3E%2C%20%3Cul%3E%2C%20%3Cli%3E%2C%20%3Ca%3E%2C%20%3Cp%3E%2C%20%3Cdiv%3E%2C%20%3Cb%3E%2C%20%3Cstrong%3E%2C%20%3Ci%3E%2C%20and%20%3Cem%3E). 

2. **AZFaqAggregatorSubscriber**: Reads all `faq_questions_*` entries, merges the questions, and sorts them to match the order as they are displayed on page, builds a single `FAQPage` JSON-LD block, discarding the individual `faq_questions_` entries. Runs before HtmlResponseSubscriber. 

### Other Solutions Investigated:
1. FAQ Module: https://www.drupal.org/project/faq. Not regularly maintained. Last updated in 2024. Hasn't been updated to be compatible with Drupal 11 so we can't easily `composer require` it. And neither stable nor dev versions offer FAQ Schema out of the box. 

2. Instead of the formatter, use `hook_page_attachments()` and `hook_page_attachments_alter`. We should be able to if we were only concerned with accordions directly associated with node, but a big part of the feature is trying to get FAQ schema for embedded accordion flexible blocks. So the formatter is prob the best way here and if we use formatter, our`#attached` entries won't exist yet until after hook_page_attachments_alter:
```
KernelEvents::VIEW
  └─ MainContentViewSubscriber::onViewRenderArray()
       └─ HtmlRenderer::renderResponse()
            ├─ HtmlRenderer::prepare(), line 131
            │    ├─ $page_display->build()   ← page render array built (not yet rendered)
            │    └─ invokePageAttachmentHooks()
            │         ├─ $attachments = []   ← starts empty, NOT the page's #attached
            │         ├─ hook_page_attachments()   ← modules may add to $attachments
            │         └─ hook_page_attachments_alter(&$attachments)   ← our hook would run here
            │              ← faq_questions_* entries DO NOT EXIST yet at this point
            │
            └─ $renderer->render($html), line 162, AFTER prepare() returns
                 └─ AZAccordionDefaultFormatter::viewElements()
                      └─ attachFaqSchema()
                           └─ $element['#attached']['html_head'][] = [..., 'faq_questions_ENTITY_ID']
                                     ← faq_questions_* entries ARE created here

KernelEvents::RESPONSE (priority 10)
  └─ AZFaqAggregatorSubscriber::onRespond()
           ← faq_questions_* entries ARE available here in $response->getAttachments()

KernelEvents::RESPONSE (priority 0)
  └─ HtmlResponseSubscriber::onRespond()
       └─ HtmlResponseAttachmentsProcessor::processAttachments()
                ← converts #attached into rendered <script> tags; faq_questions_* keys are gone after this
```

3. `schema_metatag` / `schema_qa_page` module. Good for user-curated schema. But Ryan Dool says he would prefer if they were dynamically generated. We could potentially compute the JSON-LD at node save time and save it to the metadata fields provided by `schema_metatag`, but changes to an embedded block saved independently of the page node would make the stored JSON-LD on the node stale.

### Release notes
**FAQ Schema Markup for Accordion Paragraphs**

Accordion paragraphs now support an optional **"This is an FAQ"** checkbox in the paragraph behavior settings. When enabled, the accordion's title/body content is automatically structured as [FAQPage JSON-LD](https://developers.google.com/search/docs/appearance/structured-data/faqpage) and injected into the page `<head>`, making the content eligible for Google FAQ rich results.

**How to use:**

*Direct accordion on a page:*

1. Edit any page that uses an Accordion paragraph.
2. In the paragraph's behavior settings, check **"This is an FAQ"**.
3. Save the page. A `FAQPage` JSON-LD block is automatically generated from the accordion's question titles and body text.

*Accordion inside a Flexible Block:*

1. Edit the Flexible Block that contains the Accordion paragraph.
2. In the accordion's behavior settings, check **"This is an FAQ"**.
3. Save the block.
4. Embed the block into a page via CKEditor Entity Embed as normal.
5. The FAQ schema is picked up automatically when the page is rendered — no changes needed on the page itself.

Multiple FAQ-enabled accordions on the same page — including those embedded via flex blocks in CKEditor — are merged into a single `FAQPage` block in authored order, as required by Google's structured data guidelines.

**For existing sites:** This feature adds a new behavior plugin to the `az_accordion` paragraph type. Run Config Distro or `drush config:import` after updating to sync the new configuration into your active database.

**Related links:**
- Issue [#5371](https://github.com/az-digital/az_quickstart/issues/5371)
- [Google FAQPage structured data documentation](https://developers.google.com/search/docs/appearance/structured-data/faqpage)
- [Google Rich Results Test](https://search.google.com/test/rich-results)

## How to test

Whenever the Tugboat is ready:
1. Make some Accordion Flexible Blocks. Check "This is an FAQ". 
2. Make a page and add the Flexible Blocks you just made. Add some other accordions to the node. 
3. Click Save.
4. Confirm that the page has FAQPage type JSON LD block in the `head` that matches the content and order of the FAQ Accordions on the page. Copy and paste it to: https://search.google.com/test/rich-results (click on "Code" tab) to test that it's valid FAQPage schema. 

I've applied this branch to `azs-cybersecurity` multidev: https://iss5371-faq-azs-cybersecurity.pantheonsite.io/faqs. Feel free to verify that the JSON-LD Block shows up correctly.

Some other FAQ live pages that we could potentially test on:
- https://grad.arizona.edu/stackable-certificates/faq
- https://law.arizona.edu/careers/student-resources

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
+++ b/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
@@ -6,6 +6,7 @@ namespace Drupal\az_accordion\EventSubscriber;
 
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityRepositoryInterface;
+use Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem;
 use Drupal\Core\Render\HtmlResponse;
 use Drupal\Core\Routing\RouteMatchInterface;
 use Drupal\node\NodeInterface;
@@ -145,6 +146,12 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
     $order = [];
 
     foreach ($node->get('field_az_main_content') as $item) {
+      try {
+        assert($item instanceof EntityReferenceItem);
+      }
+      catch (\AssertionError) {
+        continue;
+      }
       $paragraph = $item->entity;
       if (!$paragraph instanceof ContentEntityInterface) {
         continue;
@@ -167,6 +174,12 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
             continue;
           }
           foreach ($embedded->get('field_az_main_content') as $nested_item) {
+            try {
+              assert($nested_item instanceof EntityReferenceItem);
+            }
+            catch (\AssertionError) {
+              continue;
+            }
             $nested = $nested_item->entity;
             if ($nested instanceof ContentEntityInterface && $nested->hasField('field_az_accordion')) {
               $order[] = (string) $nested->id();
@@ -183,6 +196,7 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
    * Parses <drupal-entity> tags from CKEditor HTML and returns loaded entities.
    *
    * @return \Drupal\Core\Entity\ContentEntityInterface[]
+   *   Loaded content entities found in the HTML, keyed by UUID.
    */
   protected function parseEmbeddedEntities(string $html): array {
     if (!str_contains($html, 'drupal-entity')) {
--- a/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
+++ b/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Drupal\az_accordion\EventSubscriber;
 
+use Drupal\Component\Utility\Html;
 use Drupal\Core\Entity\ContentEntityInterface;
 use Drupal\Core\Entity\EntityRepositoryInterface;
 use Drupal\Core\Field\Plugin\Field\FieldType\EntityReferenceItem;
@@ -181,7 +182,9 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
               continue;
             }
             $nested = $nested_item->entity;
-            if ($nested instanceof ContentEntityInterface && $nested->hasField('field_az_accordion')) {
+            if ($nested instanceof ContentEntityInterface
+                && $nested->hasField('field_az_accordion')
+                && $nested->access('view')) {
               $order[] = (string) $nested->id();
             }
           }
@@ -198,20 +201,29 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
    * @return \Drupal\Core\Entity\ContentEntityInterface[]
    *   Loaded content entities found in the HTML, keyed by UUID.
    */
-  protected function parseEmbeddedEntities(string $html): array {
-    if (!str_contains($html, 'drupal-entity')) {
+  protected function parseEmbeddedEntities(string $text_area_markup): array {
+    if (!str_contains($text_area_markup, 'drupal-entity')) {
       return [];
     }
-    if (!preg_match_all('/<drupal-entity\b[^>]*>/i', $html, $tag_matches)) {
-      return [];
-    }
+    $parsed_text_area = Html::load($text_area_markup);
+    $xpath = new \DOMXPath($parsed_text_area);
     $entities = [];
-    foreach ($tag_matches[0] as $tag) {
-      if (!preg_match('/\bdata-entity-type="([^"]+)"/', $tag, $type_match) ||
-          !preg_match('/\bdata-entity-uuid="([^"]+)"/', $tag, $uuid_match)) {
+    foreach ($xpath->query('//drupal-entity[@data-entity-type and @data-entity-uuid]') as $node) {
+      try {
+        assert($node instanceof \DOMElement);
+      }
+      catch (\AssertionError) {
         continue;
       }
-      $entity = $this->entityRepository->loadEntityByUuid($type_match[1], $uuid_match[1]);
+      try {
+        $entity = $this->entityRepository->loadEntityByUuid(
+          $node->getAttribute('data-entity-type'),
+          $node->getAttribute('data-entity-uuid'),
+        );
+      }
+      catch (\Exception) {
+        continue;
+      }
       if ($entity instanceof ContentEntityInterface) {
         $entities[] = $entity;
       }
--- a/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
+++ b/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
@@ -183,8 +183,7 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
             }
             $nested = $nested_item->entity;
             if ($nested instanceof ContentEntityInterface
-                && $nested->hasField('field_az_accordion')
-                && $nested->access('view')) {
+                && $nested->hasField('field_az_accordion')) {
               $order[] = (string) $nested->id();
             }
           }
@@ -224,7 +223,7 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
       catch (\Exception) {
         continue;
       }
-      if ($entity instanceof ContentEntityInterface) {
+      if ($entity instanceof ContentEntityInterface && $entity->access('view')) {
         $entities[] = $entity;
       }
     }
--- a/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
+++ b/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
@@ -47,19 +47,19 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
       return;
     }
 
-    // Collect all faq_questions_* entries and their questions.
-    $all_questions = [];
+    // Collect questions grouped by the accordion entity ID encoded in the
+    // attachment key. Grouping lets us sort whole accordions by their 
+    // position on the page.
+    $groups = [];
     $remaining_head = [];
 
     foreach ($attachments['html_head'] as $item) {
       [$element, $key] = $item;
       if (str_starts_with($key, 'faq_questions_')) {
-        // Decode the questions from the JSON value.
+        $entity_id = substr($key, strlen('faq_questions_'));
         $data = json_decode($element['#value'], TRUE);
         if (is_array($data) && !empty($data['questions'])) {
-          foreach ($data['questions'] as $question) {
-            $all_questions[] = $question;
-          }
+          $groups[$entity_id] = $data['questions'];
         }
       }
       else {
@@ -67,34 +67,35 @@ class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
       }
     }
 
-    if (empty($all_questions)) {
+    if (empty($groups)) {
       return;
+    }
+
+    // Sort groups by the DOM order of each accordion wrapper's id attribute.
+    // Non-FAQ accordions that match this pattern are ignored since uksort 
+    // only consults IDs that keyed an FAQ accordion group.
+    preg_match_all('/id="accordion-(\d+)/', $response->getContent(), $matches);
+    $order = array_flip($matches[1]);
+
+    uksort($groups, function ($a, $b) use ($order): int {
+      $pa = $order[$a] ?? PHP_INT_MAX;
+      $pb = $order[$b] ?? PHP_INT_MAX;
+      return $pa <=> $pb;
+    });
+
+    $all_questions = [];
+    foreach ($groups as $questions) {
+      foreach ($questions as $question) {
+        $all_questions[] = $question;
+      }
     }
 
     // Build the single merged FAQPage schema.
     $faq_schema = [
       '@context' => 'https://schema.org',
       '@type' => 'FAQPage',
+      'mainEntity' => $all_questions,
     ];
-
-    // Sort questions to match their visual order on the page.
-    // Decode HTML entities in the response so that encoded characters
-    // (e.g. &#039; for apostrophes) are converted back to their raw form,
-    // allowing a direct match against the raw question titles.
-    $html = html_entity_decode($response->getContent(), ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
-    usort($all_questions, function (array $a, array $b) use ($html): int {
-      $pos_a = mb_strpos($html, $a['name']);
-      $pos_b = mb_strpos($html, $b['name']);
-      if ($pos_a === FALSE) {
-        $pos_a = PHP_INT_MAX;
-      }
-      if ($pos_b === FALSE) {
-        $pos_b = PHP_INT_MAX;
-      }
-      return $pos_a <=> $pos_b;
-    });
-
-    $faq_schema['mainEntity'] = $all_questions;
 
     // Add the merged schema as a single html_head entry.
     $remaining_head[] = [
--- a/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
+++ b/modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
@@ -17,10 +17,6 @@ use Symfony\Component\HttpKernel\KernelEvents;
  * HtmlResponseSubscriber (which processes attachments into HTML), finds all
  * such entries, merges the questions, and replaces them with a single
  * 'faq_schema' entry containing one FAQPage JSON-LD block.
- *
- * This approach is compatible with Drupal's render caching because #attached
- * metadata survives caching — the subscriber sees entries from both cached
- * and freshly rendered accordions.
  */
 class AZFaqAggregatorSubscriber implements EventSubscriberInterface {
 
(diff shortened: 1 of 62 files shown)


kevdevlu:
Feedback from az digital meeting today:
- Cameron and Ryan: the current approach gives immediate ability to turn already-in-use accordions to use FAQPage schema. Other approaches seem marginally beneficial
- Josh suggested FAQ Content type: https://www.drupal.org/project/faq
    - Some examples from LPL: 
    - https://intranet.lpl.arizona.edu/faq/it-services
    - https://intranet.lpl.arizona.edu/faq/new-lpl
    - https://intranet.lpl.arizona.edu/faq/graduate-student
    - https://spacewatch.lpl.arizona.edu/faq 
- Chris liked the event subscriber approach, also suggested to look into https://git.drupalcode.org/project/schema_metatag/-/tree/8.x-2.x/schema_item_list?ref_type=heads
- Joe suggested checking for existing event subscribers instead of modifying the raw response.

camikazegreen (dismissed):
I tested this on Tugboat for the following:

- No regressions on existing accordions
- Checking the FAQ box does not cause any regressions
- No additional schema added when box is not checked
- Checking the FAQ box for one accordion adds that schema markup
- Checking the FAQ box on multiple accordions adds all accordions to the schema markup.
- The schema markup appears to be formatted correctly when viewing source:
view-source:https://pr5444-wgq7bvxavbxh7qsisuihmis3hajenmpv.tugboatqa.com/pages/accordions

modules/custom/az_accordion/src/Plugin/Field/FieldFormatter/AZAccordionDefaultFormatter.php
```diff
@@ -123,7 +135,67 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+      $body = $item->body ?? '';
+
+      if (empty($title) || empty($body)) {
+        continue;
+      }
+
+      // Keep only the HTML tags that Google displays in FAQ rich results.
+      // @see https://developers.google.com/search/docs/appearance/structured-data/faqpage
+      $allowed_tags = '<h1><h2><h3><h4><h5><h6><br><ol><ul><li><a><p><div><b><strong><i><em>';
+      $clean_body = strip_tags($body, $allowed_tags);
```
joeparsons: We could potentially use Drupal's `Xss:filter()` utility function instead of `strip_tags()` here which does some additional checking to make sure that the output is valid / well-formed (see: [Xss:filter() documentation](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Xss.php/function/Xss%3A%3Afilter/11.x))

```suggestion
      $allowed_tags = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'br', 'ol', 'ul', 'li', 'a', 'p', 'div', 'b', 'strong', 'i', 'em'];
      $clean_body = Xss::filter($body, $allowed_tags);
```

(Would also require adding `use Drupal\Component\Utility\Xss;` at the beginning of this file)

kevdevlu: ah a cool utility function

modules/custom/az_accordion/src/Plugin/Field/FieldFormatter/AZAccordionDefaultFormatter.php
```diff
@@ -123,7 +136,67 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+      $title = $item->title ?? '';
+      $body = $item->body ?? '';
+
+      if (empty($title) || empty($body)) {
+        continue;
+      }
+
+      // Keep only the HTML tags that Google displays in FAQ rich results.
+      // @see https://developers.google.com/search/docs/appearance/structured-data/faqpage
+      $allowed_tags = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'br', 'ol', 'ul', 'li', 'a', 'p', 'div', 'b', 'strong', 'i', 'em'];
```
joeparsons:
```suggestion
      $allowed_tags = [
        'h1',
        'h2',
        'h3',
        'h4',
        'h5',
        'h6',
        'br',
        'ol',
        'ul',
        'li',
        'a',
        'p',
        'div',
        'b',
        'strong',
        'i',
        'em',
      ];
```

joeparsons (dismissed):

accesswatch:
**Accessibility review — no blocking issues, one minor hardening suggestion**

The "This is an FAQ" checkbox is properly labeled via Drupal's Form API (`#title`, `#description`), and the description copy does the right thing by discouraging misuse: *"Only use this for content that is genuinely a list of frequently asked questions."* The `Xss::filter()` allowed-tags list aligns with what Google accepts in FAQ rich results, and the JSON-LD is emitted as a `<script type="application/ld+json">` block — invisible to users and leaving the interactive accordion `aria-expanded` states entirely unchanged. ✓

**One minor thing worth confirming:**

`$item->title` is written directly to the schema `name` field without sanitization:

```php
'name' => $title,
```

`Question.name` in schema.org expects plain text — any HTML in the title would land verbatim in the JSON-LD (invalid markup) and could appear as raw tags in Google's rich results panel. A defensive `strip_tags()` would prevent that:

```diff
-'name' => $title,
+'name' => strip_tags($title),
```

Accordion titles are typically plain text so this is unlikely to surface in practice, but it's a one-liner that closes the gap. Small enough for this PR if you agree, or filed as a follow-up.

kevdevlu:
> ```diff
> -'name' => $title,
> +'name' => strip_tags($title),
> ```
Thanks Jeff and Copilot! i changed it :)

joeparsons (dismissed):

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
```diff
@@ -0,0 +1,114 @@
...
+    $faq_schema = [
+      '@context' => 'https://schema.org',
+      '@type' => 'FAQPage',
+    ];
+
+    // Sort questions to match their visual order on the page.
+    // Decode HTML entities in the response so that encoded characters
+    // (e.g. &#039; for apostrophes) are converted back to their raw form,
+    // allowing a direct match against the raw question titles.
+    $html = html_entity_decode($response->getContent(), ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
```
trackleft: This is interesting, but I think you could probably use a drupalSettings array to accomplish this without having to parse HTML.

kevdevlu:
> ```php
> # AZFaqAggregatorSubscriber.php
> // Decode HTML entities in the response so that encoded characters
> // (e.g. &#039; for apostrophes) are converted back to their raw form,
> // allowing a direct match against the raw question titles.
> $html = html_entity_decode($response->getContent(), ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');
> ```
> This is interesting, but I think you could probably use a drupalSettings array to accomplish this without having to parse HTML.

This line that Chris pointed out deals strictly with sorting the questions to match the order they appear on the page. 

Chris and I been talking. He gave me a sub-branch as an example of using `drupalSettings`: https://github.com/az-digital/az_quickstart/pull/5526/changes. The sub-branch produces the same JSON-LD block, using the same `$attachments` array (but with drupalSettings) and aggregated slightly differently in the Event Subscriber.  BUT the problem with order still remains. There's no good way to sort accordions to match their order on the DOM without reading the DOM / "parsing the HTML". 

BUT Chris's concern with parsing html like `$html = html_entity_decode($response->getContent(), ENT_QUOTES | ENT_SUBSTITUTE, 'UTF-8');` is valid... The html_entity_decode (decoding an entire page) can get expensive for just doing a `mb_strpos`. Also comparing user-entered strings just feels hacky / brittle depending on what the user entered. I changed this to use regex instead and to look at the Entity ID instead of the Accordion title. The regex matching runs in O(n) instead of O(questions x body_length) for per-question substring search

@trackleft  says he's good with keeping the current way the Formatter adds to `$attachments` and Subscriber aggregates FAQ's, but still waiting on Chris's feedback on the above sorting method

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
```diff
@@ -112,4 +124,85 @@ public function onRespond(ResponseEvent $event): void {
...
+  protected function buildAccordionOrder(): array {
+    $node = $this->routeMatch->getParameter('node');
+    if (!$node instanceof NodeInterface || !$node->hasField('field_az_main_content')) {
+      return [];
+    }
+
+    $order = [];
+
+    foreach ($node->get('field_az_main_content') as $item) {
+      $paragraph = $item->entity;
```
joeparsons, line 157: Adding an `assert()` to ensure `$item` is a `EntityReferenceRevisionsItem` or `EntityReferenceItem` here would probably make PHPStan happy.

trackleft (dismissed):
Better than what currently exists IMHO, Thanks

danahertzberg:
Can you add release notes to the description of this PR please?

kevdevlu:
> Can you add release notes to the description of this PR please?

Did it, thanks!

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+    if (!preg_match_all('/<drupal-entity\b[^>]*>/i', $html, $tag_matches)) {
+      return [];
+    }
+    $entities = [];
+    foreach ($tag_matches[0] as $tag) {
+      if (!preg_match('/\bdata-entity-type="([^"]+)"/', $tag, $type_match) ||
+          !preg_match('/\bdata-entity-uuid="([^"]+)"/', $tag, $uuid_match)) {
+        continue;
+      }
+      $entity = $this->entityRepository->loadEntityByUuid($type_match[1], $uuid_match[1]);
```
tadean: [loadEntityByUuid()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Core%21Entity%21EntityRepositoryInterface.php/function/EntityRepositoryInterface%3A%3AloadEntityByUuid/11.x) can throw `\Drupal\Core\Entity\EntityStorageException` but possibly this case would never occur for an entity type that's embedded since they're embedded by uuid

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+      if ($entity instanceof ContentEntityInterface) {
+        $entities[] = $entity;
```
tadean, line 228: Should it perform an access check on the embedded entity?

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
```diff
@@ -0,0 +1,222 @@
...
+   * Parses <drupal-entity> tags from CKEditor HTML and returns loaded entities.
+   *
+   * @return \Drupal\Core\Entity\ContentEntityInterface[]
+   *   Loaded content entities found in the HTML, keyed by UUID.
+   */
+  protected function parseEmbeddedEntities(string $html): array {
+    if (!str_contains($html, 'drupal-entity')) {
+      return [];
+    }
+    if (!preg_match_all('/<drupal-entity\b[^>]*>/i', $html, $tag_matches)) {
```
tadean: The entity embed module uses [xpath selectors to find these embeds](https://git.drupalcode.org/project/entity_embed/-/blob/8.x-1.x/src/Plugin/Filter/EntityEmbedFilter.php?ref_type=heads#L146).

kevdevlu: ah good point

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
kevdevlu: Definitely wouldn't hurt

modules/custom/az_accordion/src/EventSubscriber/AZFaqAggregatorSubscriber.php
kevdevlu: Okay i'll try to gracefully catch the exception just in case

zsaenz (dismissed):

joeparsons (dismissed):

modules/custom/az_accordion/src/Plugin/Field/FieldFormatter/AZAccordionDefaultFormatter.php
```diff
@@ -123,7 +136,71 @@ public function viewElements(FieldItemListInterface $items, $langcode) {
...
+    $questions = [];
+
+    foreach ($items as $item) {
+      $title = $item->title ?? '';
+      $body = $item->body ?? '';
+
+      if (empty($title) || empty($body)) {
+        continue;
+      }
+
```
tadean, line 188: `$body` isn't rendered yet, so it seems like the text format doesn't get applied before it gets sent to the json schema. This can mean that the HTML contains things the text format doesn't allow, and that any content that gets created by text format filters is missing. (for example the "Covert URLs into links" or embedded media doesn't appear)

Maybe it should be rendered with its text format? (and then still receive the `Xss::filter` step below)

tadean:
@kevdevlu I tested the new changes with the xpath selectors, access check, and the changes look good with both direct paragraphs and embedded elements with accordions.

The only other thing I noticed is that the text that gets included hasn't been rendered yet, so its text format doesn't apply in the case where there is a text format that makes changes to the markup:
[Screencast from 05-01-2026 03:35:05 PM.webm](https://github.com/user-attachments/assets/cd3ee81c-6bdd-462c-803b-f1c5aac9dae7)

kevdevlu:
See: https://pr5444-wgq7bvxavbxh7qsisuihmis3hajenmpv.tugboatqa.com/faq-test

One issue i noticed is our URL's are some times relative URL's, and even though Google says it's okay, maybe we should make sure they're absolute ones here? 

image

tadean:
> See: https://pr5444-wgq7bvxavbxh7qsisuihmis3hajenmpv.tugboatqa.com/faq-test
> 
> One issue i noticed is our URL's are some times relative URL's, and even though Google says it's okay, maybe we should make sure they're absolute ones here?

 @kevdevlu these can be transformed with [Html::transformRootRelativeUrlsToAbsolute()](https://api.drupal.org/api/drupal/core%21lib%21Drupal%21Component%21Utility%21Html.php/function/Html%3A%3AtransformRootRelativeUrlsToAbsolute/11.x) if we'd like to.

kevdevlu:
Converted Relative URL's to Absolute Ones: https://pr5444-wgq7bvxavbxh7qsisuihmis3hajenmpv.tugboatqa.com/faq-test

joeparsons (approved):

tadean (approved):

trackleft (approved):

PR #5485, opened 2026-04-11
Opportunity content type visual changes

danahertzberg:
## Description
These are changes to the default, card, row, and small row content type displays. 
Also includes a change to the card view to put them in a grid style.


zsaenz (approved):
The datetime fix is awesome

PR #5487, opened 2026-04-14
Fixes #5486 Error when fetching nonexistent course subject code

### LINKED ISSUE #5486: Course Import of non-existent subject code causes PHP errors
## Problem/Motivation
`az_courses` can causes PHP errors if a subject code that does not exist is imported. In some cases, these errors can occur on editing pages that have nothing to do with courses, if the migration cache needs to be rebuilt.

### Describe the bug
When migration counts are checked, an invalid course will cause the following PHP error:

`ValueError: XMLReader::XML(): Argument #1 ($source) must not be empty in XMLReader->XML()`

This may be a PHP 8.4 regression.

### To Reproduce

1. visit `/admin/config/az-quickstart/courses`
2. Add a course that does not exist
3. Save courses
4. Attempt to import courses (or edit another page)

## Proposed resolution
Prevent value error from occurring

### Expected behavior
Non-existent courses should not cause PHP errors.

## Additional context
This may be an undocumented [XMLReader::XML()](https://www.php.net/manual/en/xmlreader.xml.php) PHP 8.4 change that newly causes `ValueError` to be raised. That behavior is mentioned for the `encoding`  argument, but not for the `source` argument.

tadean:
## Description
This PR fixes an uncaught `ValueError` that has taken place since PHP 8.0 when the course importer fetches an entire subject code if the subject code does not exist.

## Related issues
#5486 

## How to test

- Enable `az_course` AND `az_http`
- Visit `/admin/config/az-quickstart/courses`
- In **Courses to Import**, list one subject code that does not exist, and one that does (For example, `FCSC` and `FITS` respectively)
- Click **Save Courses List**
- Wait for the form to save
- Click **Save an Import Courses**
- Verify that courses were imported
- Visit the watchdog log at `/admin/reports/dblog`
- Verify that there is a log message for the subject code that did not exist

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_course/src/CourseSearch.php
```diff
@@ -95,34 +98,42 @@ public function fetchOptions($search) {
...
             }
           }
+          $xml->close();
         }
-        $xml->close();
       }
+      catch (\ValueError $v) {
+        $this->logger->error($this->t("Invalid response from Courses API when searching for %search", [
+          '%search' => $search,
+        ]));
```
joeparsons: Would it make sense to also log the error message?

```suggestion
        $this->logger->error($this->t("Invalid response from Courses API when searching for %search: %message", [
          '%search' => $search,
          '%message' => $v->getMessage(),
        ]));
```

modules/custom/az_course/src/CourseSearch.php
```diff
@@ -95,34 +98,42 @@ public function fetchOptions($search) {
...
+      }
+
     }
     catch (RequestException $e) {
-      $this->logger->error("Request exception.");
+      $this->logger->error($this->t("Request exception while searching for courses."));
```
joeparsons:
```suggestion
      $this->logger->error($this->t("Request exception while searching for courses: %message", [
        '%message' => $e->getMessage(),
      ]));
```

modules/custom/az_course/src/CourseSearch.php
tadean: Agreed.

joeparsons (approved):

bberndt-uaz (approved):
I tested with the main branch first and confirmed the issue. Then I tested the PR in Tugboat and verified that the import completed successfully and the log message was created:

> Invalid response from Courses API when searching for FCSC: XMLReader::XML(): Argument #1 ($source) must not be empty

In fact, two identical log messages are created, but I think this is not an issue.

tadean:
> 
> In fact, two identical log messages are created, but I think this is not an issue.

I think this happens because one occurrence is when it's counting the migration rows, and one occurrence is when it's initializing the migration.

trackleft (approved):

PR #5488, opened 2026-04-14
Closes #5441: Enable the Mobile Nav Block to use any menu

### LINKED ISSUE #5441: Enable the Mobile Nav Block to use any menu
## Motivation
The Mobile Nav Block [introduced in Quickstart 3.0.0](https://github.com/az-digital/az_quickstart/pull/4104) only supports the default "Main navigation" menu (`main`). In part, this was intended so that the mobile nav would provide visitors with access to a full menu tree of the site (a single menu). However, Quickstart should support another menu being used for this purpose. In addition, for sites using custom navigation menus at different paths, it may be useful to set up multiple Mobile Nav Block instances as a stopgap solution while a new information architecture for the site is developed. Generally, using a single navigation menu is better in terms of both content management and SEO.

Make the Mobile Nav Block configurable so that it can use a different menu. The "Main navigation" menu will remain as the default setting for the block.

No roles or permissions would be impacted by this change.

bberndt-uaz:
## Description

 - Updates the MobileNavBlock to allow the menu to be configurable instead of hard-coded to the `main` menu. The menu can be selected when configuring the block from Block Layout.
 - Updates MobileNavBlock callbacks to include `menu_id`.
 - Sets the `main` menu as the default menu for block.block.az_barrio_mobilenavblock.yml.
 - Updates the page template to add the Menu button for any MobileNavBlocks in the Navigation Offcanvas region.
 - Adds a hook to az_core to add a flag to the build array for MobileNavBlocks for use in the page template.

### Release notes

```
### Selectable menu for the mobile nav block
The Quickstart Mobile Nav Block can now be configured to use a different menu instead of the Main Navigation menu. Only one Mobile Nav Block should be displayed per page path.
```

## Related issues

 - #5441

## How to test

The goal is to test a site that has multiple navigation menus for different URL paths. Note the navigation menus on these pages as an example:
 - https://medicine.arizona.edu/
 - https://medicine.arizona.edu/family-community-medicine (mobile nav menu will be fixed by this PR)

1. Create a new menu and add a hierarchy of links to it.
2. In Block Layout, hide the existing Mobile Nav Block (az_barrio_mobilenavblock) on certain page paths.
3. In Block Layout, add a new Mobile Nav Block, select the new menu, and configure it to be shown on the same page paths from step 2.
4. Visit pages in a mobile view to confirm that the mobile nav shows different menus depending on the page path.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


mlittler (approved):

accesswatch:
**Draft accessibility note — one structural concern to resolve before going to review**

The feature direction is solid. One issue introduced by allowing multiple Mobile Nav Blocks on a page:

**Duplicate ID breaks ARIA and Bootstrap offcanvas (blocking)**

The offcanvas panel and its trigger are currently hardcoded to a fixed ID:

```html
<!-- trigger -->
<button data-bs-target="#azMobileNav" aria-controls="azMobileNav">Menu</button>

<!-- panel -->
<div id="azMobileNav" aria-label="Mobile navigation">
```

If two Mobile Nav Blocks are placed on the same page (which this PR makes possible), the DOM will contain two elements with `id="azMobileNav"`. This breaks three things at once:

1. **`aria-controls`** points to the wrong element for any button after the first — AT users will land in the wrong panel
2. **`aria-label="Mobile navigation"`** is identical on both panels — screen readers can't distinguish them
3. **Bootstrap's offcanvas** uses `document.getElementById` which stops at the first match, so only the first panel opens regardless of which button is clicked

The fix is to make the ID dynamic per-block instance. The block already has a unique plugin instance ID — using something like `az_mobile_nav_{{ block_id }}` in both the trigger and the panel, and updating `aria-label` to reflect the menu name (e.g. `aria-label="{{ menu_label }} mobile navigation"`) would resolve all three at once.

Happy to sketch out the Twig/PHP changes if helpful.

bberndt-uaz:
@accesswatch Thanks for your comment!

> If two Mobile Nav Blocks are placed on the same page (which this PR makes possible), the DOM will contain two elements with `id="azMobileNav"`.

To clarify a couple things:
- This PR doesn't make any changes related to placing multiple Mobile Nav Blocks on the same page. That is already possible in current Quickstart using Block Layout and it remains possible with this PR. However, it is not intended or supported.
- I don't think it is possible either with current Quickstart or with this PR for the DOM to contain two elements with `id="azMobileNav"`. That single hard-coded element is added in the page template here:

https://github.com/az-digital/az_quickstart/blob/7a9b19e935877762c788cd5347ab6fee3efb9c15/themes/custom/az_barrio/templates/layout/page.html.twig#L111

If you were able to get two `id="azMobileNav"` elements to show up, could you please give the steps you took so I that can reproduce the issue?

djcelaya (dismissed):
I tested this on a multidev of college-of-medicine and followed the testing instructions to check the FCM use case. It worked great!

djcelaya:
I also added a second Mobile Nav Block for Anesthesiology. Navigating around the site works as expected, with the mobile nav blocks being swapped out based on paths.

tadean (dismissed):

joeparsons (dismissed):

trackleft (dismissed):

modules/custom/az_core/az_core.module
```diff
@@ -581,3 +582,12 @@ function az_core_preprocess_entity_embed_container(array &$variables) {
     }
   }
 }
+
+/**
+ * Implements hook_block_build_BASE_BLOCK_ID_alter().
+ *
+ * Adds a build array flag for mobile nav blocks for use in the page template.
```
bberndt-uaz, line 589:
```suggestion
 * Adds a build array flag for mobile nav blocks for use in the page template.
 * This flag may be checked before the block has built by the lazy builder.
```

modules/custom/az_core/az_core.module
```diff
@@ -581,3 +582,13 @@ function az_core_preprocess_entity_embed_container(array &$variables) {
...
+ * This flag may be checked before the block has built by the lazy builder.
```
bberndt-uaz:
```suggestion
 * This flag is checked before the block has been built by the lazy builder.
```

joeparsons (approved):

tadean (approved):

PR #5182, opened 2026-01-09
Closes #3998: Trellis Opportunity Content Type and Importer

zsaenz:
Closes #3998: Non-Academic Trellis Programs

## Description
Adds a new content type—Opportunities, and an importer for it. Two modules, with Trellis Opportunities embedded within the folder.

1. Opportunities is used for things like study abroad, research, job opportunities, scholarships and internships. 

    - This will be used for student opportunities and engagement, showing the different kinds of jobs or financial assistance the college or unit offers or prepares students for. It can also be used in conjunction with course work, as some courses require volunteer work or internships as part of their capstones. Websites will be able to list opportunities in multiple formats, with filters, in case they need more advanced listings. These fit between things like Events and News—some of them have deadlines, some of them don't and are ongoing, some of them do have deadlines and a program date, where the application may happen months before the actual program starts.

2. The second part of this content type is the Trellis Importer. Trellis opportunities are called Programs, which come from Accounts. Similar to the Event Importer, we want Trellis to be able to import content right into Quickstart sites, with links between for ease of use.

 
----- Follow ups or non-critical updates
- [ ] In config > az quicksart settings, Combine both Trellis Hostname fields for events and opportunities into a single page.
- [ ] Also, possibly combine the Recurring Imports into a single page or table, with the type (opp, event) listed in the table. Two buttons, one for create recurring event and one for recurring opp.
- [ ] Search by date in trellis fields (Trellis update)
- [ ] After we migrate to Single Directory Components: make the card view the same height

## Related issues

Closes #3998: Non-Academic Trellis Programs

## How to test
1. Install both `az_opportunity` and `az_opportunity_trellis`

2. Make sure you have the latest Enterprise Attributes (API was recently updated); otherwise you won't see "Program Type" as an import filter
    a. Go to `/admin/config/az-enterprise-attributes-import/settings` and "Save Configurations

3. Go to `/admin/trellis-opportunity-importer`, make sure "Program Type" is present. 

4. Use these search parameters, as Cameron said our first clients for this module would be. You can search for partial phrases and search for only some of them in the "keyword" parameter. You can also play around with other filters like "Program Type" ("On-Campus Job", "Professional Development", "Mentorship" yield good results)

    | Name | Owner NetID | Parent Name |
    | -- | -- | -- |
    | Thrive Peer Mentoring - 1st Year | wbryan | Belonging and Service |
    | Thrive Peer Mentoring - 2nd Year+ | wbryan | Belonging and Service |
    | New Start | wbryan | New Start |
[...]

PR #5506, opened 2026-04-17
Issue/3998 Sub-Branch: Add Results Count and Full Pager to Trellis Opportunity Importer

kevdevlu:
Sub-branch of #5182 

Changes:
- Switch import view pager from mini to full so total pages are displayed
- Set pager total_items from search API ID count before slicing, enabling accurate page count without a separate count query
- Add result summary header showing "Displaying X-Y of Z results"
- Add template override to render result summary between filters and table rather than above the exposed form
- Add empty state message "No results found. Try adjusting your search filters.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.module
+++ b/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.module
@@ -13,6 +13,17 @@ use Drupal\az_opportunity_trellis\Plugin\views\filter\AZOpportunityTrellisViewsA
 use Drupal\az_opportunity_trellis\TrellisHelper;
 
 /**
+ * Implements hook_theme().
+ */
+function az_opportunity_trellis_theme() {
+  return [
+    'views_view__az_opportunity_trellis_import' => [
+      'base hook' => 'views_view',
+    ],
+  ];
+}
+
+/**
  * Implements hook_form_FORM_ID_alter() for node_az_opportunity_edit_form.
  *
  * Adds warning message and disables opportunity detail
--- a/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.module
+++ b/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.module
@@ -11,6 +11,7 @@ use Drupal\Core\Form\FormStateInterface;
 use Drupal\Core\Url;
 use Drupal\az_opportunity_trellis\Plugin\views\filter\AZOpportunityTrellisViewsAttributeFilter;
 use Drupal\az_opportunity_trellis\TrellisHelper;
+use Drupal\views\ViewExecutable;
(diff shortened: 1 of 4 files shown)


accesswatch:
**Accessibility review — two blocking items, both fixable in this PR**

**Impact:** This view uses `use_ajax: true`. When a user changes filters and submits, Drupal replaces the entire view container. Screen reader users will not hear the updated result count or the empty state message unless those elements are marked as live regions — they must manually navigate back to find what changed. This affects any staff member using a screen reader to triage imported opportunities.

---

### 1. Result count needs `aria-live` (blocking)

The new result summary header (`Displaying @start–@end of @total results`) renders inside `<div class="view-header">` in the custom Twig template. Since AJAX recreates this element on every filter operation, it is silent to assistive technology without a live region.

In `views-view--az-opportunity-trellis-import.html.twig`:

```diff
-  {% if header %}
-    <div class="view-header">
-      {{ header }}
-    </div>
-  {% endif %}
+  <div class="view-header" role="status" aria-live="polite" aria-atomic="true">
+    {% if header %}{{ header }}{% endif %}
+  </div>
```

Keeping the element present even when `header` is empty is intentional — if the element is only injected when results exist, screen readers won't announce when a search returns zero results.

---

### 2. Empty state needs the same treatment (blocking)

`'No results found. Try adjusting search filters above.'` appears inside the AJAX-replaced container with no announcement mechanism. Wrapping the rows/empty area in a live region covers both states:

```diff
-  {% if rows -%}
-    {{ rows }}
-  {% elseif empty -%}
-    {{ empty }}
-  {% endif %}
+  <div aria-live="polite" aria-atomic="true">
+    {% if rows %}{{ rows }}{% elseif empty %}{{ empty }}{% endif %}
+  </div>
```

---

### 3. Minor: `<footer>` tag in the template

The template uses `<footer>{{ footer }}</footer>` for the view footer section. Inside a `<div>` it carries no landmark semantics (it won't be treated as `contentinfo`), but it is atypical compared to other Drupal Views templates which use `<div class="view-footer">`. Not a blocker, but worth aligning for consistency.

---

**Pager labels look great** — replacing the bare `‹‹ ››` symbols with First / Previous / Next / Last text is a meaningful improvement for screen readers and keyboard users who rely on visible labels.

**No new issue needed** — both blocking items are small Twig template changes that fit naturally in this PR.

PR #5516, opened 2026-04-20
a11y: video play/pause button — aria-pressed, event-driven state, fix aria-hidden (closes #5515)

### LINKED ISSUE #5515: Text on Media video button: autoplay state mismatch and aria-pressed toggle semantics
## Follow-up from #5505

PR #5505 makes the right structural changes to the Text on Media video play/pause controls. Two accessibility improvements were deferred as follow-up rather than blocking that PR.

---

### 1. Autoplay assumption creates initial state mismatch

The PR initialises the button as "Pause Video" based on the assumption that videos always autoplay on load. If autoplay is blocked — by browser settings, user preference, or muted-but-blocked autoplay policy — the video won't be playing but the button will say "Pause Video". A user relying on a screen reader or keyboard navigation will press the button expecting to pause, but it will instead start playback.

This is a WCAG 2.1 SC 4.1.2 (Name, Role, Value) concern: the accessible name must reflect the actual current state of the control.

Both APIs support a state check on init:
- **Vimeo:** `player.getPaused().then(paused => { ... })`
- **YouTube:** `event.target.getPlayerState()` in `onPlayerReady` — returns `YT.PlayerState.PLAYING` (1) or other states

The fix is to read the actual state after the player is ready and set the button label accordingly, rather than assuming playback started.

---

### 2. Consider `aria-pressed` for toggle semantics (enhancement)

The current approach of changing `textContent` is a valid way to communicate state — NVDA and VoiceOver will re-read the button name after the change. However, the semantic toggle button pattern uses `aria-pressed`, which makes the current state programmatically determinable independent of the visible label:

```js
// On state change:
playPauseButton.setAttribute('aria-pressed', isPlaying ? 'true' : 'false');
```

With `aria-pressed`, the button name can remain static (e.g. "Toggle video playback") while state is communicated separately. This is the ARIA Authoring Practices toggle button pattern. Either approach is acceptable — this is an enhancement, not a failure.

---

## Acceptance criteria

- [ ] After player initialisation, the button label reflects the actual play/pause state rather than assuming autoplay succeeded
- [ ] (Optional) `aria-pressed` used for toggle semantics

## Related

- Follow-up to #5490
- PR: #5505

accesswatch:
## Description

Follow-up to #5505 (closes #5515).

#5505 replaced the two-button pattern with a single `<button>` and fixed `tabindex` on iframes — all the right moves. This PR addresses two issues left open as follow-up:

### 1. Autoplay state mismatch (bug)

The original button started as "Pause Video" under the assumption that autoplay always succeeds. When autoplay is blocked — by browser policy, `az_gdpr_consent`, or user preference (as noted by @kevdevlu) — the video is not playing but the button says "Pause". A screen reader user hears "Pause Video, button" and presses it expecting to pause, but it starts playback instead.

**Fix:** The button now defaults to "Play Video" / `aria-pressed="true"` (conservative: not confirmed playing). The actual player events (`play`/`pause` for Vimeo; `onPlayerStateChange` states 1 and 2 for YouTube) drive all state updates. The button is always accurate regardless of autoplay outcome.

### 2. `aria-hidden=""` removed and `aria-pressed` added (accessibility)

- `aria-hidden=""` on the button had an ambiguous empty-string value — some AT treat it as `aria-hidden="true"`, hiding the only interactive control from the accessibility tree.
- `aria-pressed` is now used for toggle semantics. Screen readers announce: *"Play Video, toggle button, pressed"* (paused state) and *"Pause Video, toggle button, not pressed"* (playing state). State is programmatically determinable independent of the label.
- The redundant `title` attribute is removed — visible text content is sufficient for the accessible name.

## Changes

| File | Change |
|------|--------|
| `media--az-remote-video--az-background.html.twig` | Remove `aria-hidden=""`, remove `title`, add `aria-pressed="true"`, set initial label to "Play Video" |
| `az_paragraphs_az_text_media_vimeo.js` | Event-driven state: `play`/`pause` events sync button; click handler only calls play/pause |
| `az_paragraphs_az_text_media_youtube.js` | `onPlayerStateChange` states 1 and 2 sync button; click handler uses `aria-pressed` |

## How to test

- Visit a Text on Media paragraph with a YouTube or Vimeo background video
- With autoplay **enabled**: button should show "Play Video" briefly, then switch to "Pause Video" once the player fires its play event
- With autoplay **blocked** (e.g., via `az_gdpr_consent` or browser devtools throttling): button should remain "Play Video" — pressing it starts playback; button then switches to "Pause Video"
- Press the button to pause: it should show "Play Video" and `aria-pressed="true"`
- Press again to resume: it should show "Pause Video" and `aria-pressed="false"`
- Verify with a screen reader (NVDA/JAWS/VoiceOver) that the button is announced and state changes are heard

## Related

- Closes #5515
- Follow-up to #5505
- Referenced in #5490

### Arizona Quickstart
- **Patch release changes**
   - [x] Bug fix

### DIFF
--- a/modules/custom/az_paragraphs/az_paragraphs_text_media/js/az_paragraphs_az_text_media_vimeo.js
+++ b/modules/custom/az_paragraphs/az_paragraphs_text_media/js/az_paragraphs_az_text_media_vimeo.js
@@ -45,8 +45,7 @@
           const pWidth = Math.ceil(height * defaultAspectRatio); // get new player width
           const pHeight = Math.ceil(width / defaultAspectRatio); // get new player height
           let widthMinuspWidthDividedByTwo = (width - pWidth) / 2;
-          widthMinuspWidthDividedByTwo =
-            `${widthMinuspWidthDividedByTwo.toString()}px`;
+          widthMinuspWidthDividedByTwo = `${widthMinuspWidthDividedByTwo.toString()}px`;
           let pHeightRatio = (height - pHeight) / 2;
           pHeightRatio = `${pHeightRatio.toString()}px`;
           // when screen aspect ratio differs from video,
@@ -175,7 +174,6 @@
           });
         }
       }
-
       once('vimeoTextOnMedia-init', 'body').forEach(initVimeoBackgrounds);
     },
   };
--- a/modules/custom/az_paragraphs/az_paragraphs_text_media/js/az_paragraphs_az_text_media_vimeo.js
+++ b/modules/custom/az_paragraphs/az_paragraphs_text_media/js/az_paragraphs_az_text_media_vimeo.js
@@ -99,7 +99,7 @@
             parentParagraph.classList.remove('az-video-paused');
             parentParagraph.classList.add('az-video-playing');
             playPauseButton.textContent = 'Pause Video';
-            playPauseButton.setAttribute('aria-pressed', 'false');
+            playPauseButton.setAttribute('aria-pressed', 'true');
           });
 
           // Sync button and class state when video pauses.
@@ -107,7 +107,7 @@
             parentParagraph.classList.remove('az-video-playing');
             parentParagraph.classList.add('az-video-paused');
             playPauseButton.textContent = 'Play Video';
-            playPauseButton.setAttribute('aria-pressed', 'true');
+            playPauseButton.setAttribute('aria-pressed', 'false');
           });
 
           // Set the iframe tabindex to -1 to prevent focus from reaching iframe.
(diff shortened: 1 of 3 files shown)


joshuasosa (approved):

accesswatch:
All 7 Copilot review comments addressed in commit 7c2d422:

- **template**: initial \ria-pressed\ corrected from \	rue\ to \alse\ — video not yet confirmed playing on load
- **youtube.js**: click handler inverted so \ria-pressed='true'\ (playing) calls \pauseVideo()\
- **youtube.js**: \classList.remove('az-video-paused')\ added when entering playing state to prevent both classes coexisting
- **youtube.js**: \onPlayerStateChange\ state 1 now sets \ria-pressed='true'\, state 2 sets \'false'\
- **vimeo.js**: \play\ event now sets \ria-pressed='true'\, \pause\ event sets \'false'\
- **vimeo.js**: click handler inverted so \ria-pressed='true'\ (playing) calls \pause()\

\ria-pressed\ semantics are now consistent across both players: \	rue\ = video is playing, \alse\ = video is not playing. Ready for re-review.

joshuasosa (approved):

PR #5562, opened 2026-04-27
Closes #5557: Add theme option to set the Arizona Header to blue

### LINKED ISSUE #5557: Add theme option to set the Arizona Header to blue
## Motivation
The background color of the Arizona Header is changing from red to blue as part of a university brand change. This change will be implemented as an option in Quickstart in 3.4.0 and as a mandatory change in 3.5.0.

See the companion Arizona Bootstrap issue:
 - https://github.com/az-digital/arizona-bootstrap/pull/1980

Add a theme option to change the Arizona Header from red to blue.

- Administrator: Can toggle the theme option to change the background color of the Arizona Header.

bberndt-uaz:
## Description

 - Adds a temporary AZ Barrio theme setting to set the Arizona Header to blue.
    - Includes a change to rename the "Institutional Logo Settings" fieldset to "Institutional Header Settings".
 - Updates the styling of the desktop search block form and mobile nav header in accordance with the blue Arizona Header.

### Questions

1. To confirm, is it correct that this change is mandatory in 3.5.0 and we will remove the new theme setting in 3.5.0?
   - Update the "Enable blue Arizona Header" setting title to indicate that the setting is experimental. Remove the sentence mentioning that the setting will be removed in 3.5.0: we can communicate about this setting and the change in general using other means.
2. This PR changes the search box magnifying glass in the mobile nav from dark gray to Arizona Blue (when the blue header is enabled). Do we want to keep this change?
   - No, we'll keep the color as dark gray for both red and blue Arizona Headers.
3. This PR sets a focus ring color based on Oasis for the search input field when the Arizona Header is blue. Is this appropriate? Should the mobile nav search input field have the same focus ring color?
   - We will not set the search field focus ring color at this time. [See comment below.](#issuecomment-4346910225)
4. Should we add the `theme-color` meta tag?
   - The [theme-color meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/theme-color) appears to have limited availability. When we introduce dark mode in Arizona Bootstrap, I believe we might prefer to implement the [color-scheme meta tag](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/meta/name/color-scheme) and [CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/color-scheme). However, this meta tag does not appear to be supported by the Metatag module.

### Release notes

```
### Blue Arizona Header setting
In the AZ Barrio theme settings, a new "Enable blue Arizona Header" setting has been added to change the Arizona Header background color from red to blue. This setting should not be enabled on any sites until after the flagship arizona.edu site has launched its redesign in July 2026. After that point, all Arizona Sites will be switched to the blue header and the setting will be removed.
```

## Related issues

 - #5557

## How to test

https://pr5562-aa8wmh075jczjbqvwasqwmhnheefahvb.tugboatqa.com

1. In AZ Barrio settings, check the checkbox for "Enable blue Arizona Header" and save the configuration.
2. Confirm that the blue Arizona Header and mobile nav appear as expected both on desktop and mobile.
3. In AZ Barrio settings, uncheck the checkbox for "Enable blue Arizona Header" and save the configuration.
6. Confirm that the red Arizona Header is unchanged from current Quickstart.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [x] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


joshuasosa:
> To confirm, is it correct that this change is mandatory in 3.5.0 and we will remove the new theme setting in 3.5.0?

I recall there was some discussion about enhancing the ability for the theme to have more customization options, and/or adding more subthemes in the future to support different looks and feels.

I would be interested to know if retaining red as an option could be considered a supported customization, along with blue and possibly gray ([example](https://web.archive.org/web/20110804174732/http://www.library.arizona.edu/)). If so, a select box or radio set would be appropriate.

danahertzberg:
This requirement in 3.5.0 and possible option to disable is currently in discussion with central marketing. Right now, it's an option to enable the blue header (not to be used before Arizona.edu launches). Options associated with this are currently being discussed. Final decision has not been made yet.

themes/custom/az_barrio/css/style.css
```diff
@@ -432,10 +432,30 @@ td.menu-disabled {
...
+
+.az-search-block.az-search-block-blue .search-block-form .input-group .button {
+  color: var(--bs-blue);
+}
+
 .az-search-block .search-block-form .input-group button:hover,
 .az-search-block .search-block-form .input-group button:focus {
-    background-color: #8B0015;
-    color: #ffffff;
+  background-color: #8B0015;
```
djcelaya:
```suggestion
  background-color: var(--bs-chili);
```
Could we use CSS vars with the red block as well? Not being super familiar with this part of the codebase, I immediately went to look up this value: https://marcom.arizona.edu/brand-guidelines/colors

themes/custom/az_barrio/css/style.css
```diff
@@ -432,10 +432,30 @@ td.menu-disabled {
...
+  color: #ffffff;
```
djcelaya:
```suggestion
  color: var(--bs-white);
```

djcelaya (commented):
Great work with this, @bberndt-uaz! I like the oasis focus ring you've introduced:

Image

But I do think it should be consistent across desktop and mobile:

Image

I _think_ my vote would be for oasis–it seems to jive better with the blue direction we're heading in. My concern is that it is less striking than the red we have currently, so it may be more difficult to zero in on the element being focused. Also, it seems like most of our existing focus styles are some variety of red so throwing another blue one in does muddy that a bit more in terms of consistency.

themes/custom/az_barrio/css/style.css
bberndt-uaz: Yes, we definitely can. I was thinking of leaving it since my understanding was that the red Arizona Header will go away in a couple months, but let's go ahead and update these.

danahertzberg:
I agree with the blue focus statements David made. Let's keep the default red outline focus for desktop and mobile. Then it will match the other focus elements. We can always change it separately in the future.

themes/custom/az_barrio/css/style.css
```diff
@@ -432,10 +432,25 @@ td.menu-disabled {
   padding: 0 8px;
 }
 
+.az-search-block.az-search-block-blue .search-block-form .input-group .form-control:focus {
+  border-color: rgba(var(--bs-oasis-rgb), .5);
+  box-shadow: 0 0 0 .25rem rgba(var(--bs-oasis-rgb), .5);
+}
+
```
bberndt-uaz:
```suggestion
```

themes/custom/az_barrio/css/az-barrio-off-canvas-nav.css
```diff
@@ -22,3 +22,18 @@
 #block-az-barrio-offcanvas-searchform .material-symbols-rounded {
   vertical-align: text-bottom;
 }
+
+/* Overrides for blue Arizona Header */
+.offcanvas-header.text-bg-blue + .region-navigation-offcanvas #block-az-barrio-offcanvas-searchform .form-control:focus {
+  border-color: rgba(var(--bs-oasis-rgb), .5);
+  box-shadow: 0 0 0 .25rem rgba(var(--bs-oasis-rgb), .5);
+}
+
```
bberndt-uaz:
```suggestion
```

themes/custom/az_barrio/css/az-barrio-off-canvas-nav.css
```diff
@@ -22,3 +22,13 @@
 #block-az-barrio-offcanvas-searchform .material-symbols-rounded {
   vertical-align: text-bottom;
 }
+
+/* Overrides for blue Arizona Header */
+.offcanvas-header.text-bg-blue + .region-navigation-offcanvas #block-az-barrio-offcanvas-searchform .button {
+  color: var(--bs-blue);
+}
+
+.offcanvas-header.text-bg-blue + .region-navigation-offcanvas #block-az-barrio-offcanvas-searchform .button:focus {
```
djcelaya, line 32: This is looking good! The only inconsistency I found was related to the submit search button in the mobile view. When the button is focused in the blue version, the background and text color change (I think via this added rule). However, doing the same with the red version produces no visual indication that the button is focused.

Image

I think my vote would be to keep the blue focus rule you've added and add a companion red rule for this focus interaction - this would make red more consistent between the desktop and mobile experience I believe.

Image

What do you think?

themes/custom/az_barrio/css/az-barrio-off-canvas-nav.css
bberndt-uaz, line 20:
```suggestion
}

#block-az-barrio-offcanvas-searchform .btn:focus {
  background-color: var(--bs-red);
  color: var(--bs-white);
}
```

themes/custom/az_barrio/css/az-barrio-off-canvas-nav.css
bberndt-uaz: @djcelaya Sounds good! I have added the focus styling, but I was wondering about the magnifying glass icon color (without focus). It is dark gray (from `var(--bs-btn-color)`), but for the blue header I changed its color to `var(--bs-blue)`. Should I remove that change so that it remains dark gray? When I tried the alternative (making the icon red to match our current Arizona Header), I wasn't sure I liked it.

bberndt-uaz: OK, the search icon should be dark gray with both red and blue Arizona Headers.

camikazegreen (approved):

djcelaya (approved):

PR #5575, opened 2026-05-06
Sanitize Application_Form_URL__c which can be an HTML anchor or plain URL

kevdevlu:
Subbranch of #5182 . 

Sanitize Application_Form_URL__c — extract href when API returns an HTML anchor

## Problem

The Trellis `/programinfo` API returns `Application_Form_URL__c` inconsistently: most records send a plain URL, but some send a full HTML anchor tag:

```html
<a href="https://forms-a.trellis.arizona.edu/148?acctid=001V400000m5V4L&amp;progcycid=a55V4000015uA1i" target="_blank">Application Link</a>
```

The migration previously mapped this field directly to field_az_application_link/uri, so anchor-tag values were stored as-is and rendered as `Apply now` on the frontend.

## Solution
Adds a custom migrate process plugin az_extract_url (AZExtractUrl.php) and wires it into the migration pipeline for field_az_application_link/uri.

### The plugin:

- If the value is a plain URL → passes it through unchanged
- If the value contains an <a> tag → uses Html::load() + DOMXPath to extract the href, then Html::decodeEntities() to decode `&amp;` → & so the stored URI is valid
- If an anchor is found but has no usable `href`, it returns '' so the link field is left empty rather than storing garbage

## Testing
Re-run the migration with drush migrate:import az_trellis_opportunities --update to reprocess existing nodes. Verify that records which previously had anchor-tag values now display a working Apply link on the frontend.


tadean:
I think this should be fixed on the Trellis side by validation, ideally.

PR #5609, opened 2026-05-14
Close #5602 Enabling Enterprise Attributes with Migration Magician already installed causes an error

### LINKED ISSUE #5602: Enabling Enterprise Attributes with Migration Magician already installed causes an error
## Problem/Motivation
Error when installing Quickstart Opportunity and Quickstart Opportunity - Trellis Opportunity Importer.

"Error message: Fatal error: The modifiers of the trait method getLegacyDrupalVersion() are changed, but this method does not exist. Error in /code/web/modules/contrib/migmag/src/Traits/MigMagMigrationConfigurationTrait.php on line 12"

The error shows the module incorrectly installed if you back out, triggering once. Uninstalling and re-installing the module does not trigger the error. Reported that there was a WSOD error when trying to access content after the error.

### To Reproduce
Seems to be related to MigMag module related to Enterprise Attributes, specifically the az_migration module. MigMag must be installed first for the error to appear. The error happens once in installation.

1. First, Install Migrate Magician module and all its included submodules.
2. Install Enterprise Attributes, or any module that requires Enterprise Attributes (like Opportunities - Trellis Importer)

## Proposed resolution
MigMag is a contrib module required by Enterprise Attributes, but doesn't seem to actually be installed when enabled. For example, Opportunities requires Enterprise Attributes, and when we install it it does not also install Migration Magician.

##  Screenshots

zsaenz:
## Description

Due to the issue appearing in Enterprise Attributes and not just Opportunities, we discovered it might be a dependency or load order issue. I found that installing az_migration first (which was not required) seemed to fix the issue - this module was not required by any of our importing tools.

-Adding it to Opportunities Trellis fixes the issue, but it would still occur for Enterprise Attributes module. (Event Trellis importer does not require attributes like opportunity does)

## Related issues

Closes #5602 
## How to test

1. First, install Migrate Magician and its 7 submodules
2. Install either Quickstart Opportunity - Trellis Opportunity Importer, or Enterprise Attributes.
3. The modules should now install without issue.
4. Testing needed for reported sites

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_core/az_enterprise_attributes_import/az_enterprise_attributes_import.info.yml
+++ b/modules/custom/az_core/az_enterprise_attributes_import/az_enterprise_attributes_import.info.yml
@@ -8,6 +8,7 @@ lifecycle_link: 'https://github.com/az-digital/az_quickstart/blob/main/RELEASES.
 dependencies:
   - az_core:az_core
   - drupal:migrate
+  - az_migration:az_migration
   - migrate_plus:migrate_plus
   - migrate_tools:migrate_tools
   - migrate_queue_importer:migrate_queue_importer
--- a/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.info.yml
+++ b/modules/custom/az_opportunity/az_opportunity_trellis/az_opportunity_trellis.info.yml
@@ -8,8 +8,8 @@ lifecycle_link: 'https://github.com/az-digital/az_quickstart/blob/main/RELEASES.
 
 dependencies:
   - az_core:az_core
+  - az_enterprise_attributes_import:az_enterprise_attributes_import
   - az_opportunity:az_opportunity
-  - az_migration:az_migration
   - az_migration:az_migration_remote
   - drupal:views
   - drupal:migrate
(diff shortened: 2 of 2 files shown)


kevdevlu:
i think the problem is in `migmag` itself, not in our Enterprise Attributes module. it seems that any module that runs a migration in hook_install without `migrate_drupal` as a dependency will trigger the same error. 

So for example, if you install `migmag` and all its other `migmag_*` submodules, then try to install `az_demo`, it will also return this fatal error

joeparsons:
> i think the problem is in `migmag` itself, not in our Enterprise Attributes module. it seems that any module that runs a migration in hook_install without `migrate_drupal` as a dependency will trigger the same error.
> 
> So for example, if you install `migmag` and all its other `migmag_*` submodules, then try to install `az_demo`, it will also return this fatal error

I agree.  I think we may want to create a `migmag` module project issue on d.o., especially if we can identify steps to reproduce the error on a vanilla Drupal install.

zsaenz:
https://www.drupal.org/project/migmag/issues/3590259

https://github.com/az-digital/az_quickstart/pull/5611

PR #5636, opened 2026-05-22
Closes #5635: Fix Recurring Import Issue, Enterprise Attributes Issue

### LINKED ISSUE #5635: Trellis Event Issues Discovered from Implementing Opportunity Content Type
### Issue 1: Recurring Import Event IDs get dropped when events are already imported**
How to Reproduce
1. Enable `az_event_trellis`. 
2. Go to the importer: `/admin/trellis-event-importer`. 
3. Import like 3 things. For example, you can filter **Topic** for `Humanities`. From the search results import 3 events. Make a note of which 3 events you have imported
4. Create a recurring import with a different set of events. For example, filter **Topic** for  `Arts & Culture` and then **Category** for `Lectures, Workshops, & Panels`. Click search, and create a recurring import from that search. Make a note of what events are going to be imported for this recurring import. 
5. To see what happens when the recurring import runs, set **Quickstart Trellis Events** in Cron migration `/admin/config/migrate_queue_importer/cron_migration` to have 0 interval. Then go to `/admin/config/system/cron` to run cron by clicking on the link in the page. 
6. If you have imported 3 events from step 3, and there are 4 events from the recurring import in step 4, the first 3 events from the recurring import got dropped, and only the last event from the recurring import would actually get imported.

### Issue 2: Recurring Import Event ID's get dropped when there are multiple recurring imports 
To reproduce, do the same steps as above, except this time, in Step 3, instead of importing, create a recurring import from the search results. After doing step 3 and 4 above, you should have 2 recurring imports. When you run cron, it will import the events in the first recurring import, but not the second recurring import within the number of events of the first recurring import, so: 

- First Recurring Import: [A, B, C]
- Second Recurring Import: [D, E, F, G, H]

What's actually being imported: [A, B, C, G, H]. 
D, E, F are dropped

### Issue 3: the Trellis Events API searches using the value of Enterprise Attributes, but we're submitting the key so if the value and key differ like in the case of Science & Research, we get empty results

1. Go to trellis event importer `/admin/trellis-event-importer`
2. Search for Topic: `Science & Research`
3. No results returned

## Proposed resolution
Description of the proposed solution, the rationale behind it, and workarounds for people who cannot use the patch.

##  Screenshots

kevdevlu:
Closes #5635 

During development of Opportunity Content Type, we found that we should use `array_merge`: [Commit 069b53d](https://github.com/az-digital/az_quickstart/pull/5182/changes/069b53da627994ff920071a17da7b971dd76e452), [Commit 892fe54](https://github.com/az-digital/az_quickstart/pull/5182/changes/892fe54e8dd1220bf7e20c49c223f6fcd822275f#diff-f6f7e348fe3d0e71ecf0f8770b6f7530a1e937cf71ddeb8f874a0c54aa511d32R69-R72) . We also found that we should use enterprise attributes values instead of the keys to search for events. [Commit c03200b](https://github.com/az-digital/az_quickstart/pull/5182/changes/c03200b9ab8bd24fb7a6e3e12e388ffdb16c648c). 

This PR adds these fixes to Trellis Events.

### Verify Issue 1 Fixed: Recurring Import Event IDs get dropped when events are already imported**

1. Enable `az_event_trellis`. 
2. Go to the importer: `/admin/trellis-event-importer`. 
3. Import 3 events. For example, you can filter **Topic** for `Humanities`. From the search results import 3 events. Make a note of which 3 events you have imported
4. Create a recurring import with a different set of events. For example, filter **Topic** for  `Arts & Culture` and then **Category** for `Lectures, Workshops, & Panels`. Click search, and create a recurring import from that search. Make a note of what events are going to be imported for this recurring import. 
5. Set **Quickstart Trellis Events** in Cron migration `/admin/config/migrate_queue_importer/cron_migration` to have 0 interval. Then go to `/admin/config/system/cron` to run cron by clicking on the link in the page. 
6. If you have imported 3 events from step 3, and there are 4 events from the recurring import in step 4, verify that all 7 events are imported. Example:

- What's already imported: [A, B, C]
- Recurring Import: [D, E, F, G]

What should be imported: [A, B, C, D, E, F, G]

### Verify Issue 2 Fixed: Recurring Import Event ID's get dropped when there are multiple recurring imports 
Do the same steps as above, except this time, in Step 3, instead of importing, create a recurring import from the search results. After doing step 3 and 4 above, you should have 2 recurring imports. When you run cron, it should import the events from both the first and second recurring import minus any duplicates: 

- First Recurring Import: [A, B, C, D]
- Second Recurring Import: [D, E, F, G, H]

What should be imported: [A, B, C, D, E, F, G, H]

### Verify Issue 3 Fixed: the Trellis Events API searches using the value of Enterprise Attributes, but we're submitting the key so if the value and key differ like in the case of Science & Research, we get empty results

1. Go to trellis event importer `/admin/trellis-event-importer`
2. Search for Topic: `Science & Research`
3. Make sure that results are being returned

- **Patch release changes**
   - [x] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_event/az_event_trellis/src/Form/AZRecurringImportRuleForm.php
+++ b/modules/custom/az_event/az_event_trellis/src/Form/AZRecurringImportRuleForm.php
@@ -134,7 +134,8 @@ final class AZRecurringImportRuleForm extends EntityForm {
       $terms = $query->execute();
       $terms = $term_storage->loadMultiple($terms);
       foreach ($terms as $term) {
-        $options[$term->field_az_attribute_key->value] = $this->entityRepository->getTranslationFromContext($term)->label();
+        $label = $this->entityRepository->getTranslationFromContext($term)->label();
+        $options[$label] = $label;
       }
 
       // Build the select element for the attribute.
--- a/modules/custom/az_event/az_event_trellis/src/Plugin/migrate/source/AZTrellisEventSource.php
+++ b/modules/custom/az_event/az_event_trellis/src/Plugin/migrate/source/AZTrellisEventSource.php
@@ -66,8 +66,10 @@ class AZTrellisEventSource extends SourcePluginBase {
     $this->trellisIds = $configuration['trellis_ids'] ?? [];
     // If no arguments are supplied, fetch the list currently on the site.
     if (empty($this->trellisIds)) {
-      $ids = $this->trellisHelper->getImportedEventIds();
-      $ids += $this->trellisHelper->getRecurringEventIds();
+      $ids = array_merge(
+        $this->trellisHelper->getImportedEventIds(),
+        $this->trellisHelper->getRecurringEventIds(),
+      );
       $this->trellisIds = array_unique($ids);
     }
   }
--- a/modules/custom/az_event/az_event_trellis/src/Plugin/views/filter/AZEventTrellisViewsAttributeFilter.php
+++ b/modules/custom/az_event/az_event_trellis/src/Plugin/views/filter/AZEventTrellisViewsAttributeFilter.php
@@ -84,7 +84,8 @@ class AZEventTrellisViewsAttributeFilter extends FilterPluginBase {
     // Build option list.
     foreach ($terms as $term) {
       if ($term->hasField('field_az_attribute_key') && !empty($term->field_az_attribute_key->value)) {
-        $options[$term->field_az_attribute_key->value] = $this->entityRepository->getTranslationFromContext($term)->label();
+        $label = $this->entityRepository->getTranslationFromContext($term)->label();
+        $options[$label] = $label;
       }
     }
     $form['value'] = [
--- a/modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
+++ b/modules/custom/az_event/az_event_trellis/src/TrellisHelper.php
@@ -216,7 +216,7 @@ final class TrellisHelper {
     $event_api_ids = [];
     foreach ($imports as $import) {
       /** @var \Drupal\az_event_trellis\Entity\AZRecurringImportRule $import */
-      $event_api_ids += $import->getEventIds();
+      $event_api_ids = array_merge($event_api_ids, $import->getEventIds());
     }
     // Remove duplicates in case searches overlapped.
     $event_api_ids = array_unique($event_api_ids);


danahertzberg (approved):

modules/custom/az_event/az_event_trellis/src/Form/AZRecurringImportRuleForm.php
```diff
@@ -134,7 +134,8 @@ public function form(array $form, FormStateInterface $form_state): array {
       $terms = $query->execute();
       $terms = $term_storage->loadMultiple($terms);
       foreach ($terms as $term) {
-        $options[$term->field_az_attribute_key->value] = $this->entityRepository->getTranslationFromContext($term)->label();
+        $label = $this->entityRepository->getTranslationFromContext($term)->label();
+        $options[$label] = $label;
```
tadean, line 138: It's worth noting that this seems like an API bug in the Trellis API. They should ideally be using the unique identifier of the attribute for searching rather than the label. Maybe we should let them know about this issue.

This change also means that any existing recurring imports on sites with one of the affected terms where the key and value are different (Mostly science & research) still won't function, since they were saved with the wrong value unless we update them. But, this is an experimental module.

tadean:
Tested that multiple recurring imports both import successfully, and that affected attributes like Science & Research find results.

tadean (approved):

modules/custom/az_event/az_event_trellis/src/Form/AZRecurringImportRuleForm.php
kevdevlu: Good point! Elsie, the API dev we've been working with, is on parental leave. I think given the other priorities we have currently, maybe when she comes back is a good time to ask them to make this change and also revert this change on our end? 

I'm doubting that any recurring imports would exist with Science & Research? Since the only way people create them (from what i understand) is searching first and then creating recurring import from that search. Since the API was returning empty results for Science & Research, i doubt people would make a recurring import of the empty searches?

tadean: It's also possible to create recurring imports directly through the configuration interface rather than the modal form on the search form. But I agree that it's probably unlikely, most users probably use the search interface to create them.

I think it's enough of an edge case for an experimental module that maybe nothing needs to be done about it.

Maybe we could add some release notes for what terms were affected?

joeparsons (approved):

PR #5639, opened 2026-05-28
High-confidence a11y: fix footer contentinfo duplication and main navigation/menu semantics

accesswatch:
Narrowly-scoped, high-confidence accessibility fixes in the `az_barrio` theme.

## Changes
1. **Footer landmark.** Remove the nested `role="contentinfo"` on the inner
   footer `<div>` in `layout/page.html.twig`. `<footer class="site-footer">` is
   already an implicit `contentinfo` landmark, so the explicit role created a
   **duplicate** landmark. (This fix was applied earlier in the PR and then
   accidentally re-added in a later commit; it is now re-applied.)
2. **Primary navigation name.** Add `aria-label="Primary navigation"` to the
   top `<nav>` so the navigation landmark is named.
3. **Main menu semantics.** Convert the dropdown submenu from a bare
   `<div class="dropdown-menu">` to a proper `<ul>`/`<li>` list, and collapse
   the now-redundant `menu_level` `if/else` around the closing `</li>` and
   `</ul>` (both branches were identical after the change).

## Review
Addresses @djcelaya and @joshuasosa: the title/description now match the diff,
the accidentally-reverted footer fix is back, and the redundant menu
conditionals are removed.


.github/workflows/ci.yml
joeparsons, line 1: We're addressing the composer audit issue in a separate PR: #5657

.github/workflows/ci.yml
```diff
@@ -176,7 +176,18 @@ jobs:
...
+          composer_audit_status=$?
+          cat composer-audit.log
+          if [ "$composer_audit_status" -eq 0 ]; then
+            exit 0
+          fi
+          if [ "$composer_audit_status" -eq 2 ] && grep -q "No security vulnerability advisories found." composer-audit.log; then
+            echo "Composer audit reported only abandoned packages; continuing."
+            exit 0
+          fi
+          exit "$composer_audit_status"
```
joeparsons:
```suggestion
          composer audit --no-dev --ignore-severity=low --ignore-severity=moderate --abandoned=report
```

joeparsons (dismissed):
Overall these changes look good to me.  We'll just want to revert the changes to the `composer audit` job in the CI workflow before merging.

.github/workflows/ci.yml
joeparsons: This was fixed while fixing the merge conflicts with main.

themes/custom/az_barrio/templates/navigation/menu--main.html.twig
```diff
@@ -65,12 +67,14 @@
           {% endif %}
         {% if menu_level == 0 %}
         </li>
+        {% else %}
+        </li>
```
joshuasosa, line 68: If both if/else cases result in `<li>`, if-else isn't needed here.

themes/custom/az_barrio/templates/layout/page.html.twig
joshuasosa: I'm a little confused; the PR description says:
> Removed nested `role="contentinfo"` inside footer wrapper.

But it still appears here with no change.

Looking at `main`'s version of [page.html.twig](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/layout/page.html.twig) doesn't seem to show a duplicate `role="contentinfo"`. Does that mean this PR's description and title is outdated?

From what I can tell, the changes here are unrelated to the footer landmark. If that's the case, I'd suggest updating the title and description to more accurately reflect the changes within (i.e., remove references to footer landmark changes since there are none here).

djcelaya:
```suggestion
        <div class="text-bg-warm-gray py-5">
```
I see the confusion, @joshuasosa. It looks like this `role="contentinfo"` was removed in https://github.com/az-digital/az_quickstart/pull/5639/changes/f145d75ca8656d713ae170522a45745be35a770a and then added back (I'm guessing accidentally) in the following commit https://github.com/az-digital/az_quickstart/pull/5639/changes/f95de121f7fad40dcc52ba0b706ebe69c0cdf1fe

djcelaya: The PR description seems OK as long as we update the code fixes to match. Perhaps we could update the PR title to something like "High-confidence a11y: fix duplicate contentinfo landmark and main menu semantics" or "Fix footer contentinfo duplication and main navigation/menu semantics".

themes/custom/az_barrio/templates/navigation/menu--main.html.twig
```diff
@@ -65,12 +67,14 @@
           {% endif %}
         {% if menu_level == 0 %}
```
djcelaya:
```suggestion
```

themes/custom/az_barrio/templates/navigation/menu--main.html.twig
djcelaya:
```suggestion
```

themes/custom/az_barrio/templates/navigation/menu--main.html.twig
```diff
@@ -65,12 +67,14 @@
...
         </li>
+        {% else %}
+        </li>
         {% endif %}
```
djcelaya:
```suggestion
```

themes/custom/az_barrio/templates/navigation/menu--main.html.twig
djcelaya: Agreed. I'm wondering if the conditional `</ul>` following this is also now not needed.

djcelaya (changes requested):

accesswatch:
Addressed the feedback in 441d0aa:

- Re-applied the footer `role="contentinfo"` removal on `page.html.twig` (it had been applied then accidentally re-added in a later commit) — thanks @djcelaya for spotting the revert.
- Collapsed the redundant `menu_level` if/else around the closing `</li>` and `</ul>` in `menu--main.html.twig` (both branches were identical after the `<div>` → `<ul>` change) — per @joshuasosa.
- Updated the title and description to match the actual changes.

Re-requesting review. 🙏

accesswatch:
@djcelaya this is ready for another look. All of your suggestions are in (441d0aa): the footer `role="contentinfo"` is removed on `page.html.twig`, and the redundant `menu_level` if/else around the closing `</li>` and `</ul>` in `menu--main.html.twig` are collapsed. I also updated the title and description to match the actual changes. If it looks good, could you please **approve as edited** so we can get it merged? Thanks for the careful review! 🙏

djcelaya (approved):

joshuasosa (approved):

bberndt-uaz (approved):

PR #5642, opened 2026-05-28
A11y scan hardening: block non-AJAX mobile nav callback access

accesswatch:
## Summary
- Prevents non-AJAX crawling of the mobile navigation callback endpoint.
- This endpoint is intended to return Drupal AJAX commands, not standalone HTML documents.

## Issue types targeted from CSV
- `html-has-lang`
- `document-title`
- `label`

These were reported on `/az_core/mobile_nav_callback/...` URLs where scanners treated AJAX payloads as full pages.

## Change
1. `modules/custom/az_core/src/Controller/MobileNavController.php`
   - Injects `request_stack` and checks `isXmlHttpRequest()`.
   - Returns `404` for non-AJAX requests.
   - Preserves current AJAX behavior for legitimate mobile menu requests.

## Impact
- **Impacted files:** 1
- **Scope:** high-confidence endpoint hardening for crawler-facing false-positive reduction.


bberndt-uaz:
The proposed solution does not appear to work. Try clicking on this and observe that you don't get a 404: https://pr5642-n4xghjxwthnpt0xaj2ad8ovkksksbpgg.tugboatqa.com/az_core/mobile_nav_callback/main/Main%20Menu.

The [isXmlHttpRequest() method](https://github.com/symfony/symfony/blob/8.2/src/Symfony/Component/HttpFoundation/Request.php#L1830-L1841) merely checks for a `X-Requested-With` header which may or may not be present with AJAX requests. Also, I don't see this method being used this way in Drupal core. Moreover, as far as I can tell, there is no reliable way to determine if a request was made with AJAX. We could explore some kind of alternate solution in the future, but I think this pull request can be closed.

bberndt-uaz:
Next steps per AZ Digital discussion:

1. Create a Quickstart issue for this (or AJAX requests in Quickstart more generally)
   - Look into https://www.drupal.org/docs/8/api/routing-system/access-checking-on-routes/csrf-access-checking for this issue.
2. Close this pull request.

bberndt-uaz:
This issue can be used for further exploration of solutions:
 - #5916

PR #5770, opened 2026-07-15
refactor: clean code improvements

raymond-julie:
Hi! 👋

I was looking through the codebase and found a few minor clean code improvements that I thought I'd share.

- **Rule applied:** Yoda Conditions
- **Rule applied:** Magic Numbers

Hope this helps! Thanks for your time and for maintaining this project.


joeparsons (changes requested):
Hi @raymond-julie.  Thanks for the PR.  The constants you're proposing would need to be defined in the class in order for this to work.

PR #5785, opened 2026-07-17
Closes #5744: Adding default metatag tokens to az_course (clean version)

### LINKED ISSUE #5744: Add basic Metatag support to az_courses
## Motivation
The existing Courses content type does not add appropriate metatags.

Better metatag support would allow search engines and answer engines to better represent this content.

The Page Title metatag should get filled in with the value from the Course Title field.
The Description metatag should get filled in with the value from the Course Description field.

I've also considered adding the schema.org metatags for the "Course" Type, but that will be a followup.

Content Editors should be able to edit the metatags on individual courses.
Content Administrators should be able to edit the metatag defaults.

camikazegreen:
## Description

Closes #5744 by adding default SEO metatag support for Course content.

This change:
- Adds the optional `field_az_metatag` attachment to the Course content type.
- Adds default metatag title and description templates for Course content using the Course title and description fields.
- Adds the Meta tags field to the Course edit form when `az_seo` is installed.

## Related issues

Closes #5744

## How to test

1. Install a site with both `az_course` and `az_seo` enabled.
2. Create or edit a Course node.
3. Verify the **Meta tags** field is available on the Course edit form.
4. View the Course page source.
5. Verify:
   - The page `<title>` uses the Course title.
   - The `<meta name="description">` uses the Course description.
   - The Open Graph title and description are populated from the default metatag configuration.

- **Patch release changes**
   - [x] Bug fix

## Checklist


camikazegreen:
I think what's happening here is that az_course now sets az_seo as a dependency, which is causing az_seo to be installed before az_event.

This configuration file is included in two locations:
/modules/custom/az_event/config/install/core.entity_form_display.node.az_event.default.yml
/modules/custom/az_seo/config/quickstart/core.entity_form_display.node.az_event.default.yml

I suspect that this is just the first one to fail, but this would also be an issue for person, news and flexible page. 
Would it require refactoring how all of these content type modules interact with az_seo in order for this to install correctly?

tadean:
> I think what's happening here is that az_course now sets az_seo as a dependency, which is causing az_seo to be installed before az_event.
> 
> This configuration file is included in two locations: /modules/custom/az_event/config/install/core.entity_form_display.node.az_event.default.yml /modules/custom/az_seo/config/quickstart/core.entity_form_display.node.az_event.default.yml
> 
> I suspect that this is just the first one to fail, but this would also be an issue for person, news and flexible page. Would it require refactoring how all of these content type modules interact with az_seo in order for this to install correctly?

Course shouldn't be installed at all in the base install, since it's an optional module. Taking a look at this... :confused:

tadean:
I think you're right though... It's possible that just the presence of the dependency is making drupal change the order it installs the modules in, even if it's not actually installing az_course.

tadean:
I wonder if there's a core issue that Drupal orders the dependency tree on install regardless of whether the modules are to be installed or not.

tadean:
@camikazegreen it does seem like that happens - as soon as the `az_seo` dependency is declared (on the optional module that's not being installed), it starts trying to install `az_seo` before `az_event`

If the dependency is removed, it starts installing them in the original order.

Strange how a module that's not being enabled affects the install order...

camikazegreen:
In order to successfully install this change to course, I ended up having to refactor the rest of the content types to eliminate duplicate config items. 

I've successfully tested this in Tugboat. It installs clean, and I've verified that all existing content types work as expected. Installing az_course and importing courses works and display the appropriate metatags.

I could use particular review on all of the database updates, and I'll attempt to test them on an existing site next week.

tadean:
I think we should potentially consider adding a feature to our quickstart override config provider that only attempts to override configuration if the dependencies are met and/or the configuration it's overriding exists. (So that the course seo configuration can exist in the az_seo module)

camikazegreen:
We'll look at making these improvements as part of a larger review of how we handle az_seo and our overall metatag strategy.

PR #5834, opened 2026-07-24
Closes #5813: Add Ranking-related SDC's and Refactor az_ranking to use them

### LINKED ISSUE #5813: Use an SDC to output the rankings paragraph.
## Motivation
It would be beneficial to be able to use our designed rankings in multiple contexts.

For example, right now, if you want to add a ranking to a page you have to use a paragraph, but this is not the only way someone can add content to a page. Drupal Canvas also allows creating content in a different way, but both ways support Single Directory Components (SDC).

Currently, the ranking component markup and styles are tightly coupled to the Paragraphs module template. Content editors using other site-building methods (such as Drupal Canvas, layout builders, or custom blocks) cannot reuse the ranking component without duplicating code or forcing a paragraph wrapper.

Refactor the ranking presentation logic into a standalone Single Directory Component (SDC) housed directly within the `az_quickstart` install profile:
- Create an SDC (e.g., `component.twig`, `component.json`, and associated CSS/JS) defining clear props and slots for rankings inside the `az_quickstart` profile component library.
- **Image Handling:** Ensure image rendering relies on standard Drupal image fields/render arrays or native Twig image handling rather than `canvas:image`, keeping `az_quickstart` decoupled from Canvas-specific components.
- Update the existing Rankings Paragraph template to act as a lightweight wrapper that embeds/includes the SDC and passes paragraph field values into the SDC props.
- Enable direct usage of the Rankings SDC across Drupal Canvas, Paragraphs, or any other SDC-compatible site-building tool without introducing module-to-theme dependencies.

- **Keeping rankings exclusive to Paragraphs:** Restricts layout flexibility and forces content creators into a single workflow.
- **Putting the SDC in the theme:** Creates an anti-pattern where modules/profile components depend on theme availability.
- **Duplicating Twig templates/styles for Canvas:** Creates maintenance overhead and risks UI inconsistency across the site over time.

- **Anonymous user:** No impact; views identical frontend ranking outputs.
- **Authenticated user:** No impact.
- **Content editor:** Gains flexibility to insert and configure rankings seamlessly using either Paragraphs or Drupal Canvas.
- **Content administrator:** Gains improved component reuse and consistent component behavior across the distribution.
- **Administrator:** No permission changes required; simplifies architecture.

## Additional context
- Aligns with Drupal Core's Single Directory Components (SDC) paradigm for component-driven development.

kevdevlu:
Closes #5813, [MAR-259](https://uarizona.atlassian.net/browse/MAR-259) , and fixes a (not super related) ranking issue (#5156) as well

### Goal: 
- Make single-source-of-truth SDC's that would be used whether a user is:
    - creating rankings with a traditional `az_flexible_page` using the az_ranking paragraph type 
    or
    - creating rankings with Drupal Canvas
- Refactor az_ranking's `AZRankingDefaultFormatter.php` to use the above SDC's.

### Key Decisions from Discussion with Chris: 
1. Use a new `components` folder in our `az_quickstart` profile as a place to put SDC's. 
    Reasoning: we want to avoid adding more components to `az_barrio`. For this specific component, that would mean making az_ranking (a module) depend on a theme, which we want to avoid. 

2. Instead of using `canvas:image` (`$ref: json-schema-definitions://canvas.module/image` in our schema), we want to use something that doesn't depend on canvas and wouldn't break if canvas was not enabled. 

### High Level Overview of Changes
1. `components` directory has the new SDC's: `ranking`, `ranking-deck`, and `ranking-image`

2. From `az_ranking`, we moved rankings related `twig`, `js`, and `css` to the SDC's

3. Refactored `az_ranking` to use a `AZRankingComponentBuilder` for both form widget previews (live form state) and the formatter (displaying stored items)

   ```
   stored field items ─┐
                       ├─→ AZRankingComponentBuilder ─┐
   live form state ────┘                              │
                                                      ├─→ az_quickstart:ranking
   Canvas prop sources ───────────────────────────────┘      (.twig / .css / .js)
   ```

4. `ranking-image`'s image prop is an object (`{src, alt, width, height}`) matching Canvas's own image shape. This was recommended by [penyaskito, a creator of Canvas, in Drupal Slack drupal-canvas channel](https://drupal.slack.com/archives/C072JMEPUS1/p1786132948687519). See [his page here](https://penyaskito.com/articles/2026/07/18/canvas-tipstricks-declaring-images-sdcs). This provides a media library picker in Canvas for editors just like before. It's matched structurally, with no `$ref`, which makes it Canvas independent.

5. Added to `az_media`: an `az_media_image_style` twig filter that applies an image style in twig

   ```
   <img src="{{ image.src|az_media_image_style('max_1300x1300') }}">
   ```

   `az_media_image_style()` works whether the prop was built from paragraphs page builder or from Canvas. 

6. Deleted the `az_ranking_responsive` image style. After testing, I realized upscaling (which is what `az_ranking_responsive` does) is not needed for focal point to work. This PR is now using `max_1300x1300` image style for ranking image, which comes with Quickstart out of the box and already scales down and converts to WebP.  
    - `az_ranking_update_1130601()` removes the old style on existing sites, and leaves it alone if some other config still depends on it.

7. Fixed a bunch of old issues with our ranking's when editing using paragraphs page builder

   a. Edit existing rankings page. Reorder a ranking, Click `Update Preview`, `Remove`, or `Add Another Item`. Rankings go all over the place, and widget previews don't match fields anymore.
   b. add a ranking. don't enter any text. save. The page displays it as a blank ranking, when it should just ignore it.
   c. Add some rankings. Then try to remove them all. You can't. Only let's you remove until you reach 2.
   d. In an existing paragraphs page builder with rankings, add a new ranking, then click "update preview" on another existing ranking: the newly added ranking in the form is treated as non-empty so it collapses instead of staying open

   ** Use a local build of main or any other PR's tugboat to see these behaviors. Some of it apply to other paragraphs like cards too. This PR's tugboat should have all of these resolved

### Flagship Site Testing
I applied this PR to flagship site (the New QS Ranking Components is this PR)
https://ranking2-azs-arizona.pantheonsite.io/about/rankings
1. Test Everything looks right

### Tugboat Testing: 
1. Enable Canvas
3. Test we can achieve this page's display of rankings using Canvas: https://test-azs-arizona.pantheonsite.io/about/rankings . 
4. Test using the traditional paragraphs page builder (`az_flexible_page`) and using az_ranking paragraph type works as well. 

### Testing Existing Sites
I applied this PR to nursing: 
https://rankings-azs-nursing.pantheonsite.io/ . compare to: https://live-azs-nursing.pantheonsite.io/
1. Test Everything looks right. Visual regressions like padding, font size, font weight, etc. are intentional if those changes match flagship's ranking page: https://www.arizona.edu/about/rankings. If a visual regression doesn't match flagship, then that's an error.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

[MAR-259]: https://uarizona.atlassian.net/browse/MAR-259?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/ranking-image/ranking-image.component.yml
trackleft, line 1: Maybe we should call this ranking-image or grid-image.  WDY Think?

components/image/image.component.yml
```diff
@@ -0,0 +1,85 @@
...
+    src:
+      type: string
+      title: Image
+      description: Pick an image from the media library.
+      format: uri
+      contentMediaType: image/*
+      x-allowed-schemes:
+        - public
+      examples:
+        - public://canyon_running.jpg
```
trackleft:
```suggestion
        - https://placehold.co/600x400?font=Proxima+Nova
```

trackleft: I think we should use placeholder images for this, preferably with a suggested size printed on them.

https://placehold.co/

kevdevlu: The problem is we're trying to match this image shape, which requires `public://` (and we're using this shape because compared to using $ref, this one doesn't crash if we don't have Canvas enabled): 
```php
(
        isset($storable_prop_shape->shape->schema['contentMediaType'])
        && $storable_prop_shape->shape->schema['contentMediaType'] === 'image/*'
        // Stream wrapper URIs can only be `format: uri|iri`.
        && JsonSchemaStringFormat::from($storable_prop_shape->shape->schema['format'])->allowsOnlyAbsoluteUri()
        // @see json-schema-definitions://canvas.module/stream-wrapper-image-uri
        && ($storable_prop_shape->shape->schema['x-allowed-schemes'] ?? []) === ['public']
      )
```
https://git.drupalcode.org/project/canvas/-/blob/1.x/src/Hook/ShapeMatchingHooks.php#L282

Which, `x-allowed-schemes` is only public, so we can have `public://___________` but not `https://_____________`

But i totally get having a placeholder image (from an editor's perspective), so this commit is what i am suggesting: https://github.com/az-digital/az_quickstart/pull/5834/changes/00b6552e46bd1a81f027167b687907dc9107be1d

Open to any other ideas

components/ranking-image/ranking-image.component.yml
kevdevlu: i changed it to `ranking-image`

composer.json
```diff
@@ -44,6 +44,7 @@
         "drupal/bootstrap_barrio": "5.5.20",
         "drupal/bootstrap_utilities": "3.0.1",
         "drupal/calendar_link": "3.0.4",
+        "drupal/canvas": "^1.8",
```
joeparsons: I'm thinking we may not want to add canvas as a composer dependency yet.  Maybe we could consider adding it as a development dependency in dev dependencies package instead?
https://github.com/az-digital/az-quickstart-dev

kevdevlu: @joeparsons i added this in the PR description: 
> Undo (if needed) before Merging:
Currently we added drupal/canvas to composer.json so Tugboat build installs Canvas for review; to be removed in a follow-up commit once review is done.

This was just the only way i knew how to add it to tugboat build for testing, but totally can remove it after review. Is there a better way to do this?

joeparsons: Assuming we want to have `drupal/canvas` available as a development dependency we could also create a PR branch in the `az-digital/az-quickstart-dev` repo with the same name as this PR branch and add `drupal/canvas` to its composer.json file.  That should make `drupal/canvas` available for Tugboat, Lando, DDEV, etc but not for production sites.

components/ranking-image/ranking-image.component.yml
```diff
@@ -0,0 +1,105 @@
...
+status: experimental
+description: A layout- and accessibility-aware image, backed by the media library. Works standalone on a page, or dropped into a Ranking Deck alongside Ranking cards.
+props:
+  type: object
+  properties:
+    src:
+      type: string
+      title: Image
+      description: Pick an image from the media library.
+      format: uri
```
trackleft:
```suggestion
      format: uri-reference
```

composer.json
kevdevlu: thanks for this suggestion! I did exactly as suggested

components/ranking-image/ranking-image.component.yml
kevdevlu: This didn't work, but if we really want it to be able to take a https url image, i made a subbranch #5849 basically reproducing the internal mechanisms of canvas onto our az_media module > i don't recommend it because it adds so much complexity / extra maintenance just to allow https://placeholder.co to work

kevdevlu:
Troy suggested at Jul 29 PR review:  What if we used an internal placeholder generating service , like what the external placehold.co does?

kevdevlu:
Asked on `#drupal-canvas` channel on Drupal Slack Workspace: https://drupal.slack.com/archives/C072JMEPUS1/p1786132948687519

> Hi all , i wanted to ask for advice on our situation :thinking_face:
> Context: We maintain a large portfolio of sites that we’d like to make Canvas-ready without forcing them onto Canvas yet.
> 
> - Current setup: We use Paragraphs for page building, with many custom paragraph types bundled in one central distribution. 
> - Goal: We’ve found we can build SDCs that serve both our Paragraphs rendering and Canvas, so we’d like to start deploying SDCs everywhere now, even on sites where Canvas isn’t installed yet. 
> - Challenge: The best-working image component for Canvas seems to be canvas:image, but depending on it means depending on Canvas.
> 
> I did a search here and https://drupal.slack.com/archives/C072JMEPUS1/p1772045518951249?thread_ts=1772029243.094249&cid=C072JMEPUS1 seems relevant, but the suggested shape seemed to give no image widget at all :grimacing: Adding
> ```
> x-allowed-schemes:
>   - http
>   - https
> ``` 
> gives me the basic image widget but not Media Library…
> 
> So my question: Is there an SDC image pattern or approach you guys would recommend that works cleanly in both environments before Canvas is enabled across all our sites?

I got a response from penyaskito: https://penyaskito.com/articles/2026/07/18/canvas-tipstricks-declaring-images-sdcs and have implemented this here in this PR!

kevdevlu:
**Wednesday - AZ Digital Meeting**: Chris said the main thing for this is, we need to make sure editors / power users of quickstart around campus, and people who support users want this kind of structure: 

- 3 simple components: `ranking-deck`, that can house either a `ranking-image` or `ranking`

1. Dana mentioned that we're going to have focus groups review Canvas's way of building pages to compare to paragraphs page builder with our wider users before putting all our eggs into the canvas 'basket'. 

2. After we reached an agreement that we want Canvas to be a parallel option (or main option) for page building, we can formalize how we want SDC's to be structured, based on the page building experience that everyone wants. For example, in this case specifically, do we want 1 BIG component that can be configured. or 3 small components like what we have now.

**Decision:** is to leave this until after 1. focus groups review and 2. formalization of a standard of how to structure SDC's for best editor experience in Canvas

PR #5808, opened 2026-07-21
Closes #5771, #5806: Update experimental AZ Navbar Fullscreen for Arizona Bootstrap v5.1.6

bberndt-uaz:
## Description

1. Updates the implementation for applying autocomplete='off' to the search block form input element (only when the blue header or AZ Navbar Fullscreen are enabled).
2. Removes temporary CSS overrides for AZ Navbar Fullscreen issues in Arizona Bootstrap v5.1.5. These issues have been fixed in Arizona Bootstrap v5.1.6.
3. Updates AZ Navbar Fullscreen template to account for changes in Arizona Bootstrap v5.1.6.
4. Makes AZ Navbar Fullscreen headings configurable with changes to the AZ Navbar Fullscreen block configuration form, AZ Navbar Fullscreen block schema and template, and the AZ Navbar Fullscreen component schema and template.
5. Fixes `.active` class not being applied to call-to-action and footer menu links

### Release notes

> ### Changes to experimental AZ Navbar Fullscreen blocks
> AZ Navbar Fullscreen blocks have been updated to give site owners more control over the top and bottom footers. New configuration settings have been added to specify the headings for top and bottom footers on both desktop devices and mobile devices. Also, the `resources_menu` and `helpful_links_menu` settings have been renamed to `footer_top_menu` and `footer_bottom_menu`. It is recommended to delete and replace existing AZ Navbar Fullscreen blocks.

## Related issues

 - #5771
 - #5806 

## How to test

1. Confirm no changes with the existing Arizona Header desktop search and mobile nav search.
2. Go to Arizona Barrio settings (`/admin/appearance/settings/az_barrio`), then go to Components -> Navbar Behaviour and check the "Enable AZ Navbar Fullscreen (experimental)" checkbox.
3. Go to Block Layout (`/admin/structure/block`) and in the Navigation Bar region, disable the "Main navigation" block.
4. Still in Block Layout, add a new "AZ Navbar Fullscreen Menu" block to the Navigation Bar region and configure the desired menus and search form blocks (`az_barrio_search` for the desktop navbar and `az_barrio_offcanvas_searchform` for the offcanvas modal). The footer headings have default values that are automatically set unless changed.
6. Clear cache.
7. Confirm the functionality and appearance of the new navbar on both desktop and mobile.

Note: There is an issue with the mobile navigation where the Resources by Audience menu page is displayed as the initial menu page on the home page. This is due to the link in that menu also pointing to the front page. This behavior will be corrected in Arizona Bootstrap.

[...]

PR #5842, opened 2026-07-27
Prevent duplicate az-search class on search block when blue header AND AZ Navbar Fullscreen are enabled

djcelaya:
See #5808, specifically [this comment](https://github.com/az-digital/az_quickstart/pull/5808#pullrequestreview-4790069232).

Fix duplicated `.az-search` class when both blue header and AZ Navbar Fullscreen are enabled at the same time.

### How to test
- Verify right spacing to nav toggle when only AZ Navbar Fullscreen is enabled, only blue header is enabled, and AZ Navbar Fullscreen AND blue header are enabled together.


bberndt-uaz:
@djcelaya Thanks! I'll merge this fix.

PR #5846, opened 2026-07-27
Closes #5840 Use CAPTCHA patch for webform challenge placement

### LINKED ISSUE #5840: Replace webform CAPTCHA placement code with patch
## Motivation
The az_spam_prevention module has custom code to add CAPTCHA to all webforms. However, there's an existing issue and patch for the CAPTCHA module that achieves the same result. It could be good to replace the custom code with the patch and align efforts with the CAPTCHA module issue.

Remove custom code that adds CAPTCHA to webforms, then add the patch from [captcha/issues/3380095](https://www.drupal.org/project/captcha/issues/3380095).

Leave the az_spam_prevention module as-is. Potentially remove custom code if and when a solution is adopted by CAPTCHA module.

## Additional context
PTYS dept sites that have used CAPTCHA have used the patch successfully.

Webforms with multi-page parts will see ALTCHA on each page part. This is true with the patch and with the az_spam_prevention approach.

Possible solutions:
- Uninstall az_spam_prevention module on those sites with multi-page parts (not a solution if the patch is adopted)
- Change the ALTCHA challenge to auto
- Change the CAPTCHA to math or image_captcha
- Leave as-is, meaning the user will need to complete the challenge on each page part

joshuasosa:
## Description

This PR replaces custom code for placing CAPTCHA challenges on all webforms with a patch from [CAPTCHA issue 3380095](https://www.drupal.org/project/captcha/issues/3380095).

This proposed change would effectively enable CAPTCHA on all webforms regardless of `az_spam_prevention` module's install status. Currently, if the global application of CAPTCHA on webforms is not desired, `az_spam_prevention` would need to not be installed. Changing to the patch means if the functionality is not desired, the CAPTCHA module would need to not be installed. Ideally, if there are concerns with implementation, they should be brought up in the original issue before the patch is merged in the contrib module.

### Webform Multi-page Note
There is a [CAPTCHA issue 3496865](https://www.drupal.org/project/captcha/issues/3496865) where multi-page webforms will see the CAPTCHA challenge on every page part. This issue exists with the current implementation of `az_spam_prevention` module as well where enabling the custom module will result in the challenge being presented on every webform page part.

The issue may be minimal enough to not warrant a code change to `az_spam_prevention` or blocking of this PR. But if it's considered significant enough, a potential workaround could be to set ALTCHA's "Auto verification" to "On page load."

### Release notes
```
_Notes: When the CAPTCHA module is enabled, it is now applied to all webforms automatically with the configured default challenge. Enable az_spam_prevention module to enable and configure the recommended ALTCHA CAPTCHA challenge.
```

## Related issues

Closes #5480 

## How to test

1. Install Webform, Webform UI and CAPTCHA. Optionally install Quickstart Spam Prevention.
2. Create a form or visit the default Contact webform.
3. View the form while logged out to see the default CAPTCHA challenge on the form.

Demo: https://pr5846-wyg2d4drazj01xb5k2ty97hhqrs9g9go.tugboatqa.com/form/contact

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [x] Patch or minor level update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


zsaenz (approved):

tadean:
In review I noticed that this patch has a difference in functionality from `az_spam_prevention`, in cases where the site default captcha is a different type of captcha than a captcha element configured on the form, the configured captcha will be replaced with the default captcha:

**Form configuration:**
Screenshot from 2026-07-29 09-10-00


**Rendered captcha**
Screenshot from 2026-07-29 09-09-49

This issue doesn't occur with the `az_spam_prevention` hook implementation.

Another issue (which affects both `az_spam_prevention` and this patch) is that if the webform contains a container wrapper for the captcha, a duplicate is created, as neither the patch or the `az_spam_prevention` hook implementation sees it, for example here with a captcha nested in a details element:

Screenshot from 2026-07-29 09-03-01

tadean:
I don't think either of the above issues need to block merging.

tadean:
The patching channel doing testing reported that the `az_spam_prevention` hook causes errors on forms that lack a base form id (such as the block placement page)

```
TypeError: preg_match(): Argument #2 ($subject) must be of type string, null given in preg_match() (line 74 of /code/web/profiles/custom/az_quickstart/modules/custom/az_spam_prevention/src/Hook/AZAntiSpamHooks.php)
```

This PR would resolve this I think, since the hook is removed. If we don't merge this, we should probably create an issue for this.

tadean (approved):

trackleft (approved):

joshuasosa:
Thanks for the review!

Noted caveats for reference:

- Existing captcha elements with a different captcha type is overridden to use the default captcha type
- Existing captcha elements in a container will still result in duplicate captchas
- Existing captcha elements with a key other than `captcha` will still result in duplicate captchas

joshuasosa:
Follow up:
- Add captcha cacheable patch from [captcha/issues/3496865](https://www.drupal.org/project/captcha/issues/3496865)
- Set ALTCHA to auto verify on page load
- Update ALTCHA to 2.x beta

PR #5610, opened 2026-05-14
Fixes #5600 az_search_api document extraction

tadean:
## Description
This PR adds usage of document extraction to the search_api index, so that PDF contents will be searchable. This is done using `search_api_attachments` and a searchstax document extraction endpoint.

### Release notes
```
`az_search_api` will now index the contents of document files attached to index entities. In environments that are not on Pantheon that are using this module, the newly defined secrets `SEARCH_ENVIRONMENT_UPDATE_ENDPOINT` and `SEARCH_ENVIRONMENT_UPDATE_TOKEN` must be put in place.
```

## Related issues
#5600

## How to test
**Note:** to test this PR you need the endpoint and token for a searchstax environment. You can look these up for a searchstax environment with the following steps:

- Login at [searchstudio.searchstax.com](https://searchstudio.searchstax.com/)
- Select or create a **Drupal** site search endpoint
- Click **App Settings**
- Click **All APIs**
- Use **Update Endpoint** as the value below of **SEARCH_ENVIRONMENT_UPDATE_ENDPOINT**
- Copy the **Read & Write** token to use as the value below for **SEARCH_ENVIRONMENT_UPDATE_TOKEN** 
- Copy the **url** in the Document Extractor tab for **AZ_SEARCH_API_EXTRACTOR_ENDPOINT**
- Copy the **Discovery API key** under the Discovery tab for **AZ_SEARCH_API_EXTRACTOR_TOKEN** 

### **If testing locally:**
**Note:** You **must** complete this step before starting your lando container!
Create a `.env` containing:
- a secret called  `AZ_SEARCH_API_TOKEN` with a value equal to JSON matching the following format:
```
{"update_endpoint": "SEARCH_ENVIRONMENT_UPDATE_ENDPOINT", "update_token": "SEARCH_ENVIRONMENT_UPDATE_TOKEN"}
```
- a secret called  `AZ_SEARCH_API_EXTRACTOR_ENDPOINT` with the URL of the searchstax extraction endpoint
- a secret called  `AZ_SEARCH_API_EXTRACTOR_TOKEN` with the authentication token of the searchstax extraction endpoint
In Lando, the env file should look like this:
```
AZ_SEARCH_API_TOKEN={"update_endpoint": "SEARCH_ENVIRONMENT_UPDATE_ENDPOINT", "update_token": "SEARCH_ENVIRONMENT_UPDATE_TOKEN"}
AZ_SEARCH_API_EXTRACTOR_TOKEN=yourauthtoken
[...]

PR #5851, opened 2026-07-28
Fixes #5833, Fixes #5835 metatag keywords, source domain.

### LINKED ISSUE #5833: az_search_api keywords functionality
## Motivation
@RyanDool reports it would be useful if the search_api integration had a way to indicate keywords that are associated with a particular page or site, to make it easier to identify topics of particular pages to boost them in results to a particular query.

Marketing would like it to be easier to identify which pages need to be boosted in accordance with particular specific searches.

## Proposed Resolution
@RyanDool suggests it would be helpful if there was a way to specify keywords for a page that indicate related concepts for a page.

One potential way of handling this is via metatags. (Another way would be adding a dedicated field for containing the keywords).

Given that Quickstart already contains the keywords metatag, this is one potential route. However, [keywords is considered obsolete for SEO purposes](https://developers.google.com/search/blog/2009/09/google-does-not-use-keywords-meta-tag).

In this case, whether or not search engines use the keywords would be irrelevant- they would be for purposes of internal search categorization (and thus hopefully reliable - the lack of reliability is why they became unused in the seo space).

**todo**
- [ ] Consider if this is something we can provide
- [ ] Are metatags the right mechanism? Alternatives?
  - [ ] Dedicated field
  - [ ] Taxonomies
  - [ ] Some other method of boosting
  - [ ] Categorization algorithms  

If this is considered an appropriate approach, the contents of the keywords metatag should be published to the search api index.

### LINKED ISSUE #5835: az_search_api formal source domain field
## Motivation
@RyanDool has proposed it would be useful for the index to have a formal field indicating what site the document came from. The `site` field almost provides this, but can vary in unexpected ways depending on pantheon environment and status of attached pantheon domains on the site. (the `site` field contains the domain that was the base url during the cron run where indexing occurred. In some cases, this can be a pantheon platform URL, an attached domain, or other examples.

@RyanDool proposes a new field, `ss_source_domain` that always contains something like the xmlsitemap base of the site, so it's obvious in the index how to create searchstax promotions for a set of documents from a site.

There should ideally be a field that makes it easy to create a promotion for documents from a certain site.

## Proposed Resolution
Add `ss_source_domain` based upon the xmlsitemap base.

tadean:
## Description
This PR adds a dedicated source domain to the index, and includes the metatag keywords, both as the existing values and as versions transformed to hyphenated tags.

### Release notes
```
`az_search_api` will now index the site's XMLsitemap domain, and metatag keywords.
```

## Related issues
#5833 
#5835 

## How to test
- Complete secrets setup or deploy to pantheon as explained in #5610 
- Create some page content.
  - include metatag keywords.
    - Include a single word
    - Include content keywords that are a delimited list
  - Ensure the site has an xmlsitemap base domain set
- enable `az_search_api`
- visit `/admin/config/search/search-api/index/az_search_api_index`
- click `index now`
- Wait for documents to be indexed ( can take several minutes to commit and be visible in the cluster)
- Examine documents in the index
- Verify `ss_source_domain` contains the base xml site domain
- Verify `tm_X3b_en_keywords` contains the keywords split on comma delimiters
- Verify `sm_keyword_tags` contains the lowercase keywords with whitespace turned into hyphens

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist


modules/custom/az_search_api/src/Plugin/search_api/processor/AZSourceDomain.php
```diff
@@ -0,0 +1,89 @@
...
+    $state = $container->get('state');
+    // Get the xmlsitemap base.
+    $baseUrl = $state->get('xmlsitemap_base_url');
+    if (!empty($baseUrl)) {
+      // Parse the URL to get just the domain.
+      $domain = parse_url($baseUrl, PHP_URL_HOST);
+      if (!empty($domain)) {
+        $processor->baseDomain = $domain;
+      }
+    }
```
joeparsons, line 73: I could be wrong but I think assigning a config or state object to a property during construction / initialization is considered an anti-pattern.  I believe this is because retrieving config and state objects involves touching cache and database storage, which ideally shouldn't be needed to initialize a service (or in this case a plugin).  I think it can also hypothetically cause problems when services/plugins/dependencies are serialized.

I think the preferred approach would be to assign the state service to a property and then use the injected service to retrieve the actual state value in an object method.

See:
https://mglaman.dev/blog/dependency-injection-anti-patterns-drupal
https://mglaman.dev/blog/factories-and-dependency-injection

If this suggested change makes sense to you, I think we'd also want to apply the same change to the existing `AZBaseUrl` search_api processor plugin as well (sorry I didn't notice this when that one was added).

tadean: `baseDomain` is a string rather than state object so I don't think serialization is an issue.

But I think you're right it would be a better practice to avoid calling the service during instantiation. I wouldn't have any objection to that change.

trackleft (dismissed):

modules/custom/az_search_api/src/Plugin/search_api/processor/AZSourceDomain.php
```diff
@@ -0,0 +1,110 @@
...
+  stages: [
+    'add_properties' => 0,
+  ],
+  locked: TRUE,
+  hidden: TRUE,
+)]
+class AZSourceDomain extends ProcessorPluginBase {
+
+  /**
+   * The base_url.
```
joeparsons:
```suggestion
   * The baseDomain.
```
Nitpick

modules/custom/az_search_api/src/Plugin/search_api/processor/AZSourceDomain.php
```diff
@@ -0,0 +1,110 @@
...
+   *
+   * @var string
```
joeparsons:
```suggestion
   * @var string|null
```

joeparsons (approved):

bberndt-uaz (approved):

Issue #5336, opened 2026-02-20
Upgrade existing sites to use AZ Navbar and deprecate theme setting


djcelaya:
## Conditions of satisfaction
- [ ] Existing AZQS3 sites should be upgraded to use AZ Navbar via a config distro update.
- [ ] The theme setting to enable/disable AZ Navbar should be deprecated.
- [ ] Profile-level install hooks introduced in #5312 to enable AZ Navbar on new builds only should be removed.

djcelaya:
We will not enforce 3rd level nav with this upgrade.

Issue #5107, opened 2025-12-03
Add support for AZ Navbar


djcelaya:
## Motivation
A new navbar style `navbar-az` will be available in AZBS 5.0.4. See [AZBS#1751](https://github.com/issues/assigned?issue=az-digital%7Carizona-bootstrap%7C1751). We would like to optionally use it on Quickstart sites.

### Is your feature request related to a problem? Please describe.
We have new styling requirements from central marketing which `navbar-az` addresses. It also includes styling support for a third level of navigation.

### Describe the solution you'd like
Add a mechanism to switch from the current navbar to the new `navbar-az` implementation.

## Conditions of satisfaction
- [x] Add a AZ Barrio theme setting that applies `navbar-az` to the existing `navbar`.
- [x] Add a new template file based off of [menu--main.html.twig](https://github.com/az-digital/az_quickstart/blob/main/themes/custom/az_barrio/templates/navigation/menu--main.html.twig) that will be conditionally used when the new theme setting is enabled.

## Questions
1.  If I have expanded a second level menu and the click another second level menu item to expand it, does the first expanded menu get auto collapsed? Like an accordion? Or do both menus remain expanded at that level? Another way to think about it: is [this](https://github.com/user-attachments/assets/cc6144f3-4fe1-4c96-ada2-3193c4d74189) allowed?

PR #5958, opened 2026-09-04
Closes #5848: Remove former navbar implementation

### LINKED ISSUE #5848: Remove former navbar implementation
## Conditions of satisfaction
- [x] The former navbar implementation is removed from Quickstart.
- [x] AZ Navbar replaces the removed implementation.
- [x] The option to disable AZ Navbar is removed.

djcelaya:
## Description

Closes #5848. Removes the former main navigation implementation so that AZ Navbar is the only main navigation in Quickstart, and removes the theme setting that allowed it to be turned off.

**Background:** AZ Navbar was added in 3.2.0 behind the `az_navbar` theme setting, which defaulted to `false` (#5107). In 3.4.0 the default was flipped to `true` and a post update enabled it on every existing site, with the setting marked deprecated (#5336). The checkbox itself stayed editable, so a site could still turn it back off. This PR completes that deprecation.

## Changes:

1. **Replace the former `menu--main.html.twig` with the AZ Navbar template.** The AZ Navbar markup lived in `menu--main-navbar-az.html.twig` and was selected by a theme suggestion. It is now simply `menu--main.html.twig`. Renaming it over the former template matters: az_barrio is a Bootstrap Barrio subtheme, so deleting the file outright would have silently fallen back to the parent theme's `menu--main.html.twig` instead of removing the old implementation. This also means the AZ Navbar markup is used anywhere the main menu renders, not just in one block.
2. **Always add the `navbar-az` class to `#navbar-top`.** The class was previously conditional on the setting. With one implementation left there is nothing to branch on.
3. **Remove the `menu__main_navbar_az` theme suggestion and the `az_navbar` preprocess variable.** Both existed only to switch between the two implementations. With the template consolidated in step 1 and the class hardcoded in step 2, nothing reads them.
4. **Remove the `az_navbar` theme setting, its config schema entry, and its default value.** This is the condition of satisfaction for the issue — the option to disable AZ Navbar is gone from `/admin/appearance/settings/az_barrio`. The schema entry carried the deprecation notice, which is no longer needed once the key itself is removed.
5. **Empty `az_barrio_post_update_enable_az_navbar()` and add `az_barrio_post_update_delete_az_navbar_setting()`.** The 3.4 enable hook is obsolete once the setting cannot be read, and leaving its former body in place would write back a key that no longer has a schema entry, which trips config schema checking in tests. It is kept as a documented no-op rather than deleted, so the update registry stays consistent between existing sites — which recorded this update when they ran the 3.4 updates — and fresh installs. The new hook deletes the key. It also branches on the previous value: sites that had the setting *disabled* get a logged warning naming the markup change, because those are the only sites whose navigation actually changes. Sites that had it enabled get a plain notice.
6. **Tighten the two existing navbar test assertions** to require `.navbar-az` on `#navbar-top`, so the class is covered now that it is unconditional.

AZ Navbar Fullscreen is unaffected. Its description text changed only because it referred to disabling a setting that no longer exists.

### Release notes

```
**Removal of the former navbar implementation**

AZ Navbar is now the only main navigation implementation in Arizona Quickstart, and the theme setting to disable it has been removed.

The AZ Navbar theme setting was deprecated in Quickstart 3.4, which also enabled AZ Navbar on all existing sites. Sites that have not changed the setting since then will see no change to their main navigation.

Sites that turned the AZ Navbar setting back off after updating to 3.4 or 3.5 will see their main navigation markup change when they update to 3.6. Custom styling or scripts that target the former markup may need to be updated, including anything relying on the `flex-lg-row`, `nav-item-parent`, `keep-open`, `menu-item--expanded`, or `menu-item--collapsed` classes. A warning naming this change is logged during the database update on affected sites.

Existing sites keep their current Main navigation block depth. To use a third level of navigation, set the Main navigation block depth to 3 at Structure > Block layout.

See [Main navigation](https://quickstart.arizona.edu/site-admin/managing-menus/main-navigation) for more information.
```

## Related issues

Closes #5848

Completes the deprecation started in #5336 (Upgrade existing sites to use AZ Navbar and deprecate theme setting), which in turn built on #5107 (Add support for AZ Navbar).

## How to test

**Case 1 — fresh install from this branch.**
The quickest check — build this branch locally; no special setup needed.
- The `az_navbar` key never exists in `az_barrio.settings`.
- The main navigation renders as AZ Navbar, with the Main navigation block at depth 3.

The remaining cases cover update paths. Because the setting has changed defaults twice, existing sites can be in several different states going into this update, and each is worth testing separately.

The most convenient baseline for both is a local build of `main`, which installs `az_navbar: true` and still ships the former navigation implementation, so neither case needs a real site. A real site running 3.5.x is the more faithful test where one is available. To move a local build onto this branch, check out the branch in the same Lando app and run `drush cr` before `drush updb -y`, so the new templates and post updates are picked up. Each update-path case starts from its own `main` build — once `drush updb -y` has run the post update is recorded as complete, so rebuild before testing the other case.

In both cases, check the logged messages with `drush watchdog:show --type=az_quickstart` (or Reports > Recent log messages) and confirm the key is gone afterward with `drush config:get az_barrio.settings az_navbar`, which should return nothing.

**Case 2 — `az_navbar: true` (expected for nearly all existing sites).**
Any site that has run the 3.4 database updates, or was installed fresh on a 3.4 or 3.5 release or on current `main`, is in this state. A plain build of `main` is already in it and needs no further setup.
- Capture the rendered main navigation markup before checking out this branch. Templates are code, so the markup changes as soon as the branch is deployed, not when `drush updb` runs.
- After `drush updb -y`: the `az_navbar` key is deleted, a plain notice is logged ("Deleted removed az_navbar configuration key during post update."), and **the main navigation markup is unchanged**. This is the important regression check — the vast majority of sites must see no visual difference.

**Case 3 — `az_navbar: false` (site opted back out after 3.4/3.5).**
No site is known to be in this state, so simulating it is the practical option. From a build of `main`, run `drush config:set az_barrio.settings az_navbar 0 -y` followed by `drush cr`; the site then renders the former navigation implementation. On a real 3.5.x site, set the same value before deploying this branch.
- Capture the rendered main navigation markup before checking out this branch; it should be the former implementation (no `navbar-az` class on `#navbar-top`, `navbar-nav flex-lg-row` on the menu `<ul>`, parent items as a single `dropdown-toggle` button).
- After `drush updb -y`: the key is deleted and a **warning** is logged naming the markup change and linking the Main navigation docs.
- The main navigation now renders as AZ Navbar: `navbar-az` on `#navbar-top`, `<span class="vr">` separators between top-level items, and parent items that have a link rendering as a split `btn-group` toggle.

**Also check, on any of the above:**
- `/admin/appearance/settings/az_barrio` no longer offers an **AZ Navbar** checkbox, still offers **AZ Navbar Fullscreen**, and saves without error.
- Menu depth: existing sites keep their current Main navigation block depth (2 for sites updated from before 3.4), so they get AZ Navbar styling without a third level until they raise it themselves. This is intentional and matches the decision made in #5336.
- Third-level navigation still expands and collapses correctly on a site with the block depth set to 3.
- Disabling the Main navigation block removes `#navbar-top` from the page entirely, as before.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

- [x] My change requires release notes.


themes/custom/az_barrio/az_barrio.post_update.php
joeparsons, line 1: I think we may want to leave the existing `az_barrio_post_update_enable_az_navbar()` as-is or potentially just make it an empty function rather than deleting it.

djcelaya: Hi, @joeparsons! Thanks for your review. Is [this](https://github.com/az-digital/az_quickstart/pull/5958/changes/c7e3bd5738d45d4c069a1959cb60edd06b51cd74) what you're thinking?

joeparsons: Hi @djcelaya, yes, except I think the old update function should remain where it is in the file (with its content removed) and the new update function should move to the end of the file.

djcelaya: Hopefully fixed with https://github.com/az-digital/az_quickstart/pull/5958/changes/cabe0ece158a39caf9ec32cfa6343491e1ee0120. 🤞🏼 I found the extra comments helpful but let me know if you would prefer them removed.

djcelaya:
I have confirmed case 2 by local simulation. I built main locally, switched to `issue/5848`, and [ran `lando drush updb -y`](https://github.com/user-attachments/assets/65ecfe53-eac8-47d3-bdcb-925495ab4e4f).

[Before](https://github.com/user-attachments/assets/c5db81b4-fe38-450d-8cb3-3448b8935baf) vs [after](https://github.com/user-attachments/assets/86e3f1a4-9ed8-4a1d-852e-4f3db1bcaced). [Log message](https://github.com/user-attachments/assets/950f8889-11ab-42c5-b2c3-7e02494d674f) confirmed after applying db update.

djcelaya:
I have confirmed case 3 by local simulation as well. I built main locally, [unchecked "Enable AZ Navbar"](https://github.com/user-attachments/assets/4401e83a-5f64-49f7-94d0-411a4ca935fd), switched to `issue/5848`, and then ran [`lando drush updb -y`](https://github.com/user-attachments/assets/722a32bd-070f-42f0-8d96-9b21ccea8765).

[Before](https://github.com/user-attachments/assets/26e2c8b8-9e16-48ea-8dc1-bc0fbe6b316a) vs [after](https://github.com/user-attachments/assets/9540ab5f-1f4a-4d93-ab23-c64d5883758b). [Log message confirmed](https://github.com/user-attachments/assets/82313c07-2272-40e4-aa1e-4a113f7fdc40) after applying db update.

themes/custom/az_barrio/az_barrio.post_update.php
joeparsons: > Hopefully fixed with [cabe0ec](https://github.com/az-digital/az_quickstart/commit/cabe0ece158a39caf9ec32cfa6343491e1ee0120). 🤞🏼 I found the extra comments helpful but let me know if you would prefer them removed.

Looks good to me now.  Thanks.

danahertzberg (approved):

themes/custom/az_barrio/az_barrio.post_update.php
```diff
@@ -293,3 +300,26 @@ function az_barrio_post_update_enable_az_header_blue(&$sandbox = NULL) {
...
+  $az_navbar = $theme_settings->get('az_navbar');
+
+  // Delete the az_navbar key if it exists. AZ Navbar is now the only main
+  // navigation implementation and can no longer be disabled.
+  if ($az_navbar !== NULL) {
+    $theme_settings
+      ->clear('az_navbar')
+      ->save();
+    if ($az_navbar !== TRUE) {
+      \Drupal::logger('az_quickstart')->warning('The AZ Navbar theme setting was disabled on this site. The former main navigation implementation has been removed and AZ Navbar is now always used, which changes the main navigation markup. Custom styling that targets the former markup may need to be updated. See https://quickstart.arizona.edu/site-admin/managing-menus/main-navigation');
```
tadean, line 319: Should this logger message be `az_barrio` rather than `az_quickstart`, or is `az_quickstart` intended?

camikazegreen (approved):
Approving both the requirements and that this meets all of the requirements.

themes/custom/az_barrio/az_barrio.post_update.php
djcelaya: We decided in AZ Digital on 9/11 that this is OK as-is.

joeparsons (approved):
Testing this on existing sites will definitely be important but I'm approving in anticipation of the pre-release testing we'll be doing for alpha1.

bberndt-uaz (approved):

tadean (approved):

djcelaya:
After copying the DB from live onto https://azqs5958-azs-abs.pantheonsite.io/, I observed that the [theme setting was correctly removed from the UI](https://azqs5958-azs-abs.pantheonsite.io/admin/appearance/settings/az_barrio). I then [ran the DB update](https://github.com/user-attachments/assets/5533d1d6-15c3-4de4-9fea-243159d76123) on this PR that removes the theme setting from the DB and saw the [log messages](https://azqs5958-azs-abs.pantheonsite.io/admin/reports/dblog/event/500522) as expected. Everything looks good to me.

trackleft (approved):

PR #5978, opened 2026-09-10
Remove data-bs-parent from accordion items.

djcelaya:
The attribute previously pointed at the collapse element's own id, which made Bootstrap's mutual-exclusion inert: _getFirstLevelChildren() searched inside the panel itself and never matched a sibling, so every accordion has in practice allowed multiple items open at once.

Pointing it at the real container activated that grouping for the first time, which broke Expand all (Collapse.show() hides shown siblings, and its _isTransitioning guard then aborts the remaining items) and changed behavior for every existing accordion. The conditional form suppressed it only when clipboard_link was enabled, but the setting that cannot tolerate mutual exclusion is expand_all, so the demo content masked the regression.

Quickstart has never wanted one-item-at-a-time accordions, so drop the attribute entirely rather than re-key the condition. Deep linking is unaffected: Arizona Bootstrap's hash handler treats a null data-bs-parent as "nothing to close" and scrolls once shown.bs.collapse fires.

- **Patch release changes**
   - [ ] Bug fix
- **Minor release changes**
   - [ ] New feature
- **Other or unknown**
   - [ ] Other or unknown

### Drupal core
- **Patch release changes**
   - [ ] Security update
- **Minor release changes**
- **Other or unknown**
   - [ ] Other or unknown

### Drupal contrib projects
- **Patch release changes**
   - [ ] Security update
   - [ ] Add new module
- **Minor release changes**
   - [ ] Major level update
- **Other or unknown**
   - [ ] Other or unknown

## Checklist

### DIFF
--- a/modules/custom/az_accordion/templates/az-accordion.html.twig
+++ b/modules/custom/az_accordion/templates/az-accordion.html.twig
@@ -27,7 +27,7 @@
     {% endif %}
   </button>
 </h3>
-<div id="{{ accordion_item_id }}" class="accordion-collapse collapse {{ collapsed }}" aria-labelledby="{{ accordion_header_id }}"{% if not clipboard_link %} data-bs-parent="#{{ accordion_container_id }}"{% endif %}>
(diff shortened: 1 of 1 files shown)


danahertzberg (approved):
This works as expected for QS.

